PR #15, opened 2016-05-10
Added logging for invite spec

alisterscott:
This enables Calypso logs and captures them after the invite has been
accepted.

It also captures these logs on any test failure - to help us diagnose
what went wrong

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 screenshots/
+logs/
 node_modules/
 .idea/
 # image-uploads/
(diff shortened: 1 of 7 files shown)


alisterscott:
@hoverduck I've merged this so just a heads up incase this affects any of the visual comparison tests in firefox etc. It shouldn't, but just in case.

alisterscott:
@hoverduck: this approach actually means we could check for JavaScript console errors on every page we instantiate. It's just that just about every page _has_ errors so every test would fail.

PR #24, opened 2016-05-12
Drop mocha git submodule, use package.json instead

ockham:
It's possible to pin a dependency in `package.json` to a specific commit, even in a project fork, and let `npm` take care of it instead of using `git submodule`.

Note -- this is untested, I haven't finished setting up e2e tests on my system yet. Proceed with care ;-)


hoverduck:
I was trying to merge this change in (finally, it's only been sitting out there for a month and a half), and I messed up the rebase.  Since it's a simple change in itself, I just recreated it in #103

PR #32, opened 2016-05-18
Media: add upload tests for some basic filetypes

gwwar:
I'm adding a few cases for Editor image upload. We has some recent regressions and some e2e tests could help us catch these more quickly.

Questions:
- ~~Should I sanitize the filenames in `createFileWithFilename`, or just remind folks to use it cautiously?~~
- ~~Would you prefer that I add test cleanup to delete the uploaded files Calypso. Without them we'll eventually run out of media storage space for our test account.~~

cc @alisterscott @hoverduck @aduth

### DIFF
--- a/lib/media-helper.js
+++ b/lib/media-helper.js
@@ -1,6 +1,7 @@
 import fs from 'fs-extra';
 import path from 'path';
 import webdriver from 'selenium-webdriver';
+import sanitize from 'sanitize-filename';
 
 export function createFile( notRandom ) {
 	let d = webdriver.promise.defer();
@@ -22,6 +23,27 @@ export function createFile( notRandom ) {
 	d.fulfill( {
 		imageName: newImageName,
 		fileName: newFileName,
+		file: newFile
+	} );
+
+	return d.promise;
+}
+
+export function createFileWithFilename( filename, skipNameCheck ) {
+
+	if ( ! skipNameCheck && sanitize( filename ) !== filename ) {
+		throw new Error( `Invalid filename given ${ filename }` );
+	}
+
+	const d = webdriver.promise.defer();
+	const originalFileName = 'image0.jpg';
+	const originalFile = path.resolve( __dirname, `../image-uploads/${originalFileName}` );
+	const newFile = path.resolve( __dirname, '../image-uploads/' + filename );
+	fs.copySync( originalFile, newFile );
+
+	d.fulfill( {
+		imageName: filename,
+		fileName: filename,
 		file: newFile
 	} );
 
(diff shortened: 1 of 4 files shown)


hoverduck:
> Should I sanitize the filenames in `createFileWithFilename`, or just remind folks to use it cautiously?

I think it's fine either way.  The end user of these scripts is never going to be the general public, so it's okay if it's not perfect.  But at the same time, it could save somebody a minor headache down the road to do a little extra work on it now and put it in.

> Would you prefer that I add test cleanup to delete the uploaded files Calypso. Without them we'll eventually run out of media storage space for our test account

It's a good excuse to test the delete workflow as well, so I'd say it's worth putting in.  But we also have some crons to clean out other things that build up over time, so we can always add something there too.

gwwar:
Sounds good @hoverduck I'll be sure to add filename sanitization and the delete workflow.

gwwar:
@hoverduck @alisterscott I've added some new tests for media upload that are ready to be looked at. I ended up not needing `clickBackWithUnsavedChanges` but thought it might be of use to someone else.

alisterscott:
I had a look locally and it looks good, well done. Whilst it ran perfectly on desktop mode, the mobile size didn't work which I believe is because the dialog is slightly different, so you might need to add a little bit of conditional logic for this.

It highlighted that there is a bit of dependency between the different file uploads, because if the first upload fails for example closing the media dialog, the second test also fails because the media dialog is open. The way around this I guess would be to start by navigating to a specific page each time, but this will add execution time 🤔

I noticed you check for the existence of errors after upload, is there also a positive assertion you could do after the file is uploaded, like it is displayed in the media gallery by name? It might need to have a unique name based on a timestamp to ensure previous files don't invalidate this assertion.

gwwar:
> I noticed you check for the existence of errors after upload, is there also a positive assertion you could do after the file is uploaded, like it is displayed in the media gallery by name?

Let me take another look. I think there wasn't that much in the DOM to latch onto depending on the type.

> Whilst it ran perfectly on desktop mode, the mobile size didn't work which I believe is because the dialog is slightly different, so you might need to add a little bit of conditional logic for this.

I'll take a look at mobile. :)

gwwar:
@alisterscott 👍  I fixed the delete call for the mobile screen size and moved the modal dismiss to test cleanup so any failing tests won't leave the modal open.

alisterscott:
LGTM 👍 
Feel free to merge it when you're ready

gwwar:
Thanks for the reviews! Going to squash a bit.

PR #38, opened 2016-05-26
Fix visdiff tests in parallel execution

hoverduck:


hoverduck:
Re-doing the FF fix in a separate branch to avoid this conflict

PR #57, opened 2016-06-07
Plans: update page helpers

gwwar:
After fixing issue https://github.com/Automattic/wp-calypso/issues/4486, in desktop view we now select a plan immediately if we click in the following areas
ddd5f55e-f827-11e5-8081-fcf7cf89beb5

This PR updates the helpers to only wait for the plan button when we detect that we are in mobile view. This had been previously timing out tests.
screen shot 2016-06-07 at 10 53 16 am

cc @alisterscott @hoverduck


rralian:
I get five passing tests when running this. If that's expected, I think we can 🚢  this.

lamosty:
When I didn't checkout this branch, I got 3 failed tests out of 58. Now, all 58 tests pass. Nice work! :ship:

PR #64, opened 2016-06-10
After a failed test, check for the existence of the 'navigate away' dialog and get rid of it

alisterscott:
This stops subsequent tests failing with:

`unexpected alert open`

unexpected alert


alisterscott:
@hoverduck: can you please 👀  at this one too?

alisterscott:
Merging this as I really would like it merged for the weekend

PR #66, opened 2016-06-10
Make sure the tinyMCE editor is shown, if not refresh and check again

alisterscott:
This caters for the scenario where the editor occasionally fails to load - this refreshes the browser to see if the issue is intermittent, it fails if it still doesn't load

1465496894862-failed-mobile-new-user-can-create-a-new-post


alisterscott:
Merging this as I would like it in for the weekend

PR #76, opened 2016-06-14
Add live branch option

hoverduck:
I don't think this is ready to merge yet, but putting it out here for review (ping @alisterscott).  There are three issues:

1) I have a function in the driverManager that directly loads https://calypso.live/?branch=[branchName] to ensure that the cookies and everything are set for the given branch name.  It's called first thing when starting the browser, and also every time the cookies are cleared.  I'm not sure if this is the cleanest way to configure things, but I couldn't come up with a way to guarantee the query string was added everywhere it's necessary.  Thoughts?  My other approach was to add the query string parameter to the base-container under the `if visit === true` section, but that had issues with the `wp-login.php` call.

2) I'm not sure if it's because of my implementation above, or something entirely different, but it runs fine locally and crashes every time on CircleCI, failing to load the page.  Feels like maybe a Chrome version issue, but it happens even on the most recent.  Still investigating.

3) It might be good to add some smarts to detect if the branch hasn't booted yet so we can provide a useful error message.

### DIFF
--- a/config/test.json
+++ b/config/test.json
@@ -1,6 +1,8 @@
 {
-  "calypsoBaseURL": "https://wordpress.com",
-  "explicitWaitMS": 20000,
+  "calypsoBaseURL": "https://calypso.live",
+  "liveBranch": "true",
+  "branchName": "try/branch-for-e2e-testing",
+  "explicitWaitMS": 30000,
   "mochaTimeoutMS": 160000,
   "mochaVisDiffTimeoutMS": 320000,
   "mochaDevDocsTimeoutMS": 500000,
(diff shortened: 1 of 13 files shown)


alisterscott:
@hoverduck: I've got this fairly much working now in a fairly simple manner which upon logging on switches to the live branch from the `root page`, or when visiting either the `/start` or `/accept-invite` entry points.

The only issue is when running in CircleCI the connection to calypso.live is too slow so a lot of tests time out with empty screens. This will probably be a blocker for this project since we'll have to increase timeouts and will have much more flaky results.

This might be enough of a business-case to convert to more scalable infrastructure instead of calypso.live.

hoverduck:
@alisterscott I went ahead and merged this in since it looks like everything is functional.  But I agree the stability of calypso.live coming from the CI hosts isn't very reassuring, so we won't be able to actually use this full-time yet.

PR #94, opened 2016-06-20
Added more explicit wait/sleep statements to themes page

hoverduck:


alisterscott:
Is there a longer term way to do this without sleeping?

hoverduck:
I didn't get to it today, but I plan on writing an issue on wp-calypso to document the fact that the theme price dropdown auto-closes before the theme list is fully populated...same with the ellipsis menu.

The other sleep I added was to account for the fact that the search field is re-populated when the page is reloaded now (I think that's a recent fix, right?).  So we either need something else to trigger off of to know that the page contents have settled, or manually re-do the search so we can guarantee the contents.

hoverduck:
It would probably be a good practice for us that when we add a "hack" type fix like this in that we write an issue to document it so we know to go back and add a better solution in later.

alisterscott:
> I didn't get to it today, but I plan on writing an issue on wp-calypso to document the fact that the theme price dropdown auto-closes before the theme list is fully populated

I noticed this bug today and raised it: https://github.com/Automattic/wp-calypso/issues/6168

> same with the ellipsis menu

I couldn't reproduce this part, can you please add a comment to the issue about this?

PR #98, opened 2016-06-23
GuidedTours: Force a tour, check if initial step is present

ehg:
Checks if an initial step is present, when we force the `main` tour with a the `tour` query arg.

TODO:
- [x]  should make a GuidedTour component
- [x] also probably add a LoginFlow function to visit the tour, to avoid
  reloading the page


alisterscott:
This is quite old and out of date with the codebase. If this is still required please open a new PR.

PR #100, opened 2016-06-23
Try higher timeouts for live branches

hoverduck:
@alisterscott with this I've enabled a build_parameter to override the NODE_ENV variable to 'live-branches' to give the tests a higher timeout value.  I ran it against https://calypso.live?branch=master and it ran fine (a couple of failures on the themes test, but nothing unique to this).

You did most of the test runs while getting the live branch code working on our end...do you think most of the failures you were seeing would be solved by higher timeouts like this, or is there a more fundamental stability issue on the calypso.live backend?

### DIFF
--- a/config/live-branches.json
+++ b/config/live-branches.json
@@ -1,0 +1,9 @@
+{
+  "calypsoBaseURL": "https://calypso.live",
+  "explicitWaitMS": 40000,
+  "mochaTimeoutMS": 320000,
+  "mochaVisDiffTimeoutMS": 640000,
+  "mochaDevDocsTimeoutMS": 750000,
+  "browser": "chrome",
+  "saveAllScreenshots": true
+}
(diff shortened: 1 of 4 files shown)


alisterscott:
@hoverduck: this looks fine, but I think once we're potentially running multiple builds against different live branches it may slow down even further. I guess we can just try it and see

alisterscott:
I can't get this to pass against calypso.live when pretty much every other build against WordPress.com is passing...

hoverduck:
> I can't get this to pass against calypso.live when pretty much every other build against WordPress.com is passing...

Boo.

I'm merging this in anyway so I can troubleshoot with the latest changes from `master` without having to worry about a rebase.  If nothing else this gives the infrastructure for other temporary overrides to the NODE_ENV in CI.

PR #101, opened 2016-06-23
Create a common approach to reporting warnings to Slack

alisterscott:
Create a helper function that allows easy recording of warnings to slack
* Ensure it only works if configured - defaults to off for local environments
* Includes screen size and CircleCI build # and branch
* Writes to console so it's amongst test results also
* Allows an option to stop reporting the same issue in the same build - so missing ab test keys are only reported once


alisterscott:
@hoverduck: I've merged this since it gives us benefit straight away in that the localstorage messages no longer repeat, and we can see the build # etc. But if you wouldn't mind reviewing it anyway.

hoverduck:
LGTM!

PR #106, opened 2016-06-24
Replaced sleep statements with waiting for a specific className

hoverduck:
Depends on https://github.com/Automattic/wp-calypso/pull/6308


hoverduck:
The Themes page loading issues have been resolved on the Calypso side, so this is no longer necessary

PR #109, opened 2016-06-28
Check for javascript console errors automatically

### LINKED ISSUE #5: Capture JavaScript errors on every Calypso page
There's two parts to this:

1. Add capability for Calypso to provide JavaScript errors for any pages
2. Check there's no errors when each page object is instantiated

alisterscott:
Closes #5


alisterscott:
I'm merging this in, but leaving it OFF since there's so many console errors - we can return to it later if we reduce the console noise

PR #127, opened 2016-07-06
Try re-searching once for domain names on error

alisterscott:


alisterscott:
Failing due to a known issue with currencies - merging regardless

PR #128, opened 2016-07-06
Added trashing the comment to the notifications spec to avoid build up of comments

alisterscott:


alisterscott:
This isn't necessary any more

PR #137, opened 2016-07-12
Update stats page to handle when all sites are selected

alisterscott:
When all sites are selected, the stats page selector is wrong
Updating it to be valid for all sites or a single site


alisterscott:
Failed due to separate payments error

PR #154, opened 2016-07-22
Switch checkout to use Stripe

janm6k:
I have enabled Stripe for 15% of payments, so using random card will no longer work reliably.

I don't have the tests to set up locally - I have not tested this change.


hoverduck:
@janm6k Looks good to me, feel free to merge

PR #166, opened 2016-07-28
Integrate mobile (iOS) Appium tests with general e2e tests

hoverduck:
@alisterscott - Can I get your :eyes: on this before I merge?  I think it's safe, all of the mobile-specific code is well isolated...but should I follow your lead with the Jetpack tests and create another separate repo instead?  I'm still torn on the decision, but leaning towards keeping it integrated.  Although with Jetpack being separate there's a precedent for splitting off...


hoverduck:
Closing, these changes were included with #167

PR #177, opened 2016-08-03
Set user agent to 'wp-e2e-tests' in Chrome and Firefox

alisterscott:
@hoverduck: can you please 👀  this? I could test Chrome locally, but Firefox wouldn't start and I went into a rabbit-hole of updating the [Marionette driver](https://developer.mozilla.org/en-US/docs/Mozilla/QA/Marionette/WebDriver) which I thought could break our CI systems where we use Firefox (Applitools?)

Also, I believe we run some NUX tests against IE11 in Saucelabs, I don't believe it's possible to override the user-agent for IE

### DIFF
--- a/lib/driver-manager.js
+++ b/lib/driver-manager.js
@@ -91,7 +91,7 @@ export function startBrowser() {
 				options = new chrome.Options();
 				options.setProxy( this.getProxyType() );
 				options.addArguments( '--no-sandbox' );
-				options.addArguments( 'user-agent=wp-e2e-tests' );
+				options.addArguments( 'user-agent=Mozilla/5.0 (wp-e2e-tests) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.106 Safari/537.36' );
 				builder = new webdriver.Builder();
 				builder.setChromeOptions( options );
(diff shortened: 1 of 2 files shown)


alisterscott:
Since we use twitter.com to verify a tweet has been sent, as we're now changing the user agent the twitter view is different as it doesn't know what browser we are using. What are your thoughts on this @hoverduck ?

1470206713302-failed-mobile-can-see-post-message

hoverduck:
I've updated the UA to be `Mozilla/5.0 (wp-e2e-tests) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.106 Safari/537.36`, which includes enough extra detail for Twitter to be happy, and we can still pattern match against `wp-e2e-tests` for the account filtering.

All the tests passed, so I'm going to go ahead and merge this in.

On the subject of Firefox, as of yesterday we're not using it anymore for the standard visdiff tests, so it will only be an issue for cross-browser checks.  And those don't do account signup, so setting the UA shouldn't be a concern.

The IE signup tests however are probably an unavoidable problem if we can't manually change the UA...but those run relatively infrequently compared to the Chrome signups, so hopefully not a significant number to throw of statistics.

PR #182, opened 2016-08-04
highlight element api

manoj9788:
Useful to see visually during the webdriver run

### DIFF
--- a/lib/driver-helper.js
+++ b/lib/driver-helper.js
@@ -6,9 +6,27 @@ import * as SlackNotifier from './slack-notifier.js';
 
 const explicitWaitMS = config.get( 'explicitWaitMS' );
 
+export function highlightElement( driver, selector, waitOverride ) {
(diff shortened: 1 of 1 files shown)


manoj9788:
for now I have added this only to clickWhenClickable we can add the function call to other API's once its proved working. Its working fine in my local.

hoverduck:
Works well for me.  My only thought is that before adding it to any other function calls is it might be useful to add a conditional check for the `CI` environment variable to disable the highlighting when we're running it without a visible window to watch.  That will avoid throwing off our visual diff tests.

hoverduck:
Another option would be to make it configurable via a variable in the NODE_CONFIG setup.  That way it would be easy to turn on/off when you want it locally, and we just won't add it to the CI setup.

manoj9788:
I agree, its good to add a On/Off flag.

PR #185, opened 2016-08-08
Updated themes-page to use new theme filter buttons [skip ci]

### LINKED ISSUE #152: Use themes pseudo-select dropdown instead of editing URL
Right now the themes list is filtered by directly editing and reloading the URL (https://github.com/Automattic/wp-e2e-tests/commit/9fd5292a602bb065a85fbbe856b73ec576049f6d), because the select dropdown has been unreliable at the speed the tests run.  The filter slug was updated in https://github.com/Automattic/wp-calypso/pull/6886, which broke the tests, but did not break the actual user experience.  I've fixed the tests in #151.

This issue is to track the fact that using the URL directly is just a workaround, and at some point we should go back and get the dropdown working correctly (either by handling it in our tests,  or driving a fix to wp-calypso).

cc: @alisterscott

hoverduck:
This change is for https://github.com/Automattic/wp-calypso/pull/7214, and closes #152

### DIFF
--- a/lib/pages/themes-page.js
+++ b/lib/pages/themes-page.js
@@ -5,24 +5,14 @@ import * as driverHelper from '../driver-helper.js';
 export default class ThemesPage extends BaseContainer {
 	constructor( driver ) {
 		super( driver, by.css( '.theme__active-focus' ) );
+		this.allThemesSelector = by.css( 'a.item-index-0' );
+		this.freeThemesSelector = by.css( 'a.item-index-1' );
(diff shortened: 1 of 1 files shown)


budzanowski:
Change from https://github.com/Automattic/wp-calypso/pull/7214 will now be pulled from https://github.com/Automattic/wp-calypso/pull/7348

alisterscott:
This isn't necessary any more, closing it now.

PR #190, opened 2016-08-08
Added config flag for highlighting elements before clicking

hoverduck:
New flag in NODE_CONFIG - highlightElements (true/false)

cc: @manoj9788


hoverduck:
Pairs with #182

manoj9788:
Awesome thanks @hoverduck

PR #191, opened 2016-08-08
Add support for Clickable Domain Rows

coreh:
This should (hopefully) fix the E2E tests for the new clickable rows interface we're adding as part of the following PR: https://github.com/Automattic/wp-calypso/pull/7239


alisterscott:
LGTM 👍

alisterscott:
(the errors on Circle are another issue with site title being changed)

PR #197, opened 2016-08-10
Add instructions for testing mobile apps

hoverduck:


hoverduck:
@alisterscott Can you take a look at this and see if there's any missing information?

alisterscott:
@hoverduck: thanks for this. 
I have now got the iOS app building and running locally, but I am not sure how to generate the .app file that Appium needs to run.
How did you do this?

hoverduck:
@alisterscott: It definitely seems harder than it should be to actually get a working build.  I should probably just talk to the mobile devs to get a lesson on how they get it to work, because I'm sure there's something simple I'm missing.

But it's easy if you just check out the branch of the most recent release in the App Store.  So right now that's `release/6.3.1`.  Once you have that locally you can just run `rake dependencies` and that will get everything set up and built in wherever you have XCode set to send the build products.

You'll also need to set up the wpcom credentials file - https://make.wordpress.org/mobile/handbook/pathways/ios/tutorials-guides/#3-%c2%a0setup-wpcom_app_credentials

Give all that a try and see if it works.  It's still possible there's something that I configured inside XCode previously that's making it work for me but not you.

rachelmcr:
If you build and run the app in Xcode, you can also go to the Project navigator (left-hand menu in Xcode), open the Products folder, and find WordPress.app there. Right-click and select Show in Finder to get to the .app file for the build.

alisterscott:
Closing as iOS tests will live in iOS repository

PR #198, opened 2016-08-11
Fully removing Followers after invites

hoverduck:
Previously the "Invite Follower" test was only deleting the new user from the "Email Followers" list, but left them in the main "Follower" list (since they created an account).  This caused a buildup of ~3500 followers, which was overflowing our Mailosaur account with too many daily e-mails.

With this PR we now remove both the e-mail follower and the other one (although only the latter is actually necessary, I left the direct e-mail remove in there so we still test that functionality).


hoverduck:
@alisterscott Mind reviewing this one?  I think it's all clean, but you've done more work with the invites system.

alisterscott:
LGTM  👍  merging now to reduce the number of emails

PR #214, opened 2016-08-25
Updated accounts for signup/invite to prefix e2eflowtesting

hoverduck:
Per conversation here - p4qSXL-1r9-p2


hoverduck:
@alisterscott Can you take a look and make sure I'm not missing anything?

hoverduck:
@alisterscott I forgot you're AFK.  I'm going to go ahead and merge this in before the weekend, but I'd appreciate a double-check when you get back and we can revert/add to it as necessary.

PR #239, opened 2016-09-06
Changed preview link selector from anchor to button

hoverduck:


lancewillett:
@hoverduck Couldn't the selector just be `'.customize-controls-preview-toggle'` without the HTML element prefix? CSS works that way nicely to just grab any element with that class — unless it has to be this specific one...

hoverduck:
Yeah, that would be simpler and would have avoided this problem.  But at the same time, it's good to be specific in case it changes the other way...what if this class was instead applied to a span that wrapped a button and some other text? We'd need the click action to be sent to the button itself, not the wrapper.

PR #251, opened 2016-09-20
Premium plan verification test

hoverduck:
This PR was built in the A8C GM 2016 Kibbling Flow class.  I reworked it a bit from what we built in class, renaming the `comparePlans()` function to be more accurate `openPlansTab()`, and moved the plan verification test out of its own spec file into the other general Plans spec.

cc: @alisterscott @rachelmcr @kevko @stephdau @starxedsteph @ChrissiePollock


rachelmcr:
👍 Looks good to me. I ran through the `wp-plan-purchase-spec.js` spec on this branch and confirmed that the "Can Verify Current Plan" test passes as expected, and fails as expected when I change the selector in the `confirmCurrentPlan()` function to another value.

PR #264, opened 2016-09-28
More graceful method of constructing the known AB test string

hoverduck:
Switched from manually building a big string of AB tests to processing an array

cc: @alisterscott @bisko

### DIFF
--- a/lib/base-container.js
+++ b/lib/base-container.js
@@ -71,20 +71,24 @@ export default class BaseContainer {
 		} );
 	}
 	setABTestControlGroupsInLocalStorage( culture = 'en', flow = 'main', personalPlanValue = 'hide' ) {
-		const signupStoreKey = 'signupStore_20160927';
-		const signupStoreValue = 'designTypeWithStore';
-		const signupStoreBenchmarkingKey = 'signupStoreBenchmarking_20160927';
-		const signupStoreBenchmarkingValue = 'pressable';
-		const browserNotificationsKey = 'browserNotifications_20160628';
-		const browserNotificationsValue = 'disabled';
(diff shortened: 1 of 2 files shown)


lib/base-container.js
```diff
@@ -71,20 +71,18 @@ export default class BaseContainer {
...
-		const browserNotificationsKey = 'browserNotifications_20160628';
-		const browserNotificationsValue = 'disabled';
-		const domainSuggestionPopoverKey = 'domainSuggestionPopover_20160809';
-		const domainSuggestionPopoverValue = 'hidePopover';
-		const domainSuggestionsClickableRowKey = 'domainSuggestionClickableRow_20160802';
-		const domainSuggestionsClickableRowValue = 'clickableButton';
+		const overrideABTests = config.get( 'overrideABTests' );
 
-		const expectedABTestValue = `{"${signupStoreKey}":"${signupStoreValue}","${browserNotificationsKey}":"${browserNotificationsValue}","${domainSuggestionPopoverKey}":"${domainSuggestionPopoverValue}","${domainSuggestionsClickableRowKey}":"${domainSuggestionsClickableRowValue}","${signupStoreBenchmarkingKey}":"${signupStoreBenchmarkingValue}"}`;
+		const expectedABTestValue = overrideABTests.reduce( function( prev, curr ) {
```
bisko: This can be improved so we don't have to track the current index to make sure we place the commas in the right places.

```
const expectedABTestValue = overrideABTests.map( ( entry ) => {
	return '"' + entry[ 0 ] + '":"' + entry[ 1 ] + '"'
} ).join( ',' );
```

It will first format the entries in the proper form: `"key": "value"` with `map` and then `join` them with a comma.

PR #268, opened 2016-09-29
Add new test for opening notifications with n key

### LINKED ISSUE #244: Create new test for opening notifications with n key
To create coverage for this bug here: https://github.com/Automattic/wp-calypso/issues/1907

rachelmcr:
Resolves #244 

@alisterscott Could you review this for me? I appreciate any and all feedback. :)

### DIFF
--- a/lib/components/navbar-component.js
+++ b/lib/components/navbar-component.js
@@ -33,4 +33,12 @@ export default class NavbarComponent extends BaseContainer {
 			}
 		} );
 	}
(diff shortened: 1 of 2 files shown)


rachelmcr:
Good point about keeping the old method there. I agree, it's a good idea to keep it in, especially since the tests that follow need the tab open regardless of how it's opened. I'll add it back in.

Thanks for the review!

PR #285, opened 2016-10-07
Add support for Site Title step in signup

hoverduck:
Forces the AB test key siteTitleStep to showSiteTitleStep, which adds a new step to the signup flow, and then sets the site title to the blog's name.

This PR also dynamically numbers the steps in the spec file, so we can add/remove/move steps without editing each describe block further down the line.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -65,6 +65,6 @@
 	[ "signupStoreBenchmarking_20160927", "pressable" ],
 	[ "signupThemeUpload_20160928", "hideThemeUpload" ],
 	[ "signupSurveyStep_20161005", "surveyStepV1" ],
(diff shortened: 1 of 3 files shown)


alisterscott:
> This PR also dynamically numbers the steps in the spec file, so we can add/remove/move steps without editing each describe block further down the line.

This is great, thanks!

alisterscott:
Is this necessary any more with newer PRs? Can I close it?

hoverduck:
Bah, I didn't realize this hadn't been merged.  I think it would still be nice to have the dynamic numbering, but it's not worth rebasing this branch after all the other changes to the signup spec in the meantime.  I'll close this for now and move it over to an issue to drive adding the dynamic numbering back in later as an enhancement.

PR #288, opened 2016-10-11
Reworked actions on the Customizer page to hopefully avoid timing races

hoverduck:
Most of the actions in the Customizer page have a short sleep statement to allow for iFrame switching.  This PR just moves the action after that switch into a `.then()` block to ensure the switch is done before moving on.


hoverduck:
@alisterscott I'm going to merge this because it ran cleanly in CI, but wanted to ping you in case you had any ideas for getting rid of that sleep statement altogether, since you got that working for post preview a while back.

PR #319, opened 2016-11-11
Fix signup test after PR #9309

coreh:
This PR fixes E2E tests after removing the A/B test on https://github.com/Automattic/wp-calypso/pull/9309

PR #331, opened 2016-11-30
Revert "Fix signup test after PR #9309"

hoverduck:
Reverts Automattic/wp-e2e-tests#319


hoverduck:
Signup tests still failed on /create on desktop...but I'm merging anyway to get a better idea of what's wrong.

PR #332, opened 2016-12-02
Add default variation for the noSurveyStep ABtest

bisko:
This PR adds the default variation for the Automattic/wp-calypso#9789 PR, which will test the signup flows with and without the `survey` step.

While we're testing, we want to leave the tests to run with the default value.


michaeldcain:
👍

PR #351, opened 2016-12-22
Signup: Deprecate Survey step in Signup

bisko:
With the merge of Automattic/wp-calypso#10210 we're deprecating the Survey step in Signup for the time being.

This PR removes the Survey step from the tests and add another test for a flow where the Survey step exists for testing. 

Additionally to that, I noticed that there were several extra instances of `StartPremiumPage` and `StartBusinessPage` which were just different flows in Signup. I thought it would be good to simplify the code a bit and added the `flow` parameter to the `StartPage` class.
This way different flows can be tested without creating separate files/classes for each different flow.

TL; DR; changes list:

* Removed Survey step from all the flows
* Added a test for a flow that still has the Survey step in it
* Removed `StartPremiumPage` and `StartBusinessPage`
* Added a `flow` parameter to `StartPage` to be able to test specific flow.

cc @michaeldcain


bisko:
The merge will probably happen tomorrow, after a complete review there. 

I am planning to merge the E2E fix here and then merge the change there, to verify the tests are passing correctly. Does this sound like a good plan?

hoverduck:
> I am planning to merge the E2E fix here and then merge the change there, to verify the tests are passing correctly. Does this sound like a good plan?

Yep, that's what I'd recommend.  Specifically I'd say merge the e2e changes and then immediately cancel the CI build since we know it's going to fail.  Then when the wp-calypso changes are deployed that will kick off a fresh run that should be all green.

bisko:
That's awesome! I will follow these steps with the merge.

PR #360, opened 2017-01-09
Add new Jetpack e2e documentation

alisterscott:
Add some doco about configuring a new Jetpack site to run these tests against

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@ Automated end-to-end acceptance tests for the [wp-calypso](https://github.com/Au
 
 [![Circle CI](https://circleci.com/gh/Automattic/wp-e2e-tests/tree/master.svg?style=svg)](https://circleci.com/gh/Automattic/wp-e2e-tests/tree/master)
 
+**Note:** see [detailed instructions](/docs/JETPACK.md) for configuring these tests to run against the Jetpack plugin.
+
 ## Pre-requisites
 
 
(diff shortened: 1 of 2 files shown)


alisterscott:
@hoverduck: would you mind testing this doco to see whether you can get these tests running on a new WordPress instance? You can use the same test social media accounts if you like for ease of set up.

hoverduck:
@alisterscott I wasn't able to get this to work.  I tried both with a local WP install on my laptop (via [this setup](https://github.com/hoverduck/falconwp)) and with a fresh Pressable site.  In both instances the connection failed, leaving me on this screen:

screen shot 2017-01-09 at 12 26 32 pm

Aside from that, I have a few suggestions for the documentation:

1. Add a step to install/activate Jetpack.  Pressable has it pre-installed, but not all hosts do.
2. If you're not using Pressable the `pressablesiteusername`/`pressablesitepassword`variable names are confusing, and should probably be renamed something generic.
3. After connecting manually (per Step Three) you're prompted to "Activate Recommended Features"...should I select that?
4. On Step Four, clarify that the Sharing page is under Settings.  First I was looking around in the Jetpack panel and couldn't find it

alisterscott:
Thanks @hoverduck 

I'm seeing the same URL issue now. I will discuss this with the Jetpack team and work out why...

alisterscott:
This no longer applies as Jetpack testing is done via Calypso. Closing.

PR #361, opened 2017-01-11
New ensureMobileMenuOpen() procedure, added to login flow

hoverduck:
https://github.com/Automattic/wp-calypso/pull/10505 changed the default behavior for when the My Site link is clicked on mobile, dumping the user immediately into Stats instead of on the sidebar.  This PR adds a call to ensureMobileMenuOpen to all login flows that required the sidebar to be open.

And for reasons I'm not clear on, the existing method for ensureMobileMenuOpen stopped working, and I had to find an alternate way.

### DIFF
--- a/lib/driver-helper.js
+++ b/lib/driver-helper.js
@@ -342,12 +342,12 @@ export function scrollToFindElement( driver, selector, { maxSwipes = 2 } = {} ) 
 export function ensureMobileMenuOpen( driver ) {
 	const self = this;
 	const mobileHeaderSelector = by.css( '.section-nav__mobile-header' );
+	const openSidebarSelector = by.css( 'header.current-section svg.gridicons-chevron-left' );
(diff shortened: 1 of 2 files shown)


lancewillett:
@hoverduck Might be worth asking on the original PR if the mobile change was expected. Could be only desktop was tested for the flow.

hoverduck:
@lancewillett Hmm...that's true.  I was just thinking about it in the context of it's not "broken" as it is now, but it is a little bit of an awkward flow for mobile users, depending on how you look at it.

lancewillett:
Yes, could be a UX regression for sure — especially if not intended.

hoverduck:
Ref https://github.com/Automattic/wp-calypso/pull/10567#issuecomment-271958709

alisterscott:
Closing as no longer relevant

PR #367, opened 2017-01-18
Add waitUntilThemesLoaded() to solve race condition on themes page

alisterscott:
It seems like there’s a race condition on the themes page if you select
the theme really really quickly then it doesn’t work. So this PR
introduces a wait until all the placeholders are no longer present
before continuing.


hoverduck:
LGTM :shipit:

PR #372, opened 2017-01-20
Fix/run sh support dot only

hoverduck:
This PR fixes the issue where using `run.sh` would fail to recognize the `.only` flag on a test.  This was because it would pass the `-i` option to mocha, which inverted the selection (as intended when using the SKIP_TEST_REGEX envvar).  So now that parameter is only used if the SKIP_TEST_REGEX variable is actually set.

It also adds the capability to run the Jetpack tests from the main CI repo as opposed to just the https://github.com/Automattic/wp-e2e-tests-jetpack repo if the word `Jetpack` or `jp` is in the branch name

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -8,7 +8,6 @@ general:
     - "logs"
   branches:
     ignore:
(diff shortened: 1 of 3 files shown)


alisterscott:
Thanks for this 👍

PR #381, opened 2017-02-06
Adds support for boostedPostSurvey AB test

alisterscott:
Disables the `boostedPostSurvey` ab test for e2e runs


alisterscott:
@hoverduck: would you mind reviewing and merging this?

PR #384, opened 2017-02-07
Removed check to use mobile.twitter.com

hoverduck:
The mobile version of Twitter's website is throwing API rate limit errors when accessed from the CI servers.  We're not actually doing any visual validation of the Twitter site, just verifying that the tweet showed up, so there's no need to actually check that.  This PR just removes that check and loads twitter.com for everything.


lancewillett:
Nice fix.

PR #388, opened 2017-02-14
Try a new method of clicking login

alisterscott:


hoverduck:
@alisterscott I'm curious to see if you notice this behaving any better.  If so, would it be worth updating the clickWhenClickable to use this behavior everywhere?  The only difference is that this waits until it's displayed instead of just accessible, right?

PR #389, opened 2017-02-14
Update/all post editor tests to work on jetpack

alisterscott:
This adds a new editor spec to be run against Jetpack sites on Go Daddy, Bluehost and Pressable

### DIFF
--- a/lib/components/post-editor-sidebar-component.js
+++ b/lib/components/post-editor-sidebar-component.js
@@ -84,19 +84,22 @@ export default class PostEditorSidebarComponent extends BaseContainer {
 		this.driver.wait( until.elementLocated( viewPostSelector ), this.explicitWaitMS, 'Could not locate the view blog post or page link.' );
 	}
 
-	publishAndViewContent() {
+	publishAndViewContent( { reloadPageTwice = false } = {} ) {
 		this.publishPost();
 		this.waitForSuccessViewPostNotice();
-		return this.viewPublishedPostOrPage();
+		return this.viewPublishedPostOrPage( { reloadPageTwice: reloadPageTwice } );
 	}
 
-	viewPublishedPostOrPage() {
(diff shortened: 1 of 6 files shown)


alisterscott:
@hoverduck do you mind taking a look at this?

It continually fails on CircleCI but passes locally. What I am trying to do is use the existing post editor specs but run them on Jetpack sites using Calypso. The issue (I believe) is something to do with having to authenticate to both Calypso and wp-admin on the Jetpack site (using SSO) since the front end requires you to be logged into wp-admin to see private posts for example.

But when running via CircleCI it seems like the post can't be shown as the user isn't logged in even though I use the login via SSO method to log in each scenario.

Another set of 👀 would be great as this is driving me crazy ATM

hoverduck:
@alisterscott - I thought I had it figured out, as I occasionally saw an `Invalid Nonce` error when executing the SSO (which I was able to reproduce locally frequently, possibly thanks to slow coffee shop wifi).  So I added a retry to the SSO step...it could be more graceful, in line with other retry logic we have elsewhere, but this seems to work for now.

But then even after putting this change in I still saw the problem with a successful SSO executing.  What I noticed here is that from Calypso the View Post button is pointing to an http address, which 404s.  But if you do https then it works.  I have to run now, but I can dig into this more this weekend or on Monday if you don't get a chance to.  But it seems like this could be a bug against wp-calypso to use https, or could just be a setup issue on the host to do a redirect.

One other note, I had a Git problem when pushing the changes, and it claimed that there were upstream changes that I didn't have yet.  Take a look at my last commit here and see if it's what you want, or if I inadvertently reverted something.  It's changing wp-login.php to wp-admin when attempting SSO.

alisterscott:
Thanks for your help, it seems to be passing but I don't fully understand why it didn't work before. Can we chat tomorrow about this before I merge it?

PR #391, opened 2017-02-17
Updated README and config examples

hoverduck:
This PR does the following:

1) Moved the Config / Environment Variables section of the README up so it's seen before the instructions to execute run.sh
2) Cleaned up the NODE_CONFIG parameters, added and removed the necessary entries to match what's currently in use
3) Updated the local.example.json file with all necessary accounts and other missing fields
4) Updated argument order in run.sh to match the help output


hoverduck:
@alisterscott - This is just documentation updates, so I'm going to go ahead and merge it.  But I'd appreciate it if you took a look to make sure I'm not missing anything.

alisterscott:
This looks great, thanks for doing this

PR #394, opened 2017-02-20
Switched Slack pings from @here to @flow-patrol-squad-team

hoverduck:


lancewillett:
A bit of context for the change. This gives the channel a bit less noise for Calypso developers who want to be in there but not get constant `@here` pings. They usually end up leaving the channel.

Instead, they'll probably get a ping if they are mentioned as a commit author for a failing test. That's a good thing, we want them to pause and look at the failure.

PR #395, opened 2017-02-20
Proper subteam ping format for Slack

hoverduck:
Follow-up to #394, as the subteam ping apparently has a unique format for the API - To ping a subteam there’s extra special formatting - https://api.slack.com/docs/message-formatting#linking_to_channels_and_users (down under Variables for some reason)


lancewillett:
That's-a-crazy.

PR #406, opened 2017-03-02
Support @canary test runs

alisterscott:
This marks some e2e tests as @ canary and adds a '-C' option to run.sh to run these only

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -10,12 +10,14 @@ SCREENSIZES="mobile,desktop"
 BRANCH=""
 RETURN=0
 CLEAN=0
+GREP=""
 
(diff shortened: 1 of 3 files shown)


alisterscott:
@hoverduck 

> I'm curious why you added the MOCHA+=" --compilers js:babel-register" arguments. That's already added on every run by the contents of the test/mocha.opts file, isn't it?

I noticed the Jetpack e2e tests had this so I thought it was required to work

hoverduck:
@alisterscott 

Ha...the only reason I added it there was because you had it in the previous commands directly on the Jetpack repo.  I just didn't bother to ask why at that time.  😄

PR #408, opened 2017-03-03
Try setting dns prefetch disable

alisterscott:
Trying to solve ‘timeout: cannot determine loading status’ errors


hoverduck:
The version of Chrome in the apt repository is still v54, but I was able to pull v56 directly from Google and install it.  [Didn't help](https://circleci.com/gh/Automattic/wp-e2e-tests/7204).

alisterscott:
I'm going to try adding detailed webdriver logging to see if I can work out where the exception actually happens and whether there is a pattern or whether it's particular pages in our app etc.

alisterscott:
The individual browsers didn't appear to fix our issues so it's probably best to revert this since it adds execution time to the suite

hoverduck:
@alisterscott - Reverted - https://github.com/Automattic/wp-e2e-tests/pull/418

PR #410, opened 2017-03-09
Add support for editor layout design

alisterscott:
This is controlled by an environment variable USE_NEW_EDITOR which should be set to true on CircleCI when the editor goes live from this PR: https://github.com/Automattic/wp-calypso/pull/11536


alisterscott:
@hoverduck: I've done most of the work so that all that should be needed is adding the environment variable to CircleCI.

There's currently two bugs in the editor which means some of these tests are failing against the new editor:
1. Private posts and pages won't publish
2. The trash icon doesn't appear on the mobile sidebar

PR #415, opened 2017-03-13
Try sleeping after creating the driver

alisterscott:
As per: https://bugs.chromium.org/p/chromedriver/issues/detail?id=402


alisterscott:
Didn't work

PR #420, opened 2017-03-13
Stubbed out large number of ads hosts for performance

hoverduck:
Attempting to troubleshoot persistent timeout issues on CI, I've loaded the hosts file from [this repo](https://github.com/StevenBlack/hosts) into our circle.yml

This is not meant to be a permanent solution, but if it resolves the issue we'll whittle down the list to determine any offending hosts.


hoverduck:
CircleCI can't handle the size of the .yml file, and won't execute the build.  I guess I'll go ahead and trim it down 😄

alisterscott:
Closing this as we now use uBlock for this

PR #421, opened 2017-03-13
Removed references to USE_NEW_EDITOR flag

hoverduck:
Since the new editor is now live, I removed the code for the old one (mostly because I was forgetting to set the envvar locally and constantly hitting failing tests).


alisterscott:
Completed in #422

PR #423, opened 2017-03-14
Remove checking for console errors

alisterscott:


alisterscott:
This has no effect on the timeout errors - closing

PR #429, opened 2017-03-15
Update selenium to 3.3.0

alisterscott:
Trying to update to Selenium 3.3.0

### DIFF
--- a/lib/after.js
+++ b/lib/after.js
@@ -1,4 +1,3 @@
-import test from 'selenium-webdriver/testing';
 import config from 'config';
 import SlackUpload from 'node-slack-upload';
 import * as slackNotifier from './slack-notifier';
@@ -12,7 +11,7 @@ import * as driverHelper from './driver-helper';
 const afterHookTimeoutMS = config.get( 'afterHookTimeoutMS' );
 var allPassed = true; // For SauceLabs status
 
-test.afterEach( 'Sauce Breakpoint', function() {
+afterEach( 'Sauce Breakpoint', function() {
 	const driver = global.__BROWSER__;
 
 	if ( process.env.SAUCEDEBUG && this.currentTest.state === 'failed' && config.has( 'sauce' ) && config.get( 'sauce' ) ) {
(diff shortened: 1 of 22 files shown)


alisterscott:
@hoverduck: I tried doing this but it's failing with:

`  TypeError: Cannot read property 'call' of undefined`

I have a feeling this is something to do with how webdriver works with mocha. I tried updating Mocha but it seems we've patched it with your personal repository, which we need for bailsuite.

Would you mind taking a look at this and see if updating your mocha repo fixes this?

hoverduck:
:eyes:

hoverduck:
@alisterscott The problem wasn't tied to our mocha fork (although I did use this excuse to pull in all the upstream changes, so now we're on v3.2.0 instead of 2.4.5), but instead due to the way we were calling the before/after hooks.  We have been calling `test.before( 'String Description', function() {})`, but apparently in the newer versions of Selenium it no longer has a version to handle that string description.  So instead now I took out the `test.` on all of those calls so we're using the Mocha version instead, which does handle those strings.

The tests still don't work, but I'll continue troubleshooting...

hoverduck:
> So instead now I took out the test. on all of those calls so we're using the Mocha version instead, which does handle those strings

Note that this causes some unexpected behavior in a few cases, and needs to be watched.  In the `wp-theme-multisite-spec.js` spec replacing the test.after() call for closing the customizer with just after() led to some concurrency issues with the next test in sequence running before that after() call was complete.  So in that case I've gone back to test.after() with no description.

hoverduck:
Turns out it was ALL completely inconsistent with the mocha before/after, so I've returned to the `test.` versions, but removed the offending strings.  In most cases I added a comment above the line saying what it used to be, but where it was evident from the code itself I left that off.

@alisterscott - At this point it looks like everything on this branch is working, at least so far as the core e2e tests go (there was still one timeout failure on the [last run](https://circleci.com/gh/Automattic/wp-e2e-tests/7542), but not the `cannot determine loading status` kind, so that's a plus).  I briefly checked out the visdiff tests, but I haven't looked at the Jetpack specs.  Can you give those a try?  I did not update any of the i18n specs, since those require manual setup next time we need to run them anyway.

Even if this doesn't solve the timeout issues it's probably worth going ahead and merging so we're on the latest version of Selenium, right?

alisterscott:
Thanks @hoverduck 

The Jetpack tests are all passing for me so I'm merging this now.

Issue #243, opened 2016-09-08
Create new test for clicking site content link mobile sidebar


alisterscott:
Create coverage for this bug: https://github.com/Automattic/wp-calypso/issues/1881

hoverduck:
This should also cover Jetpack sites.

Also ref this [recent failure](https://github.com/Automattic/wp-calypso/pull/11244)

chenchaoyi:
Tests added as in specs/wp-view-site-spec.js

PR #443, opened 2017-03-20
Add tests for #243, add .DS_Store to .gitignore

chenchaoyi:


chenchaoyi:
This is for adding a test flow for site view from sidebar. 

A new page was added as lib/pages/mysite-page.js. The test flow is login and go to my site tab, then click on `view site` from the top of the sidebar to check it opens the site view. Then close the view and check sidebar is shown up again.

Let me know if there is anything you would like to change. Thanks.

lib/components/sidebar-component.js
```diff
@@ -41,6 +41,9 @@ export default class SidebarComponent extends BaseContainer {
 	selectAllSites() {
 		return driverHelper.clickWhenClickable( this.driver, By.css( '.all-sites a' ) );
 	}
+	selectViewThisSite() {
+		return driverHelper.clickWhenClickable( this.driver, By.css( '.site__content[title="View this site"]' ) );
```
alisterscott: We try to avoid using English strings as locators, since a) these can and do change quite often, and b) this test couldn't run against non en locales, which we do for i18n testing on occasion. For example, in German the element looks like:

`<a class="site__content" href="http://watirmelon.blog" data-tip-target="site-card-preview" target="_blank" title="Diese Website ansehen" aria-label="Vorschau für watirmelon.blog-Website anzeigen">`

so something like:

`a[data-tip-target="site-card-preview"]`

should work

lib/pages/mysite-page.js
```diff
@@ -0,0 +1,29 @@
+// This is the page shown after clicking on the site name from side bar
+
+import { By as by, until } from 'selenium-webdriver';
+import config from 'config';
+
+import BaseContainer from '../base-container.js';
+import * as driverHelper from '../driver-helper.js';
+
+export default class MySitePage extends BaseContainer {
```
alisterscott: I would probably call this a `component` over a `page`, and it therefore probably should live under /components.

The main difference is a component isn't accessible by visiting it, so therefore it doesn't the visit parameter and you don't need to determine the URL or anything like that. So the constructor would look something like:

```
	constructor( driver ) {
		super( driver, by.css( '.web-preview__frame' ) );
	}
```

Check out `post-preview-component.js` for inspiration - as this does a very similar thing to this except it previews a post not a site

lib/pages/mysite-page.js
```diff
@@ -0,0 +1,29 @@
...
+	constructor( driver, visit = false ) {
+		let url = config.get( 'calypsoBaseURL' );
+		if ( config.has( 'liveBranch' ) && config.get( 'liveBranch' ) === 'true' ) {
+			url = url + '?branch=' + config.get( 'branchName' );
+		}
+		super( driver, by.css( '.web-preview__frame' ), visit, url );
+	}
+	isWebPreviewPresent() {
```
alisterscott: I'm not sure this is strong enough of an assertion. This will show whether there is a frame to hold content, but it could be empty.

If you look at post-preview-component you'll see where it switches to the iFrame and asserts some content is shown, for a site we may just want to assert that `body.home` is present in the frame or something quite high level.

specs/wp-view-site-spec.js
```diff
@@ -0,0 +1,71 @@
...
+	this.bailSuite( true );
+
+	test.describe( 'View site and close:', function() {
+		test.before( function() {
+			driverManager.clearCookiesAndDeleteLocalStorage( driver );
+		} );
+
+		test.it( 'Can Log In and go to My Sites', function() {
+			const loginFlow = new LoginFlow( driver );
+			loginFlow.loginAndSelectMySite();
```
alisterscott: not every spec has it currently, but we try to use an explicit return statement for the last line of each `test.it` block to ensure the WebdriverJs promise manager works as expected.

So this would be something like:

`return loginFlow.loginAndSelectMySite();`

specs/wp-view-site-spec.js
```diff
@@ -0,0 +1,71 @@
...
+		} );
+
+		test.it( 'Can view the default site from sidebar', function() {
+			this.sidebarComponent = new SidebarComponent( driver );
+			this.sidebarComponent.selectViewThisSite();
```
alisterscott: same return statement as per above, and all other blocks

chenchaoyi:
Thanks @alisterscott , for all the review comments!

alisterscott:
@hoverduck: would you mind also reviewing this?

hoverduck:
@chenchaoyi Feel free to merge.  If you still don't have permissions there just ping me and I'll hit the button.

PR #444, opened 2017-03-20
Adds a workaround for the failing signup tests on wpcalypso

hoverduck:
The signup tests running as part of the Canary builds against wpcalypso are frequently failing because the user is being redirected to the login screen instead of the checkout page.  This PR adds a check to those steps to handle the failure and retry.


alisterscott:
Thanks for this workaround, works well

PR #446, opened 2017-03-21
Remove iOS tests

chenchaoyi:
README update will come in a following PR.


alisterscott:
Looks good to me, always great to be cleaning up old code.

PR #453, opened 2017-03-28
Fix/issue 452 update add cookie

chenchaoyi:
A quick fix for specs-i18n tests. The previous cookie also has a `regular` in it, which I am not sure which new attribute it should be mapped to in the new format.


yoavf:
👍

PR #456, opened 2017-03-28
Update addCookie to use expected format

yoavf:
Otherwise an error is thrown: `addCookie called with non-cookie parameter`


chenchaoyi:
This is covered in #453 : )

yoavf:
Oops, thanks @chenchaoyi

PR #455, opened 2017-03-28
Localization test flow

akirk:
Run with: 
`./node_modules/mocha/bin/mocha specs-i18n/search-wordpress-on-google.js lib/after.js`

Compare with our Localisation test instructions (Team Global meetup project):

### WordPress.com Signup Flow Test Script
Please take notes & screenshots as you go through the steps so you can prepare for the final report.

1. Set your browser to the locale of your choice. Log out from your WordPress.com account.
2. Go to the most popular search engine for the locale (e.g. https://www.google.co.jp/ for Japanese) and search for “WordPress”.
   - Does the search ad for WordPress.com look ok? Does it sound natural and enticing?
   - Does locale.wordpress.com show up on the first page? How does the description sound?
   - Are there any untranslated WordPress.com pages indexed?
   - (Bonus) Do you find any ads or site listings that violates the trademark/domain policy?
3. Click the locale.wordpress.com link (not the ad) and go to the localized home page. Click around and see if you can find any issues on this page or any other linked pages.
4. Create a new dummy account (create a Gmail Plus address if you need an additional address) & go through the signup process.
   - Check the signup process pages for any issues using a popular local payment method (credit/debit/prepaid card, bank wire via paypal) rather than WPCOM credits (don’t forget to cancel & get a refund later).
   - Are all email messages correctly translated? Do all the links in the email work?

PR #457, opened 2017-03-28
Add acceptLanguage option to startBrowser()

deBhal:
This PR adds support for setting language in support of #455, to enable the first step in the the walkthrough:
>  1. Set your browser to the locale of your choice. Log out from your WordPress.com account.

pass the desired locale through as `startBrowser( { ..., acceptLanguage = "it" } )`, for example, to set the `accept_language` header.

Needs testing


yoavf:
Tried testing this by creating a flow that opens a page that output request headers (i.e http://headers.cloxy.net/request.php). Doesn't seem to work right - and apparently isn't possible without the use of a browser plugin or a proxy.

deBhal:
I tried running multiple tests in the one file and reusing the code, but the driver-manager stores the browser in a global variable which is then used in the `lib/after.js` which does a bunch of testing and cleanup. Theoretically we could create and quit the browser repeatedly in the one test, but changing the after hooks to let us do it would break all the other tests (unless we ran a second instance without the `lib/after.js` argument).

I think it's just easier to emulate the existing tests and duplicate the test into one file per locale:
```
for i in es pt-br de fr he ja it nl ru tr id zh-cn zh-tw ko ar sv; do   cat logged-out-redirect-spec.js.template | sed "s/LLLL/$i/" > "logged-out-redirect-$i-spec.js"; done;
```
and then fix pt-br by hand.

deBhal:
Alright, a few changes here:

I've added a -I flag to `run.sh` which causes the mag 16 to get passed in as the `LOCALE_TEST` environment variable.

I tweaked the existing tests to use this, and added a new `logged-out-redirect-i18n-spec.js` that just tests the redirect.

deBhal:
We're still not sure what to pass in to the tests, but I feel like the locale is the right way to go because it matches up with the limitations we've got on `startBrowser()` when we create the driver. We're running the tests repeatedly because we need to set the locale on the browser when we create it, and it's hard to create a new browser.

It might turn out to make more sense to pass in the names of tests, so I've added support for passing in a TEST_LOCALES variable as well. This should be a bash array, which just means adding brackets:
```
LOCALE_TESTS=(
  en-wordpress
  es-wordpress
  pt-br-wordpress
  de-wordpress
  fr-wordpress
  he-wordpress
  ja-wordpress
  it-wordpress
  nl-wordpress
  ru-wordpress
  tr-wordpress
  id-wordpress
  zh-cn-wordpress
  zh-tw-wordpress
  ko-wordpress
  ar-wordpress
  sv-wordpress
) run.sh
```
(newlines optional, any whitespace should work inside the brackets)

deBhal:
the actual RUN_ARGS should be `-i -I`

PR #460, opened 2017-03-29
Add docker wrapper for 'headless' execution

### LINKED ISSUE #390: Add Vagrant/Docker wrapper to run tests in "headless" mode
This is an enhancement suggested by @bisko - Add a wrapper to run the tests inside a Vagrant/Docker instance so they essentially run headlessly.  VNC can be used to observe them when necessary.

It's pretty much exactly what's being done via CircleCI, but this would be for running locally.

hoverduck:
Fixes #390 

This PR adds a capability to run the tests from the context of a Docker container.  This enables 'headless' execution without a visible window, as well as consolidating our Chrome/driver versions for troubleshooting.

There's plenty of room to grow with the capability to make it more useful/flexible, but this should be a good start.

@alisterscott can you give it a run through to make sure it works and the instructions in the README are clear before I merge it?


alisterscott:
This is looking good. 

I'm still trying to fully understand it, but my initial observations:

1. It wasn't obvious to me how to exit out of the interactive shell, it looks like you need to use `Ctrl+p Ctrl+q`
2. Creating/updating the container took a long time on my machine - mounting the directory and interactive mode would mostly alleviate this I believe
3. It ran first time without me creating local-docker.json even though the instructions said this is necessary
4. I can see how this would make CI easy, but my original thoughts was we'd use a local copy of the tests against chrome running remotely in a docker container, what's the benefits of building the entire project in the container?
5. I noticed CircleCI isn't running on this branch, is this because it's expecting to use Docker now?

This will be great to talk further about and refine at our meetup 👍

hoverduck:
Thanks for the feedback.  Mostly I'm just excited that it all worked and wasn't reliant on some random config on my machine :)

Responses inline:

> 1. It wasn't obvious to me how to exit out of the interactive shell, it looks like you need to use `Ctrl+p Ctrl+q`

I'll add a note to the README.  The shortcut you used will detach from the running session, but you can quit it entirely via `ctrl-d` (EOF) or just typing `exit`, same as any shell.

> 2. Creating/updating the container took a long time on my machine - mounting the directory and interactive mode would mostly alleviate this I believe

 The initial creation is going to take a long time no matter what.  Updating it shouldn't be _too_ big of a deal, but the file copy does still take a while.  And you're right, just running with the mounted local directory alleviates that.  Although I was also thinking about it more last night, and I'm going to add a `git pull` option to refresh from GitHub before running.  My thought process there is that as a Calypso developer running the tests I'm not likely to remember or want to refresh the repo every time, so that will at least mean they're always running against the latest master.

> 3. It ran first time without me creating local-docker.json even though the instructions said this is necessary

Hmm...that's weird.  I'm not sure how that worked.  Did you set your NODE_ENV manually inside the interactive run from the container?  I'll think about it some more.

> 4. I can see how this would make CI easy, but my original thoughts was we'd use a local copy of the tests against chrome running remotely in a docker container, what's the benefits of building the entire project in the container?

There are existing containers on Docker Hub that could accomplish that...I can add instructions if that's all people want to do.  The main benefit of putting everything inside the container is probably for running them in a CI scenario.

My thought process through all this has always been tracking along the lines of containing everything in a single container for portability and ease of use for the Calypso developers.  But now after the fact I'm realizing that all of the "ease of use" I've built in comes via the shell script wrappers, which could just as easily automate things from outside the container.  My gut feeling is still that keeping everything inside will lead to fewer random setup problems, but who's to say for sure.

This will be something good to talk over in person next week.

> 5. I noticed CircleCI isn't running on this branch, is this because it's expecting to use Docker now?

No, I only cancelled the test runs for this branch because the default flow won't touch any of my changes, and I didn't want to waste resources on CircleCI.  There were a lot of concurrent builds going on yesterday.

> 6. When running using xvfb (`-x`), is there a way to VNC or see the tests execute?

No, I decided that's not a feature worth putting in v1, I'll add that later.  For now if you want to see the tests execute you just run them locally without Docker.

alisterscott:
I tested this out against the selenium chrome docker container and it works well 👍

PR #463, opened 2017-03-30
Try resetting customizer menu if it doesn't work

alisterscott:
Tries again to set the menu in customizer if it fails


chenchaoyi:
This looks like a retry for specific step inside a flow on intermittent failure. I am wondering if a simple test level retry is more clear ?

alisterscott:
> This looks like a retry for specific step inside a flow on intermittent failure. I am wondering if a simple test level retry is more clear and straightforward ?

We tried this before and it caused difficulty with the way we nest our test.describe and test.it blocks.

I'll merge this for now and investigate later if we can re-try the test retries :)

PR #465, opened 2017-03-31
Enhancement/import-hook

chenchaoyi:
An attempt to import root level hooks in each spec, so lib/after.js could be removed from run command line.

PR #483, opened 2017-04-07
Update/add after hook to mocha opts

hoverduck:
It just occurred to me that there's already a feature of mocha that lets you apply command line options to every call (`test/mocha.opts`)...this PR just adds `lib/after.js` to that.  It _technically_ adds it to the beginning of the command line rather than the end where we had it before, but ordering doesn't really matter.

cc: @alisterscott 
h/t @chenchaoyi for the improvement idea to start with (#465)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ The local configurations are excluded from the repository, in order to prevent a
 
 ### To run an individual spec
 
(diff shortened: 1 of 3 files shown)


alisterscott:
This is awesome, thanks @hoverduck and @chenchaoyi

PR #488, opened 2017-04-11
Add first WooCommerce e2e tests

alisterscott:
This adds the very first WooCommerce test against Calypso

It just logs on as the new `wooCommerceUser` user and checks My Stats are shown so not actual WooCommerce testing yet.  But this spec can be used from now on to build upon this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ The local configurations are excluded from the repository, in order to prevent a
 
 ### To run an individual spec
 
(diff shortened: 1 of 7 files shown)


alisterscott:
This PR will enabled WooCommerce functionality in `wpcalypso`: https://github.com/Automattic/wp-calypso/pull/12987

alisterscott:
Added a few basic scenarios that make sure all the new routes that are now working in wpcalypso are working.

PR #492, opened 2017-04-12
Use docker-compose to eliminate messy command lines

hoverduck:
This abstracts all of the messy command line options needed to configure Docker-based Selenium runs into a docker-compose file, for both standalone and local Calypso runs.


gziolo:
Nice changes. It works properly for me 👍

PR #516, opened 2017-05-04
Try disabling ad blocker for DNT

### LINKED ISSUE #441: Manage uBlock extension as a git submodule
I'm not a huge fan of the way we have the entire directory structure of the ad blocker extension just sitting in our repo.  It would be cleaner to hold it as a submodule and then add the build steps to circle.yml.  That process wasn't immediately obvious to me, though, so I'm just adding this as an issue to track the work for later.

Note that one thing to consider is making sure the extension directory is contained in CircleCI's cache, so we don't have to rebuild it on every run.

hoverduck:
The Do Not Track flag should now be honored by Calypso, per https://github.com/Automattic/wp-calypso/pull/13269, so this PR removes/disables the uBlock extension.  If we see an increase in timeout failures we should revert this and re-investigate.  The fact that the tests are auto-retried in CI now may also mask the problem, so we need to be vigilant.

@alisterscott - Was there anything else you're aware of that's still pending before we remove the extension?

Closes #441


alisterscott:
Looks good, thanks for this. Should we re-run the Circle build a few times just to make sure it is okay?

hoverduck:
Thanks for taking a look.

Ran the tests in CI 3 more times this morning with no timeouts or failures.  Merging.

PR #525, opened 2017-05-10
Removed xpath selectors from notifications and post editor sidebar components

zhongruige:
This pull request removes xpath selectors from the following:

- notifications-component
- post-editor-sidebar-component
- posts-page

### DIFF
--- a/lib/components/notifications-component.js
+++ b/lib/components/notifications-component.js
@@ -1,5 +1,6 @@
 import { By as by, until } from 'selenium-webdriver';
 import * as driverHelper from '../driver-helper.js';
+import webdriver from 'selenium-webdriver';
 
 import BaseContainer from '../base-container.js';
 
@@ -15,13 +16,17 @@ export default class NotificationsComponent extends BaseContainer {
 		let text = this.driver.findElement( by.css( '.wpnc__notes' ) ).getText();
 		return text;
 	}
+
 	selectCommentByText( commentText ) {
 		const self = this;
 		const commentSelector = by.css( 'div > div.wpnc__text-summary > div:nth-child(1)' );
-		this.driver.findElement( by.css( 'div.wpnc__excerpt' ) ).getText().then( function( selectedCommentText ) {
-			if( selectedCommentText === commentText ) {
-				return driverHelper.clickWhenClickable( self.driver, commentSelector );
-			}
+		let comments = this.driver.findElements( by.css( 'div.wpnc__excerpt' ) );
+		webdriver.promise.filter( comments, function( comment ) {
+			return comment.getText().then( function( text ) {
+				if( text === commentText ) {
+					return driverHelper.clickWhenClickable( self.driver, commentSelector );
+				}
+			} );
 		} );
 	}
 
--- a/lib/components/notifications-component.js
+++ b/lib/components/notifications-component.js
@@ -21,12 +21,17 @@ export default class NotificationsComponent extends BaseContainer {
 		const self = this;
 		const commentSelector = by.css( 'div > div.wpnc__text-summary' );
 		let comments = this.driver.findElements( by.css( 'div.wpnc__excerpt' ) );
+		this.driver.wait( until.elementLocated( commentSelector ), self.explicitWaitMS, 'Could not locate comment exerpt.' );
 		webdriver.promise.filter( comments, function( comment ) {
 			return comment.getText().then( function( text ) {
-				if( text === commentText ) {
-					return driverHelper.clickWhenClickable( self.driver, commentSelector );
-				}
+				return text === commentText;
 			} );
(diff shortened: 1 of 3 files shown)


lib/components/post-editor-sidebar-component.js
```diff
@@ -90,7 +90,7 @@ export default class PostEditorSidebarComponent extends BaseContainer {
 	}
 
 	publicizeToTwitterAccountDisplayed() {
-		const twitterAccountSelector = By.xpath( `//h5[text()='Twitter']/..//label` );
+		const twitterAccountSelector = By.css( 'span[data-e2e-service="Twitter"]' );
```
alisterscott: this is great 👍

lib/components/notifications-component.js
```diff
@@ -16,8 +16,13 @@ export default class NotificationsComponent extends BaseContainer {
 		return text;
 	}
 	selectCommentByText( commentText ) {
-		const commentSelector = by.xpath( `//div[normalize-space(text())='${commentText}']` );
-		return driverHelper.clickWhenClickable( this.driver, commentSelector );
+		const self = this;
+		const commentSelector = by.css( 'div > div.wpnc__text-summary > div:nth-child(1)' );
+		this.driver.findElement( by.css( 'div.wpnc__excerpt' ) ).getText().then( function( selectedCommentText ) {
+			if( selectedCommentText === commentText ) {
```
alisterscott: what happens where the comment text doesn't match? It seems like this will result in different behaviour and may be confusing to work out what is going on...

zhongruige: Good point! I'll look at handling that case.

hoverduck: I also think this breaks down if the expected comment isn't first in the list.  99 times out of 100 I'm sure it will be, but if multiple copies of the test are running at the same time there could be a timing race that puts them out of order.

zhongruige: Thanks for pointing that out @hoverduck! I've updated the code to use filter instead. It's my first time using it, so let me know if there's any issues with the implementation of it. Running the wp-notifications-spec locally, at least, passed.

zhongruige: @alisterscott I just pushed a change that checks and throws an error if the comment text doesn't match. Let me know what you think!

lib/components/notifications-component.js
```diff
@@ -15,9 +16,23 @@ export default class NotificationsComponent extends BaseContainer {
...
+		const commentSelector = by.css( 'div > div.wpnc__text-summary' );
+		let comments = this.driver.findElements( by.css( 'div.wpnc__excerpt' ) );
+		this.driver.wait( until.elementLocated( commentSelector ), self.explicitWaitMS, 'Could not locate comment exerpt.' );
+		webdriver.promise.filter( comments, function( comment ) {
+			return comment.getText().then( function( text ) {
+				return text === commentText;
+			} );
+		} ).then( function( filteredComments ) {
+			if ( filteredComments.length === 1 ) {
+				return driverHelper.clickWhenClickable( self.driver, commentSelector );
```
hoverduck: This filter just tells me if the comment I'm looking for exists _somewhere_ in the list, and then it clicks on the first one regardless of whether that was the match or not.  Instead of returning `text === commentText` can we return the `comment` WebElement itself, and then use that in the `clickWhenClickable` action?  I haven't tested that to see if it works or not.

BTW, the method I'm using to force this scenario is just modifying the wp-notifications-spec.js file to write the comment with the randomly-generated phrase, but then in the second half of the spec where the other user logs in I override `this.comment` with a value I know exists elsewhere in the list.

zhongruige: Thank you for the suggestions! After our chat early today, I ended up deciding to revert back to the original xpath selector as it was more accurate for selecting the comment text in this case.

lib/pages/posts-page.js
```diff
@@ -19,11 +19,11 @@ export default class PostsPage extends BaseContainer {
 	}
 
 	isPostDisplayed( title ) {
-		const postTitleSelector = By.xpath( `//h4[text()='${ title }']` );
+		const postTitleSelector = By.linkText( `${ title }` );
```
alisterscott: Looks good 👍

lib/pages/posts-page.js
```diff
@@ -19,11 +19,11 @@ export default class PostsPage extends BaseContainer {
...
 		return driverHelper.isElementPresent( this.driver, postTitleSelector );
 	}
 	editPostWithTitle( title ) {
-		const postTitleSelector = By.xpath( `//h4[text()='${ title }']` );
+		const postTitleSelector = By.linkText( `${ title }` );
```
alisterscott: Looks good too 👍

PR #526, opened 2017-05-10
Use run wrapper with ci 2.0 again

hoverduck:
No code changes here from the previous attempt (#521).  The problem last time was that the trigger used to kick off the runs on Calypso deploys wasn't updated to also pass the `-x` flag.


hoverduck:
Server-side change in r155961-wpcom

PR #530, opened 2017-05-11
Added theme switch spec for Jetpack sites

hoverduck:
@alisterscott I finally got time to get involved with the Jetpack tests, and figured I'd start by just porting over a simple one from the main specs directory.  It works great on the sites we have set up on Bluehost and Pressable, but fails on GoDaddy.  I assume because the Jetpack version is out of date over there.

I just wanted to check with you before I upgraded it, though, to make sure it wasn't an older version on purpose.  And I of course just wanted a sanity check on this PR to make sure there's nothing different that you've been doing for the Jetpack specs that I'm missing.

### DIFF
--- a/specs-jetpack-calypso/wp-jetpack-theme-switch-spec.js
+++ b/specs-jetpack-calypso/wp-jetpack-theme-switch-spec.js
@@ -24,7 +24,7 @@ test.before( function() {
 	driver = driverManager.startBrowser();
 } );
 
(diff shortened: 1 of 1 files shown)


alisterscott:
> I just wanted to check with you before I upgraded it, though, to make sure it wasn't an older version on purpose. 

As per Slack, not on purpose, I try to keep them all up to date on latest releases to find new and upcoming issues.

There's a recent change to the Jetpack specs you might need to merge into this to get them to pass on CircleCI

PR #534, opened 2017-05-13
Replaced site-selector-component and editor-page xpaths with css selectors

zhongruige:
Replaced Xpath with CSS selectors in the following files:

- site-selector-component
- editor-page

### DIFF
--- a/lib/components/site-selector-component.js
+++ b/lib/components/site-selector-component.js
@@ -6,7 +6,7 @@ import BaseContainer from '../base-container.js';
 class SiteSelectorComponent extends BaseContainer {
 	constructor( driver ) {
 		super( driver, By.css( '.site-selector-modal' ) );
-		this.firstSiteSelector = By.xpath( "//div[contains(@class, 'site-selector-modal')]//div[contains(@class, 'site-selector__sites')]/*[1]/a" );
+		this.firstSiteSelector = By.css( 'div.site-selector-modal div.site-selector__sites a' );
 	}
 
 	back() {
--- a/lib/pages/editor-page.js
+++ b/lib/pages/editor-page.js
@@ -85,7 +85,7 @@ export default class EditorPage extends BaseContainer {
 		self.driver.wait( driverHelper.elementIsNotPresent( this.driver, '.media-library .notice.is-error' ), 500, 'Upload error message is present' );
 		let imageUploadedSelector = webdriver.By.css( 'img[alt="' + newImageName + '"]' );
 		self.driver.wait( until.elementLocated( imageUploadedSelector ), this.explicitWaitMS, 'Could not locate the uploaded image in the media library' );
-		return driverHelper.clickWhenClickable( self.driver, by.css( 'button[data-e2e-button="accept"]' ) );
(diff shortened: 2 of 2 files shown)


lib/pages/editor-page.js
```diff
@@ -104,7 +104,7 @@ export default class EditorPage extends BaseContainer {
 
 	dismissMediaModal() {
 		const driver = this.driver;
-		const cancelSelector = webdriver.By.xpath( "//div[contains(@class, 'editor-media-modal')]//span[contains(@class, 'dialog__button-label') and contains(text(), 'Cancel' )]" );
+		const cancelSelector = webdriver.By.css( 'div.dialog__action-buttons > button:nth-child(1) > span' );
```
alisterscott: This works, and is better than the selector which requires on English, but the `nth-child(1)` seems a bit hacky as it's really just the first child, so this would also work (`div.dialog__action-buttons button`) which is the first button.

What do you think about adding a data-e2e- attribute to this button?

lib/components/site-selector-component.js
```diff
@@ -6,7 +6,7 @@ import BaseContainer from '../base-container.js';
 class SiteSelectorComponent extends BaseContainer {
 	constructor( driver ) {
 		super( driver, By.css( '.site-selector-modal' ) );
-		this.firstSiteSelector = By.xpath( "//div[contains(@class, 'site-selector-modal')]//div[contains(@class, 'site-selector__sites')]/*[1]/a" );
+		this.firstSiteSelector = By.css( 'div.site-selector-modal div.site-selector__sites a' );
```
alisterscott: This works well 👍

lib/pages/editor-page.js
hoverduck: Yeah, I'd vote either add a data attribute or use `div.dialog__action-buttons button:not(.is-primary)` over the nth-child() bit

zhongruige: Thanks for the suggestions!

I ended up going with the data-e2e attribute, which is now in a PR over on the calypso project,
 [here](https://github.com/Automattic/wp-calypso/pull/14075).

I'll be pushes changes shortly that use this new attribute.

alisterscott:
Looks like there's some references to `accept` still which is why it's failing

lancewillett:
@zhongruige As a tip, in GitHub if you drop a bare link to another issue or pull request (not HTML formatted) it'll auto-link it and cross-post it. A reference is added automatically in the other place, which is handy.

zhongruige:
@lancewillett That is handy, thanks for the tip!

zhongruige:
@alisterscott That accept one is interesting, the tests for me pass locally (without any changes), so I'll have to dig into what might be causing them to fail.

zhongruige:
@alisterscott I solved the issue with the `accept` as well as the confirms, which required a change on Calypso (so the CircleCI checks above will still fail on the confirm). Once the changes are pushed then those tests should pass. Here's the PR for reference:

zhongruige:
Hi there! Just wanted to make a quick follow up on this, since it looks like it's good to be merged. Let me know if you have any other questions on it :)

hoverduck:
Thanks @zhongruige!

zhongruige:
@hoverduck Thanks for all the help along the way :)

Issue #535, opened 2017-05-14
Delete comment step in notifications spec not consistently deleting the comment


rachelmcr:
The notifications spec ends with a step to delete the comment that was just left on the test site: https://github.com/Automattic/wp-e2e-tests/blob/master/specs/wp-notifications-spec.js#L98

However, even though that test is passing, the comments aren't being deleted. The test site has pending comments from this test — currently, the oldest pending comment is from 2017/03/16 at 10:02 pm.

I did a manual test, and I believe the issue is that the "Comment trashed" message appears on the notification before the delete request actually occurs. If you close the browser window immediately after deleting the comment, it isn't actually deleted. If you leave the browser window opens for a couple seconds, the comment is deleted as expected.

alisterscott:
Thanks for investigating this.

> I did a manual test, and I believe the issue is that the "Comment trashed" message appears on the notification before the delete request actually occurs.

Do you think this is a wp-calypso issue? We can add a delay before closing the window to make sure the request happens during test runs.

alisterscott:
> Do you think this is a wp-calypso issue?

On a closer look I think it's waiting for the UNDO option to be unavailable to actually call the service to delete the comment.

I made a change to do this in #538 but on a fresh post I am still seeing some comments not deleted, so maybe it still needs to wait a couple of seconds even after UNDO disappearing for the request to be fully sent. It's a tricky one.

PR #536, opened 2017-05-14
Add e2e test for leaving reader comment

### LINKED ISSUE #477: Add e2e test for leaving reader comment

rachelmcr:
This PR fixes #477:

- It updates the existing ReaderPage class with current classes and reader behavior.
- It also adds a new reader spec to test leaving a comment from the reader.

As with the current notifications spec, the comment may not necessarily be deleted from the test site even if the notification confirms it was deleted (and the final test passes). We'll need to address that issue for both specs per https://github.com/Automattic/wp-e2e-tests/issues/535.

### DIFF
--- a/lib/pages/reader-page.js
+++ b/lib/pages/reader-page.js
@@ -12,24 +12,31 @@ export default class ReaderPage extends BaseContainer {
 		}
 		super( driver, by.css( '.is-section-reader' ), visit, url );
 	}
+
 	siteOfLatestPost() {
-		return this.driver.findElement( by.css( 'article.reader__card .site__domain' ) ).getText();
+		return this.driver.findElement( by.css( '.reader-visit-link' ) ).getAttribute('href').then( ( href ) => {
+			return href.split('/')[2];
+		} );
 	}
+
 	likeLatestPost() {
-		return driverHelper.clickWhenClickable( this.driver, by.css( 'article.reader__card .like-button' ) );
+		return driverHelper.clickWhenClickable( this.driver, by.css( '.reader-post-card .like-button' ) );
 	}
+
 	latestPostIsLiked() {
-		return this.driver.findElement( by.css( 'article.reader__card .like-button' ) ).getAttribute( 'class' ).then( ( classNames ) => {
+		return this.driver.findElement( by.css( '.reader-post-card .like-button' ) ).getAttribute( 'class' ).then( ( classNames ) => {
 			return classNames.includes( 'is-liked' );
 		} );
 	}
+
 	latestPostTitle() {
-		return this.driver.findElement( by.css( 'article.reader__card .reader__post-title' ) ).getText();
+		return this.driver.findElement( by.css( '.reader-post-card__title' ) ).getText();
 	}
+
 	commentOnLatestPost( comment ) {
 		driverHelper.clickWhenClickable( this.driver, by.css( '.comment-button' ) );
 		driverHelper.setWhenSettable( this.driver, by.css( '.comments__form textarea' ), comment );
 		driverHelper.clickWhenClickable( this.driver, by.css( '.comments__form button' ) );
-		return this.driver.wait( until.elementLocated( by.css( '.comment__moderation' ) ), this.explicitWaitMS, 'Could not see the comment moderation message after submitting a new comment' );
+		return this.driver.wait( until.elementLocated( by.css( '.comments__comment-moderation' ) ), this.explicitWaitMS, 'Could not see the comment moderation message after submitting a new comment' );
 	}
 }
--- a/lib/pages/reader-page.js
+++ b/lib/pages/reader-page.js
@@ -14,8 +14,8 @@ export default class ReaderPage extends BaseContainer {
 	}
 
 	siteOfLatestPost() {
-		return this.driver.findElement( by.css( '.reader-visit-link' ) ).getAttribute('href').then( ( href ) => {
-			return href.split('/')[2];
+		return this.driver.findElement( by.css( '.reader-visit-link' ) ).getAttribute( 'href' ).then( ( href ) => {
+			return href.split( '/' )[2];
 		} );
 	}
 
@@ -37,6 +37,9 @@ export default class ReaderPage extends BaseContainer {
 		driverHelper.clickWhenClickable( this.driver, by.css( '.comment-button' ) );
 		driverHelper.setWhenSettable( this.driver, by.css( '.comments__form textarea' ), comment );
 		driverHelper.clickWhenClickable( this.driver, by.css( '.comments__form button' ) );
+	}
+
+	waitForModeratedCommentToAppear() {
 		return this.driver.wait( until.elementLocated( by.css( '.comments__comment-moderation' ) ), this.explicitWaitMS, 'Could not see the comment moderation message after submitting a new comment' );
 	}
 }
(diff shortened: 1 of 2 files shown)


specs/wp-reader-spec.js
```diff
@@ -0,0 +1,76 @@
...
+	this.timeout( startBrowserTimeoutMS );
+	driver = driverManager.startBrowser();
+} );
+
+test.describe( 'Reader: (' + screenSize + ') @parallel', function() {
+	this.bailSuite( true );
+	this.timeout( mochaTimeOut );
+
+	test.before( function() {
+			driverManager.clearCookiesAndDeleteLocalStorage( driver );
```
alisterscott: This is indented slightly too much

specs/wp-reader-spec.js
```diff
@@ -0,0 +1,76 @@
...
+		} );
+
+	test.describe( 'Log in as commenting user', function() {
+		test.it( 'Can log in as commenting user', function() {
+			const commentingUser = 'e2eflowtestingcommenter';
```
alisterscott: I don't think this line is necessary

specs/wp-reader-spec.js
```diff
@@ -0,0 +1,76 @@
...
+			this.loginFlow = new LoginFlow( driver, 'commentingUser' );
+			return this.loginFlow.login();
+		} );
+
+		test.describe( 'Leave a comment on the latest post in the Reader', function() {
+			test.it( 'Can see the Reader stream', function() {
+				this.readerPage = new ReaderPage( driver );
+				return this.readerPage.waitForPage();
```
alisterscott: This reads well as it's clear what it's doing, but it's not necessary per se, as creating a new instance of a page will automatically wait for that page. You can still leave this for readability if you like 😊

specs/wp-reader-spec.js
```diff
@@ -0,0 +1,76 @@
...
+			} );
+
+			test.it( 'The latest post is on the expected test site', function () {
+				const testSiteForNotifications = config.get( 'testSiteForNotifications' );
+				return this.readerPage.siteOfLatestPost().then( ( siteOfLatestPost ) => {
+					assert.equal( siteOfLatestPost, testSiteForNotifications, 'The site feed is not the expected test site' );
```
alisterscott: This is a great fail safe check 👍

specs/wp-reader-spec.js
```diff
@@ -0,0 +1,76 @@
...
+				} );
+			} );
+
+			test.it( 'Can comment on the latest post', function() {
+				this.comment = dataHelper.randomPhrase();
+				return this.readerPage.commentOnLatestPost( this.comment );
```
alisterscott: It comes down to style, but it might be worth splitting out the leaving the comment from checking the comment is shown pending moderation. That way the test is clearer what it is doing.

specs/wp-reader-spec.js
```diff
@@ -0,0 +1,76 @@
...
+					return this.loginFlow.login();
+				} );
+
+				test.it( 'Can delete the new comment', function() {
+					this.navBarComponent = new NavbarComponent( driver );
+					this.navBarComponent.openNotifications();
+					this.notificationsComponent = new NotificationsComponent( driver );
+					this.notificationsComponent.selectCommentByText( this.comment );
+					this.notificationsComponent.trashComment();
+					return this.notificationsComponent.waitForUndoMessage();
```
alisterscott: See possible fix for this in #538

specs/wp-reader-spec.js
```diff
@@ -0,0 +1,76 @@
...
+				} );
+			} );
+
+			test.it( 'Can comment on the latest post', function() {
+				this.comment = dataHelper.randomPhrase();
+				return this.readerPage.commentOnLatestPost( this.comment );
+			} );
+
+			test.describe( 'Delete the new comment', function() {
+				test.it( 'Can log in as test site owner', function() {
```
hoverduck: Can you add a `test.before()` call to `driverManager.clearCookiesAndDeleteLocalStorage( driver );` here before logging int as the other user to make sure we don't get any cross-user contamination?

specs/wp-reader-spec.js
rachelmcr: Got it! I think I'll leave this — it's a good point that it isn't necessary, but I like it for readability.

rachelmcr:
My time with Xcode has made me forgetful about eslint, oops. Thanks for the reminder!

I've addressed most of the feedback above with my latest changes. Once #538 is merged I can add `waitForUndoMessageToDisappear()` to fix deleting the comment at the end.

alisterscott:
#538 is now merged if you would like to rebase and incorporate those changes here 👍

rachelmcr:
Ok, those changes are incorporated. Ready for another review!

rachelmcr:
After discussing with @hoverduck I'm going to make sure this test passes in wpcalypso and/or a local Calypso install before merging.

However, the test is currently failing in both wpcalypso and production due to this issue (the comment disappears from the reader when held in moderation): https://github.com/Automattic/wp-calypso/issues/14781

There's some discussion about how to resolve that issue, so I'm going to hold off on this test a bit longer to see how that's resolved — once resolved, I'll make any necessary adjustments to that step in the test and make sure it's passing in both environments.

rachelmcr:
@hoverduck @alisterscott I've added a workaround for Automattic/wp-calypso#14781 (and opened a new issue to track that), and now this test is passing. Would you mind doing another review on this?

hoverduck:
LGTM

lib/pages/reader-page.js
```diff
@@ -12,24 +12,34 @@ export default class ReaderPage extends BaseContainer {
 		}
 		super( driver, by.css( '.is-section-reader' ), visit, url );
 	}
+
 	siteOfLatestPost() {
-		return this.driver.findElement( by.css( 'article.reader__card .site__domain' ) ).getText();
+		return this.driver.findElement( by.css( '.reader-visit-link' ) ).getAttribute( 'href' ).then( ( href ) => {
+			return href.split( '/' )[2];
```
alisterscott: This works and I don't think it would be fragile, but it can be a little hard to understand what it's doing at a glance.
You could try something like:
`return URL.parse( href ).hostname;`
(which requires this also `import URL from 'url';`) which is the in built URL parsing library: https://www.w3schools.com/nodejs/nodejs_url.asp
Saying that, I've probably written very similar or the same split code as you previously so I'm happy with that too

lib/pages/reader-page.js
```diff
@@ -12,24 +12,34 @@ export default class ReaderPage extends BaseContainer {
...
 	commentOnLatestPost( comment ) {
 		driverHelper.clickWhenClickable( this.driver, by.css( '.comment-button' ) );
 		driverHelper.setWhenSettable( this.driver, by.css( '.comments__form textarea' ), comment );
-		driverHelper.clickWhenClickable( this.driver, by.css( '.comments__form button' ) );
-		return this.driver.wait( until.elementLocated( by.css( '.comment__moderation' ) ), this.explicitWaitMS, 'Could not see the comment moderation message after submitting a new comment' );
+		return driverHelper.clickWhenClickable( this.driver, by.css( '.comments__form button' ) );
+	}
+
+	waitForModeratedCommentToAppear() {
+		return this.driver.wait( until.elementLocated( by.css( '.comments__comment-moderation' ) ), this.explicitWaitMS, 'Could not see the comment moderation message after submitting a new comment' );
```
alisterscott: We have a driverHelper method which does something similar and also caters for elements which are located but not hidden on screen, see: `driverHelper.waitTillPresentAndDisplayed(`

rachelmcr:
Thanks for the reviews! I made those last changes to the Reader page (really appreciate that feedback, @alisterscott) and the CI check passed, so I'll merge this one.

PR #538, opened 2017-05-15
Fix notifications spec trashing the comment

### LINKED ISSUE #535: Delete comment step in notifications spec not consistently deleting the comment

alisterscott:
As @rachelmcr pointed out in #535, closing the browser quickly after clicking trash doesn't actually send the request to delete the comment.

This PR waits until the 'Undo' button disappears so that the request is sent (rather than waiting for an arbitrary time).

Fixes #535

### DIFF
--- a/specs/wp-notifications-spec.js
+++ b/specs/wp-notifications-spec.js
@@ -58,6 +58,7 @@ test.describe( 'Notifications: (' + screenSize + ') @parallel', function() {
 
 			test.it( 'Can leave a comment', function() {
 				this.comment = dataHelper.randomPhrase();
+				console.log( this.comment );
 				return this.viewPostPage.leaveAComment( this.comment );
 			} );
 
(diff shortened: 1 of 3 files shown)


alisterscott:
@rachelmcr: would you mind 👀 at this to see whether it fixes #535?

specs/wp-notifications-spec.js
```diff
@@ -98,7 +98,8 @@ test.describe( 'Notifications: (' + screenSize + ') @parallel', function() {
 					test.it( 'Can delete the comment', function() {
```
hoverduck: As long as you're in here can you update this step description to something like "Can delete the comment (and wait for UNDO grace period)" so there's some indication for anyone watching why this step is taking so long?

PR #541, opened 2017-05-18
Go back to CircleCI 1.0

hoverduck:
The performance on CircleCI 2.0 today has been terrible, leading to a crazy ridiculous number of timeout failures.  This PR just temporarily points us back at 1.0 for better stability.  We lose the personal control over Chrome/driver versions, but that's not a big deal.


hoverduck:
The CircleCI run of this branch was better, but did still have some timeout failures.  So this may not be the solution we're looking for, but I merged it anyway to get some more runtime against it.

hoverduck:
Didn't help.  Reverting.

PR #554, opened 2017-05-25
Remove step-header class dependency

### LINKED ISSUE #553: Updated step-header class name broke tests
Some tests depend on the `step-header` class to know whether components have been mounted. The class was updated in https://github.com/Automattic/wp-calypso/pull/14341, which [broke tests](https://circleci.com/gh/Automattic/wp-e2e-tests-canary/1515).

The changes in https://github.com/Automattic/wp-calypso/pull/14341 appear correct, a solution needs to be found for testing (update class to `formatted-header` or check for the presence of a more relevant element).

sirreal:
Use `.step-wrapper` over `.step-header` for test. `.step-header` will be removed.

Fixes #553

### DIFF
--- a/lib/pages/signup/start-page.js
+++ b/lib/pages/signup/start-page.js
@@ -24,7 +24,7 @@ export default class StartPage extends BaseContainer {
 
 		let ABTestControlFlow = flow !== '' ? flow : 'main';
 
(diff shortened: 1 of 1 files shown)


hoverduck:
@sirreal Looks good to me, thanks for the quick fix!  I'm merging it now.

PR #556, opened 2017-05-25
Force redirect to /read for Reader

hoverduck:
With https://github.com/Automattic/wp-calypso/pull/14070 a new feature has been added to Calypso that remembers the last place you were and redirects you there when you visit `/`.  It's currently only active on development and wpcalypso, but will eventually move to production.

It breaks a few of our tests that expect the Reader to load.   This PR fixes that by updating the `ReaderPage` object to visit `/read` instead of just `calypsoBaseURL`.  This also required updating a couple of the login flows to force a visit to the Reader (via that URL) instead of just assuming that it would load upon login.

### DIFF
--- a/lib/flows/login-flow.js
+++ b/lib/flows/login-flow.js
@@ -58,7 +58,7 @@ export default class LoginFlow {
 	loginAndStartNewPost() {
 		this.login();
 
-		let readerPage = new ReaderPage( this.driver );
+		let readerPage = new ReaderPage( this.driver, true );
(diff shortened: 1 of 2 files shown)


hoverduck:
@alisterscott There's no rush on this, I don't think the feature will be turned on in production any time in the immediate future.  But I'd appreciate your eyes on it when you get back from AFK to make sure there's nothing I've missed.

hoverduck:
I went ahead and merged this since it was holding up some of my local testing.

PR #559, opened 2017-06-05
Removed some unnecessary waits to match current functionality

hoverduck:
Delta made some updates to the Customizer, and that meant we were able to remove some of the conditional waiting that we had been doing.


lancewillett:
Nice. 👍

Issue #586, opened 2017-06-23
Workaround: Skipping Jetpack upgrade user invite flow


hoverduck:
Skipping the "Upgrade Contributor -> Author" test - https://github.com/Automattic/wp-calypso/issues/15456

PR #587, opened 2017-06-23
Run invite tests against Jetpack sites

hoverduck:
With https://github.com/Automattic/wp-calypso/pull/15357 we now support the Calypso invite flow on Jetpack sites.

This PR adds support for that.  The only two tests not run at the moment are the Private site Viewer (because we don't have a Jetpack site configured for that yet) and the Contributor->Author test (because #586)


hoverduck:
Note that this PR shows 0 files changes because I accidentally committed directly to `master` first.  Oops.  But I wanted to have a PR on the books for documentation.

PR #591, opened 2017-06-27
Added support for new wpcom login page

hoverduck:
There's a new WordPress.com login page in the works, and this PR adds support for either version.


alisterscott:
Great work with the or selector.
Once we have a single login page and the old is deprecated we could reduce it down to one.
Raised as #594

PR #595, opened 2017-06-28
Add AB test postPublishPreview

hoverduck:
@alisterscott - I had a little trouble with this one.  This must be the first AB test we've had active on the Editor that required an override.  It took me a little by surprise that the localStorage value wasn't being overridden...I didn't realize until I dug in that the `setABTestControlGroupsInLocalStorage()` function is only called in a couple specific locations (signup flow and the Jetpack authorize page at the moment), I had just assumed since the function lived in `base-container.js` that it was run everywhere.

Is there a reason for not just calling that after every login?

EDIT:  That's exactly what I've done in this PR, because it ended up being the simplest solution.  I tried adding it in the EditorPage, and at other specific spots, but ran into various issues.  Offhand I think this is likely a safe solution, but I'm curious to hear @alisterscott's opinion


hoverduck:
I'm still seeing some timeout issues with loading the AB tests before login, so I've semi-reverted this PR in #596

specs/wp-page-editor-spec.js
```diff
@@ -281,7 +281,6 @@ test.describe( `[${host}] Editor: Pages (${screenSize})`, function() {
 				this.editorPage.enterTitle( pageTitle );
 				this.postEditorSidebarComponent = new PostEditorSidebarComponent( driver );
 				this.postEditorSidebarComponent.setVisibilityToPasswordProtected( postPassword );
-				this.editorPage = new EditorPage( driver );
```
alisterscott: I think this is actually needed for mobile as it switches context between the sidebar screen and the editor screen

specs/wp-post-editor-spec.js
```diff
@@ -458,7 +458,6 @@ test.describe( `[${host}] Editor: Posts (${screenSize})`, function() {
 				this.editorPage.enterTitle( blogPostTitle );
 				this.postEditorSidebarComponent = new PostEditorSidebarComponent( driver );
 				this.postEditorSidebarComponent.setVisibilityToPasswordProtected( postPassword );
-				this.editorPage = new EditorPage( driver );
```
alisterscott: Same as above

lib/flows/login-flow.js
```diff
@@ -65,6 +65,9 @@ export default class LoginFlow {
 		}
 
 		loginPage = new LoginPage( this.driver, true );
+		this.driver.getCurrentUrl().then( ( urlDisplayed ) => {
+			return loginPage.setABTestControlGroupsInLocalStorage( urlDisplayed );
```
alisterscott: This doesn't work because the (older) login page is on a separate domain (wordpress.com not wpcalypso.wordpress.com) - and the ab test keys in localstorage won't be set for wpcalypso.wordpress.com

PR #596, opened 2017-06-28
Ignore the showPostPreview ABTest and handle both cases for now

hoverduck:
I ran into timeouts and other issues with the solution I merged in #595.  This PR is half revert, half workaround/fix.  The changes here handle either case of the AB test, whether the post-publish preview is shown or not.  But it got complicated when dealing with the Page publishing flow that publishes and then previews.  So for now I've just skipped that piece of the flow so we can have a functioning set of tests when I leave for the day.

@alisterscott I can work on a real fix for this tomorrow, but if you have time and want to tackle it go right ahead 😄

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -74,7 +74,6 @@
 	[ "signupSurveyStep_20170329", "hideSurveyStep" ],
 	[ "readerPostCardTagCount_20170315", "showThree" ],
 	[ "signupProgressIndicator_20170612", "original" ],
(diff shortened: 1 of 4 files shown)


hoverduck:
And this solution isn't working either.  Went fine locally, but something happened on CI.

alisterscott:
Issues with setting AB test have been resolved so this isn't necessary. Closing

Issue #494, opened 2017-04-13
Project Idea: schedule i18n screenshots once a week


alisterscott:
The i18n screenshots were created as a once-off thing to generate a bunch of files that are manually reviewed.

These are out of date now, but what if we got them up to date and then ran say regularly, say once a week, and had it so they fully automated the process of uploading them as a gallery to a site (using WordPress.com REST API?) so that we could historically see the changes and easily send them out for review by translators.

This would be a good meetup project or GM team activity.

alisterscott:
@deBhal I've put this here to keep track of it

deBhal:
This is a great idea! The screenshots were really useful, and we found a bunch of problems going through them.

That said, we (i18n) also need to figure out how to consume them. I think we should have a plan to avoid having them sit there collecting virtual dust before we implement this.

Maybe we could include them in the happiness i18n rotations.

lancewillett:
Closing as this project is complete — thanks for driving this one @rachelmcr. 🎉

PR #602, opened 2017-06-29
Update i18n Screenshot Flows

rachelmcr:
This PR is the first step for #494, to make the following changes:

- Switch from using a separate `specs-i18n` directory to select flows from the main `specs` repo for i18n screenshots (so the screenshot flows don't have to be updated separately).
- Take screenshots of each test in the two canary signup flows, in 6 locales (en, pt-br, es, ja, fr, he), on two device sizes (mobile and desktop).
- Save screenshots in a `/screenshots-i18n` directory, grouped by locale/screensize/flow.

To test, use the command `./run.sh -i` and view the output in the `/screenshots-i18n` directory.

(Note: The old screenshot code run from the `specs-i18n` directory still needs to be cleaned up. I will do that cleanup in a separate PR as a followup to this one.)


lib/reporter/magellan-i18n-reporter.js
```diff
@@ -0,0 +1,206 @@
+const BaseReporter = require( 'testarmada-magellan' ).Reporter;
```
hoverduck: Rather than maintain a separate reporter for the i18n screenshots can we just set an envvar like `SCREENSHOTDIR` and have it default to `/screenshots` and just override it here to `screenshots-i18n`?

rachelmcr: Yeah, I originally was thinking about making more changes to the reporter but given that it's just this change I'll go ahead and do that.

hoverduck:
I doubt this is related to your code, but likely just a fluke on my end as I was doing other stuff on my computer while the tests were running...but a handful of the screenshots came out looking like this.  Have you seen anything like this in your testing?

rachelmcr:
That's odd — I haven't seen that in my testing at all. If it doesn't turn out to be a fluke on your end I can definitely investigate.

hoverduck:
I tested it out and everything seems to work as advertised 😄 

Is the plan to stick with the `@canary`-tagged signup tests, or should we add a new tag `@i18n` to hit other flows as well?

rachelmcr:
I thought about adding a new tag, but I think it makes more sense to stick with `@canary` tests. The same flows that are selected as important enough to be used as canaries for general e2e testing are most likely to be ones that are useful for i18n review.

If there's a compelling reason in the future to add i18n review for a non-canary flow we can always revisit this, but for now I think the `@canary` tag is a good way to identify flows that should be prioritized for i18n review.

rachelmcr:
@hoverduck I've added the `SCREENSHOTDIR` envvar as suggested — ready for another review. :)

lib/media-helper.js
```diff
@@ -84,9 +84,6 @@ export function writeScreenshot( data, prefix, i18nScreenshot ) {
 	}
 
 	let directoryName = 'screenshots';
-	if ( i18nScreenshot === true ) {
```
hoverduck: Is there a reason you removed this?  And can we reference the envvar here as well just to make sure it's always in sync?

rachelmcr: I removed this because it only applies to the temp directory used while the test runs. No matter what temp directory is used, it will be moved to the directory defined in the envvar at the end of the run. We could reference the envvar here if we want the temp directory to always match the final one, so I can add that back in.

hoverduck: Ah, right right.  That makes sense.  Not necessary, but I think it's probably worth putting the envvar here too just for consistency.  And in case we ever run the i18n tests locally with just mocha and not magellan it'll make things less confusing.

rachelmcr: Consistency FTW. It was a quick change, so now we'll use the envvar here too.

PR #605, opened 2017-07-03
Remove if host===WPCOM from Jetpack theme switch spec

hoverduck:
Depends on https://github.com/Automattic/wp-calypso/pull/13621


alisterscott:
Re-tested and LGTM

PR #610, opened 2017-07-06
Update the selectors for inserting media and contact forms to use newly added data-e2e- attributes

alisterscott:
Requires https://github.com/Automattic/wp-calypso/pull/15894 to be merged


alisterscott:
I have tested this locally with the `wp-calypso` changes but this will have to be re-run in CircleCI when those changes land

PR #615, opened 2017-07-10
New selector for editor sidebar

hoverduck:
https://github.com/Automattic/wp-calypso/pull/16022 Updated the CSS class name on the editor sidebar, which broke most of the editor tests.  This PR brings them back in sync.

h/t: @rachelmcr for the quick identification of the problem

### DIFF
--- a/lib/components/post-editor-sidebar-component.js
+++ b/lib/components/post-editor-sidebar-component.js
@@ -4,7 +4,7 @@ import BaseContainer from '../base-container.js';
 
 export default class PostEditorSidebarComponent extends BaseContainer {
 	constructor( driver ) {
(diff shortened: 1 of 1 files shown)


brandonpayton:
Hi, my apologies for the breakage. Is there a good way I can avoid surprising you in the future?

hoverduck:
@brandonpayton No worries.  We're still working on improving the communication flow between the Calypso devs and the e2e tests.  At some undefined point in the future we'll have a way to automatically run the tests against your branch early so this kind of thing doesn't happen....but for now the best option is to just throw a note in the `#e2e-testing-discuss` channel in Slack to let us know if there's going to be a major CSS selector change.

Or you can always ping us in the Calypso PR, or create a new issue here in the e2e repo as well.

Issue #434, opened 2017-03-17
Canary tests fail on wpcalypso.wordpress.com but pass on production


alisterscott:
This is to keep track of the canary test failures running against wpcalypso.wordpress.com even though they are passing on production

cc: @hoverduck

alisterscott:
Could it be that wpcalypso.wordpress.com is a slower environment?

alisterscott:
Another possibility is back-to-back wpcalypso deploys cause existing running tests to fail

hoverduck:
> Could it be that wpcalypso.wordpress.com is a slower environment?
This is true, so I tried bumping up the `explicitWaitMS` from 20 to 60s, but that didn't seem to make a difference (I've since returned it to the default).

What I'm seeing now that we have screenshots to Slack enabled (also via VNC) is that frequently in the signup process we're getting dumped on the login screen instead of the checkout page.  The redirect in the URL is pointing back to checkout.

We're not seeing it on production, so I'm not sure yet what the problem is.  The investigation continues.

hoverduck:
I wasn't able to reproduce this problem locally, either manually or with the scripts.  Most of my time was spent futzing with CircleCI VNC problems, and had to keep restarting the session.

The only error I was able to track down was one time I saw a post-message failure from a wordpress.com iframe to wpcalypso, which I'm guessing may have had something to do with the oauth token.  Then once that's gone screwy it thinks we're not logged in, and dumps us back on the login page instead of checkout.

But like I said, I wasn't able to reproduce it locally to confirm any of that.  I'll dig in again and ping the Calypso folks on Monday, but we could also add a workaround to the signup flow to detect the login page and just re-authenticate.  If it happens in production that's what the end user would do anyway.

hoverduck:
I've added a workaround in https://github.com/Automattic/wp-e2e-tests/pull/444, and asked the Calypso devs on p4TIVU-6cx-p2

I'm going to leave this issue open and re-label it as workaround

hoverduck:
This bug is rearing its head more frequently, and the workaround from #444 isn't working consistently anymore.  I assume it's due to the work on the new login screen, but this screenshot (pulled from the Slack alert) looks like the `wp-login.php` page, not the new `/log-in`.

![image](https://user-images.githubusercontent.com/7233112/27929682-e1016c22-6261-11e7-8715-d164591ac671.png)

hoverduck:
[...]

PR #616, opened 2017-07-10
New attempt at canary signup workaround

hoverduck:
Now instead of matching `/wp-login.php|log-in/` I'm just checking for `! /checkout/`.  My theory is maybe the login page hasn't actually loaded in the time it takes for the URL check, so we're passing the test too early.

This new method may lead to unnecessary redirects, but that _should_ be inconsequential to the test results themselves, but something to monitor.

I'm having problems recreating the issue in manual CircleCI runs, so I'll probably just merge this and let it run its course for a while.


hoverduck:
Ref #614 #613 #434

PR #619, opened 2017-07-11
Enable site preview test for Pressable (https)

hoverduck:
Working on #618 made me realize I could turn this test on for Pressable as well, since it has https configured.


lancewillett:
Minor note: we can call these "Atomic" now because they may or may not be always hosted on Pressable.

hoverduck:
For the purposes of this test it doesn't matter whether it's an AT/Atomic site (and I don't think this one is), it's just a regular Jetpack site running on Pressable.

Issue #505, opened 2017-04-24
Move Docker image from /hoverduck to /Automattic on hub


hoverduck:
The Docker image used by CircleCI is currently hosted in my personal repo on hub.docker.com.  Need to create an A8c organization on there and move it over so more people have write permissions to it.

PR #621, opened 2017-07-12
Update node to 6.11.1

alisterscott:
To match wp-calypso


alisterscott:
Actually I think we are now using docker images and these reside under your account @hoverduck? Can we move them to a organisational account so we can all update them? Can I get rid of `circle.yml`?

hoverduck:
> Can we move them to a organisational account so we can all update them?

@alisterscott Yeah, sorry...I've had this on my list for a while now, but haven't gotten around to it. - https://github.com/Automattic/wp-e2e-tests/issues/505

I've made the necessary updates here in the Dockerfile, pushed a new version to Docker Hub, and updated the CircleCI 2.0 config file to use the new version.

I also removed the old circle.yml file.  I had been keeping it around as a quick backup in case we wanted to revert away from 2.0, but I think we're past that point now (and 2.0 is officially out of beta as of yesterday too 😄)

PR #625, opened 2017-07-13
TEST ONLY Try wpcalypso to debug canary failures

alisterscott:
**DON'T MERGE**


alisterscott:
@hoverduck I'm using this branch to test out to see if I can work out why the signup tests fail on wpcalypso, interestingly I can get it to fail locally now using this branch but I'm trying to work out what's going on using sleeps and breaks and trying to take over the browser.

Are you interested in having a look? Another set of eyes would be good to see if it's something obvious, adding a revisit on the account page seems to make the problem happen pretty consistently...

hoverduck:
@alisterscott I've spent a good chunk of today playing with this and have the same problems you had, but still no solutions.

I wrote https://github.com/Automattic/wp-calypso/issues/16199 to document the page refresh issue breaking the NUX flow...it's likely related to the bug we're seeing in Canary, but has a different enough signature I felt it warranted its own issue.

PR #626, opened 2017-07-14
Add test to add new WooCommerce product

alisterscott:
Adds a basic product addition test including name, description, category, price, dimensions, weight, quantity and backorder availability.


alisterscott:
@hoverduck: would you mind taking a quick 👀 at this to see if you know of an easy way to upload an image file? I tried setting the hidden `input` field to the file name but it doesn't seem to work (for me)

hoverduck:
@alisterscott It looks like the event to upload the file isn't actually hooked up to the hidden form input element until after the Upload Image button is clicked.  I added a commit that clicks that button first, and the setting the file path uploads the image as expected...

BUT...that then leaves the file upload dialog visible on screen, and as far as I can tell there's no way for WebDriver to dismiss it.  I'm also getting failures on actually 'publishing' the product after setting the image, but that's where I stopped troubleshooting.

Unless you know of a way to dismiss that dialog I think your best bet is to talk to the Woo folks about changing how that upload event is fired.  I did try just doing a `form.submit()` on the input form, but that causes a page refresh that further confuses things.

alisterscott:
Thanks @hoverduck 

I remember now having to add an event to the file field for media uploads in calypso to trigger the actual upload, which means we'd have to do something similar here.

I'll ask one of the Woo folks about adding this similar to `client/my-sites/media-library/upload-button.jsx`

alisterscott:
Workaround in place for saving on mobile screen: #639

PR #627, opened 2017-07-15
Add i18n screenshot upload script

rachelmcr:
This PR is the third step for #494. It adds a script (called from `./run.sh`) to upload i18n screenshots to a WordPress.com blog using the WordPress.com REST API.

## To Test

You will need to create a [REST API app](https://developer.wordpress.com/apps/) and add the follow to your local config:

```
"restApiApplication": {
    "client_id": "YOUR_CLIENT_ID",
    "redirect_uri": "YOUR_REDIRECT_URI",
    "client_secret": "YOUR_CLIENT_SECRET"
  }
```

You will also need to specify a user `i18nScreenshotUser` in your local config, with the site address or ID where you want to upload the screenshots, as follows:

```
"testAccounts": {
	"i18nScreenshotUser": [
	  "USERNAME",
	  "PASSWORD",
	  "SITE_ADDRESS_OR_ID"
	]
}
```

You can then run `./run.sh -i -U` to first generate the i18n screenshots and then upload them to the specified site.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -191,7 +191,7 @@ else # Not using multiple CircleCI containers, just queue up the tests in sequen
   if [ "$CI" != "true" ] || [ $CIRCLE_NODE_INDEX == 0 ]; then
     if [ $UPLOAD == 1 ]; then
       # Clear out screenshots-i18n/ directory
-      rm -f screenshots-i18n/*
+      rm -f ${SCREENSHOTDIR}/*
     fi
     IFS=, read -r -a SCREENSIZE_ARRAY <<< "$SCREENSIZES"
     IFS=, read -r -a LOCALE_ARRAY <<< "$LOCALES"
@@ -212,7 +212,7 @@ fi
 
 if [ $UPLOAD == 1 ]; then
   echo "Uploading i18n tests screenshots:"
-  CMD="node ./scripts/i18nscreenshots.js $LOCALES"
+  CMD="node ./scripts/i18n-screenshots-upload.js $LOCALES"
 
   eval $CMD
   RETURN+=$?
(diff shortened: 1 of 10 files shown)


hoverduck:
I'm excited to have the API/OAuth stuff worked out and here in the repo!  So much possibility.

rachelmcr:
Thanks for the review! Those are great suggestions, and I decided to address them here since they're quick changes.

For the manual runs, I'm clearing the directory contents if the `-U` parameter is specified — I'm not too worried about potentially losing screenshots but at least that way they are only cleared if you're going to do the upload as well. (I'll revisit if we find ourselves regretting that later!)

I'll plan to merge this once the checks are done.

run.sh
```diff
@@ -201,6 +210,14 @@ else # Not using multiple CircleCI containers, just queue up the tests in sequen
   fi
 fi
 
+if [ $UPLOAD == 1 ]; then
+  echo "Uploading i18n tests screenshots:"
+  CMD="node ./scripts/i18nscreenshots.js $LOCALES"
```
hoverduck: You need your new script name here

run.sh
```diff
@@ -184,6 +189,10 @@ if [ $PARALLEL == 1 ]; then
   fi
 else # Not using multiple CircleCI containers, just queue up the tests in sequence
   if [ "$CI" != "true" ] || [ $CIRCLE_NODE_INDEX == 0 ]; then
+    if [ $UPLOAD == 1 ]; then
+      # Clear out screenshots-i18n/ directory
+      rm -f screenshots-i18n/*
```
hoverduck: I think it's probably fine to leave this hard-coded here for now since this is only used for the i18n screenshots, but what do you think about replacing this with the SCREENSHOTDIR envvar?

rachelmcr: Ah, good call.

run.sh
rachelmcr: Agh, forgot to commit after I fixed that. Doing it now.

PR #629, opened 2017-07-18
Add JPO tests for Bluehost

ebinnion:
JPO on Bluehost was broken for several weeks without us knowing. To minimize this happening in the future, I'd like to add functional tests that check that the flow is working.

### DIFF
--- a/lib/pages/bluehost/bluehost-install-wordpress-details-page.js
+++ b/lib/pages/bluehost/bluehost-install-wordpress-details-page.js
@@ -1,0 +1,24 @@
+import { By as by } from 'selenium-webdriver';
+import config from 'config';
+import BaseContainer from '../../base-container';
+import * as driverHelper from '../../driver-helper';
+
+export default class BluehostInstallWordPressDetailsPage extends BaseContainer {
+	constructor( driver, visit = false ) {
+		const url = config.get( 'bluehostInstallWordPressUrl' );
+		super( driver, by.css( 'form.mm-inputs input[name="ss_site_title"]' ), visit, url );
+	}
+
+	submitAdminDetails() {
+		const siteTitle = config.get( 'bluehostWPInstallSiteTitle' );
+		const username = config.get( 'bluehostWPInstallUsername' );
+		const email = config.get( 'bluehostWPInstallEmail' );
+		const password = config.get( 'bluehostWPInstallPassword' );
+
+		driverHelper.setWhenSettable( this.driver, by.css( 'input[name="ss_site_title"]' ), siteTitle );
+		driverHelper.setWhenSettable( this.driver, by.css( 'input[name="ss_admin_user"]' ), username );
+		driverHelper.setWhenSettable( this.driver, by.css( 'input[name="ss_admin_email"]' ), email );
+		driverHelper.setWhenSettable( this.driver, by.css( 'input[name="ss_admin_pass"]' ), password, { secureValue: true } );
+		return driverHelper.clickWhenClickable( this.driver, by.css( 'button[type="submit"].mm-btn' ) );
+	}
+}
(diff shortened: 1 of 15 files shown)


ebinnion:
Hey @hoverduck - Would you mind taking a look at this? I've got the tests working through the point where a new WP install is provisioned. Next is to actually login to the site and start testing JPO.

While this currently works, I'm not a fan of the page names, and I'm not super sure if I'm doing anything silly. I can provide a sample config file if you need it.

lib/pages/bluehost/bluehost-login-page.js
```diff
@@ -0,0 +1,19 @@
...
+import * as driverHelper from '../../driver-helper';
+
+export default class BluehostLoginPage extends BaseContainer {
+	constructor( driver, visit = true ) {
+		const url = config.get( 'bluehostLoginUrl' );
+		super( driver, by.css( 'form.login_form' ), visit, url );
+	}
+
+	logon() {
+		const username = config.get( 'bluehostUsername' );
```
hoverduck: I'm torn on the username/password config options.  They could either go inside the big `bluehostConfig` parameter I suggested in another comment, or inside the `testAccounts` object with the other username/passwords.

If this is just a standalone test it probably makes more sense to keep them with the other Bluehost parameters.  But if we expand to other tests that involve logging in here then we'd probably want to put something in the LoginFlow object, in which case it might make more sense to put them in the `testAccounts`.

But we can burn that bridge when we get there :)

specs-jetpack-calypso/wp-jpo-bluehost-spec.js
```diff
@@ -0,0 +1,54 @@
...
+const startBrowserTimeoutMS = config.get( 'startBrowserTimeoutMS' );
+
+// When setting this in test.before below, it resulted in undefined? Perhaps the before never fired?
+var driver;
+test.before( function() {
+	this.timeout( startBrowserTimeoutMS );
+	driver = driverManager.startBrowser();
+} );
+
+test.describe( 'Can step through JPO flow on Bluehost', function() {
```
hoverduck: If you want this test run with the other Jetpack tests on CI you need to add `@jetpack` to the top-level `.describe()` title so the test-runner will find it.

lib/pages/bluehost/bluehost-install-wordpress-details-page.js
```diff
@@ -0,0 +1,24 @@
...
+	constructor( driver, visit = false ) {
+		const url = config.get( 'bluehostInstallWordPressUrl' );
+		super( driver, by.css( 'form.mm-inputs input[name="ss_site_title"]' ), visit, url );
+	}
+
+	submitAdminDetails() {
+		const siteTitle = config.get( 'bluehostWPInstallSiteTitle' );
+		const username = config.get( 'bluehostWPInstallUsername' );
+		const email = config.get( 'bluehostWPInstallEmail' );
+		const password = config.get( 'bluehostWPInstallPassword' );
```
hoverduck: Maybe move these config settings into a common object for easier access?  Not necessary, but would make the NODE_CONFIG more legible.  It could also include the config parameters you call in your other page objects.

```
"bluehostWPInstallDetails": {
  "url", "abc.d.com",
  "siteTitle": "abc",
  "username": "abc",
  "email": "abc@d.com",
  "password": "zyx"
}
```

And then move the whole thing into the constructor so you can pick up the URL.

ebinnion:
At this point, the suite runs all of the way through. I've occasionally seen a test failure here and there, but not quite sure why that's happening. Seems rare.

I'll plan on refactoring/renaming a bit on Monday and following back after the BH fix deploys.

ebinnion:
Closing for now since the location of JPO has changed and we are also changing the flow significantly. We now monitor JPO connections via an MC stat as well.

PR #649, opened 2017-07-31
Swap paid signup with free signup for @canary

alisterscott:
Due to ongoing issues with running paid sign up tests on wpcalypso.wordpress.com, switching to free signup tests only for the time being


alisterscott:
Undone in https://github.com/Automattic/wp-e2e-tests/pull/653

PR #658, opened 2017-08-14
Adds a data helper for test credit card details

### LINKED ISSUE #155: Create helper function to provide credit card details
Credit card details are used in multiple locations for various signup/payment tests.  They should be pulled out into a helper function and returned as a single object so we only have to update them in one place.  Ref https://github.com/Automattic/wp-e2e-tests/pull/154 for an example

alisterscott:
Fixes #155


alisterscott:
@hoverduck can you please review?

PR #659, opened 2017-08-14
Updated default authURL to /log-in

hoverduck:


alisterscott:
What do you think about a PR that gets rid of auth URL and adds `/log-in` to base url?

alisterscott:
Removed the auth-url in #660

PR #660, opened 2017-08-15
Try removing auth-url and just using base url

alisterscott:
Calypso supports native logon now so this should just work


alisterscott:
@hoverduck A heads-up on this one in case this causes problems anywhere - it shouldn't as I've checked `wpcalypso` and `calypso.localhost` and all appears to work fine

PR #661, opened 2017-08-15
Add support for postPublishConfirmation using feature toggle

alisterscott:
This is defaulted to off for the time being


alisterscott:
@hoverduck this shouldn't have any impact as this will default to be disabled, but a heads-up nonetheless just in case 😉

Issue #650, opened 2017-08-01
Jetpack tests require site on free plan


alisterscott:
The Jetpack tests require a site on a free plan, so that it specs can compare plans etc.

But this has some downsides like Akismet isn't active so the sites get a lot of unfiltered spam.

This is to investigate how we have have sites on Jetpack plans and still cover the plans functionality etc.

hoverduck:
I think we'll need to have a parameter in NODE_CONFIG that details how the current site-under-test is set up to dictate which tests are run.  In #643 I have an update to `default.json` that says whether the site is https or not, and this would fit nicely with that.

alisterscott:
The payment button test introduced in #666 is not currently run on Jetpack sites as this requires a paid Jetpack plan - so when this issue is resolved we should add `@jetpack` to it

PR #666, opened 2017-08-18
Add a payment button e2e test

alisterscott:
This adds an e2e test to add a payment button via the editor and make sure it appears on the front end of the site


alisterscott:
@hoverduck I tried running this against a Jetpack site (bluehost) and it says a newer version of Jetpack was required even though the Jetpack version was the latest. Any ideas why?

hoverduck:
Maybe the UI has changed in the last few days (sorry for the delayed review), but what I'm seeing now is this upgrade nudge:

image

If that's the case then I guess we need to just remove the `@jetpack` tag from this test and make a note in #650 that when we run the tests against non-free sites that we can add this test back in.

alisterscott:
@hoverduck I think your image shows a different version of Jetpack is needed (not a different plan)? Where are you seeing that it needs a non-free Jetpack plan?

hoverduck:
Oh you're right, I didn't even read the text...I just saw it's the standard upgrade nudge banner we use elsewhere.  And clicking the link takes you to the plans page... So either the text on the banner needs to be updated, or it should link to a different place, right?

alisterscott:
Payment buttons do require a Jetpack plan so there's actually two separate issues here. I have removed this test from running on Jetpack sites for now as we don't use plans.

PR #669, opened 2017-08-21
Move localization data into the tree

### LINKED ISSUE #664: Can't find localization_data.json
Can't run search-wordpress-on-google.js test, because

> Error: Cannot find module '../../localization_data.json'

Where I can get an example for this file?
Thanks.

akirk:
Fixes #664 as well as a wrong check for the pt-br WordPress.com homepage.

### DIFF
--- a/lib/pages/wp-home-page.js
+++ b/lib/pages/wp-home-page.js
@@ -36,9 +36,9 @@ export default class WPHomePage extends BaseContainer {
 		return d.promise;
 	}
 	checkURL( culture ) {
-		const target = culture ?
-			localizationData[ culture ].wpcom_base_url :
-			'wordpress.com';
+		const target = culture
+			? localizationData[ culture ].wpcom_base_url
+			: 'wordpress.com';
 		this.driver.getCurrentUrl().then( ( currentUrl ) => {
 			assert.equal( true, currentUrl.includes( target ), `The current url: '${ currentUrl }' does not include ${ target }` );
 		} );
--- a/lib/pages/wp-home-page.js
+++ b/lib/pages/wp-home-page.js
@@ -1,12 +1,13 @@
 import webdriver from 'selenium-webdriver';
 import assert from 'assert';
 import BaseContainer from '../base-container.js';
+import localizationData from '../../localization-data.json';
 
 const by = webdriver.By;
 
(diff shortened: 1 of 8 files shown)


lib/pages/wp-home-page.js
```diff
@@ -35,9 +36,9 @@ export default class WPHomePage extends BaseContainer {
 		return d.promise;
 	}
 	checkURL( culture ) {
-		const target = culture
-			? culture + '.wordpress.com'
-			: 'wordpress.com' ;
+		const target = culture ?
```
rachelmcr: I noticed a couple linting warnings for this line and the line below it — the characters `?` and `:` should be a the beginning of the line.

specs-i18n/search-wordpress-on-google.js
```diff
@@ -64,5 +64,8 @@ function doGoogleAdSearch( search_params ) {
 
 test_data.google_searches.forEach( function( search_params ) {
 	doGoogleAdSearch( search_params );
+	if ( process.env.CIRCLECI === 'true' ) {
+		sleep( 15000 );
```
rachelmcr: `sleep` here isn't defined. If you want to pause after each test, how about using a block like this in the test itself:

```
test.afterEach( function() {
	this.sleep(15000);
} );
```

akirk:
Thanks @rachelmcr for the review. I have incorporated your feedback and rebased, now it's passing.

akirk:
@rachelmcr I've now got it running without running into robot checks from Google, with 10 secs delay between the runs. `run.sh -I` currently gives two failures (for `ar` and `zh-tw`) but that is expected. Can I merge it like this?

One thing that's slightly odd running locally is that it 1) doesn't keep the screenshots for me and 2) when I watch the directory I see a screenshot made for the failure but it's scrolled down. Any ideas for remedy of that?

akirk:
> One thing that's slightly odd running locally is that it 1) doesn't keep the screenshots for me and 2) when I watch the directory I see a screenshot made for the failure but it's scrolled down. Any ideas for remedy of that?

Scratch that. For 1) Rachel pointed out to me that the screenshots are moved to the the `screenshots/` directory, and for 2) it might have been me interfering with the test while I was typing something else that caused the scrolling down (for example the Space key).

Issue #426, opened 2017-03-14
Add test to signup spec for e-mail verification


hoverduck:
1) Signup for a new account (likely just append this to an existing signup test, like for the Free account)
2) Compose a new post
3) Verify the user can't publish, and sees a prompt to verify their e-mail
4) Verify the e-mail
5) Publish the post

alisterscott:
Was the regression that triggered this purely visual? Would an e2e find the problem?

bsessions85:
Added in PR #671

Issue #356, opened 2016-12-29
Add dynamic step numbering to signup spec


hoverduck:
PR #285 added dynamic step numbering to the signup spec, so steps could be added/removed/moved without having to edit every step number further down the line.  But I forgot to merge it, and now it's not worth the hassle to rebase that branch and add the feature back in.  So this issue tracks it as an enhancement, and I'll add the feature back the next time I'm working on the signup spec.

alisterscott:
This would be a nice to have.

bsessions85:
Added in PR #671

PR #671, opened 2017-08-22
Adding steps to check verification before publishing #426

bsessions85:
This is for GitHub issue #426. I added some steps to the existing free account creation flow that tries to create a post, ensures that the publish button is disabled, verifies the email account and then makes sure the publish button is enabled.


specs/wp-signup-spec.js
```diff
@@ -168,14 +173,48 @@ testDescribe( `[${host}] Sign Up  (${screenSize}, ${locale})`, function() {
...
 
-									test.describe( 'Step Eight: Can activate my account from an email', function() {
-										test.before( function() {
-											return this.emailClient = new EmailClient( signupInboxId );
+									test.describe( 'Step Eight: Can not publish until email is confirmed', function() {
+										test.it( 'Can see a disabled publish button', function() {
+											const blogPostTitle = dataHelper.randomPhrase();
+											const blogPostQuote = dataHelper.randomPhrase();
+
+											this.adminTopBar = new AdminTopBar( this.driver );
```
hoverduck: Using `this.driver` here isn't working for me.  If I drop it to just `driver` then it processes fine.

specs/wp-signup-spec.js
```diff
@@ -168,14 +173,48 @@ testDescribe( `[${host}] Sign Up  (${screenSize}, ${locale})`, function() {
...
+											} );
+
+											test.it( 'Can see a single activation message', function() {
+												return this.emailClient.pollEmailsByRecipient( emailAddress ).then( function( emails ) {
+													inboxEmails = emails;
+													return assert.equal( inboxEmails.length, 1, 'The number of invite emails is not equal to 1' );
+												} );
+											} );
+
+											test.describe( 'Step 10: Can publish when email is confirmed', function() {
```
hoverduck: This is the only step in this sequence that's labelled with the digits `10` instead of `Ten`.  Not a big deal, but worth changing for consistency.

If you're ambitious, this would be as good of a time as any to tackle https://github.com/Automattic/wp-e2e-tests/issues/356 as well

specs/wp-signup-spec.js
```diff
@@ -168,14 +173,48 @@ testDescribe( `[${host}] Sign Up  (${screenSize}, ${locale})`, function() {
...
+												test.it( 'Can not see a disabled publish button', function() {
+													this.driver.executeScript( `window.open( '${inboxEmails[0].html.links[0].href}' )` );
```
hoverduck: It seems to work fine by opening a new tab like this, but just for consistency with the rest of our tests (and for my own sanity in tracking the execution thread) can you swap this out with a `driver.get( url )` to just load the activation in the current tab?

That's how we handle it in our Invite specs - https://github.com/Automattic/wp-e2e-tests/blob/master/specs/wp-invite-users-spec.js#L128

specs/wp-signup-spec.js
```diff
@@ -168,14 +173,48 @@ testDescribe( `[${host}] Sign Up  (${screenSize}, ${locale})`, function() {
...
+
+											this.adminTopBar = new AdminTopBar( this.driver );
+											this.adminTopBar.createNewPost();
+
+											this.editor = new EditorPage( this.driver );
+											this.editor.enterTitle( blogPostTitle );
+											this.editor.enterContent( blogPostQuote + '\n' );
+											return this.editor.publishEnabled().then( ( enabled ) => {
+												return assert.equal( enabled, false, 'Publish button is not disabled when activation link has not been clicked' );
+											} );
```
rachelmcr: Is testing for the disabled publish button sufficient for this step? The user should also see a prompt to verify their email at this stage, to explain the disabled publish button. What do you think of testing for that as well?

bsessions85: I like that idea. It doesn't hurt to have that additional step to make sure the message shows up.

bsessions85:
I fixed all of these, and did the autonumbering from #356 while I was at it. For the email activation notification, I had to exclude mobile because of the issue @rachelmcr mentioned. I commented in there with that issue so we can include it in the future

lib/pages/editor-page.js
```diff
@@ -140,4 +140,18 @@ export default class EditorPage extends BaseContainer {
...
+		const publishSelector = by.css( '.editor-publish-button' );
+		return this.driver.findElement( publishSelector ).getAttribute( 'disabled' ).then( ( d ) => {
+			return d !== 'true';
+		} );
+	}
+
+	emailVerificationNotice() {
+		const emailVerificationSelector = by.css( '.editor-ground-control__email-verification-notice' );
+		return driverHelper.isElementPresent( this.driver, emailVerificationSelector ).then( ( present ) => {
+			return present;
```
hoverduck: This seems a little redundant.  Can't we just `return driverHelper.isElementPresent( this.driver, emailVerificationSelector )` and leave out the `.then()`?

hoverduck:
One thing I just thought of...currently our Canary tests are executing this Free signup flow against wpcalypso, but when we get to the point in here where we load the editor to test publish capability all of those links are pointing to production.  So whether this is executed in canary or against a local calypso, that portion of the tests will always be run against wordpress.com.

@alisterscott what do you think?  Is it worth adding a step to rewrite the URL of the editor link before loading it to point back to the calypsoBaseURL (and likely the activation link as well)?  I don't like doing that if we don't have to, it feels error-prone, but it irks me to know that this bit of the test will never actually be run against pre-production code.

alisterscott:
Thanks for your PR @bsessions85 

> I fixed all of these, and did the autonumbering from #356 while I was at it.

For future reference we try to keep our PRs small(ish) and independent so they're easier to review and merge. You can leave the autonumbering in here this time, but in future it would be better to have a separate PR for that so that this PR doesn't have ~250 lines to review.

> @alisterscott what do you think? Is it worth adding a step to rewrite the URL of the editor link before loading it to point back to the calypsoBaseURL (and likely the activation link as well)?

Yes, we should always use the calypsoBaseURL not what is in the email content nor should we use the links from the front end admin toolbar since these will always point to wordpress.com and we need to be able to fully run these against any calypso environment. We already do this replacement for invite acceptance emails

alisterscott:
@hoverduck this PR adds a dependency on email (namely Mailosaur) for every canary run. Are we willing to take on that dependency when we need a quick yes/no canary test for wpcalypso? I'd be inclined to keep a simple canary sign up flow without email verification and add another longer one with email verification for full runs

lib/pages/editor-page.js
```diff
@@ -140,4 +140,16 @@ export default class EditorPage extends BaseContainer {
...
 	}
+
+	publishEnabled() {
+		const publishSelector = by.css( '.editor-publish-button' );
+		return this.driver.findElement( publishSelector ).getAttribute( 'disabled' ).then( ( d ) => {
+			return d !== 'true';
+		} );
+	}
+
+	emailVerificationNotice() {
```
alisterscott: This should have a more meaningful name like `emailVerificationNoticeShown()` or `emailVerificationNoticeDisplayed()`

specs/wp-signup-spec.js
```diff
@@ -168,14 +188,70 @@ testDescribe( `[${host}] Sign Up  (${screenSize}, ${locale})`, function() {
...
-										test.before( function() {
-											return this.emailClient = new EmailClient( signupInboxId );
+									test.describe( `Step ${stepNum}: Can not publish until email is confirmed`, function() {
+										stepNum++;
+
+										test.it( 'Can see a disabled publish button', function() {
+											const blogPostTitle = dataHelper.randomPhrase();
+											const blogPostQuote = dataHelper.randomPhrase();
+
+											this.adminTopBar = new AdminTopBar( driver );
```
alisterscott: We shouldn't use the admin bar - just use the calypso base url to visit the post page and use it directly (or navigate to it via calypso) - this is so these can run on any calypso environment including `calypso.localhost:3000`

specs/wp-signup-spec.js
```diff
@@ -168,14 +188,70 @@ testDescribe( `[${host}] Sign Up  (${screenSize}, ${locale})`, function() {
...
+											test.describe( `Step ${stepNum}: Can publish when email is confirmed`, function() {
+												stepNum++;
+
+												test.it( 'Can not see a disabled publish button', function() {
+													driver.wait( until.elementTextContains( driver.findElement( webdriver.By.css( '.editor-ground-control__save-status' ) ), 'Saved' ), mochaTimeOut );
+													driver.get( inboxEmails[0].html.links[0].href );
+													driver.navigate().back();
+
+													this.editor = new EditorPage( driver );
+													const publishButton = driver.findElement( webdriver.By.css( '.editor-publish-button' ) );
```
alisterscott: This shouldn't be in the spec - it should be pushed into the page itself and exposed as a method

specs/wp-signup-spec.js
```diff
@@ -168,14 +188,70 @@ testDescribe( `[${host}] Sign Up  (${screenSize}, ${locale})`, function() {
...
+													driver.wait( until.elementIsEnabled( publishButton ), mochaTimeOut );
```
alisterscott: The mochaTimeOut is the wrong timeout to use here, it should be the explicit wait timeout which is available in the page itself (see comment above)

specs/wp-signup-spec.js
```diff
@@ -168,14 +188,66 @@ testDescribe( `[${host}] Sign Up  (${screenSize}, ${locale})`, function() {
...
-									test.describe( 'Step Eight: Can activate my account from an email', function() {
-										test.before( function() {
-											return this.emailClient = new EmailClient( signupInboxId );
+									test.describe( `Step ${stepNum}: Can not publish until email is confirmed`, function() {
+										stepNum++;
+
+										test.it( 'Can see a disabled publish button', function() {
+											const blogPostTitle = dataHelper.randomPhrase();
+											const blogPostQuote = dataHelper.randomPhrase();
+											driver.get( calypsoBaseURL + '/post/' + newBlogAddress );
```
alisterscott: This _may_ be better in a page somewhere itself in case we want to reuse it, but we don't currenlty have a create post page (just an editor page which is also used for publishing WordPress pages) so I'm not sure where this would go 🤔

hoverduck:
cc: @alisterscott 

> this PR adds a dependency on email (namely Mailosaur) for every canary run. Are we willing to take on that dependency when we need a quick yes/no canary test for wpcalypso?

The canaries already have a dependency on Mailosaur, ever since switching to the Free Site flow - https://github.com/Automattic/wp-e2e-tests/blob/413aaf734918225d0605f78790e225087de0265b/specs/wp-signup-spec.js#L176 - But it's a good point, we should probably switch that up.  @bsessions85 can you move your activation steps out of the `@canary` flow and into the `Sign up for a Survey Step free site` flow instead?

> For future reference we try to keep our PRs small(ish) and independent so they're easier to review and merge. You can leave the autonumbering in here this time, but in future it would be better to have a separate PR for that so that this PR doesn't have ~250 lines to 

In this case I wrote #356 specifically with the idea in mind that we just take care of it next time we're working in this file, since it wasn't really worth a separate effort.   But you're right, it's definitely a best practice to avoid that sort of thing.   ¯\\\_(ツ)\_/¯

@bsessions85 - I ran into a timing issue on a couple of my test runs where the `driver.navigate().back()` call just reloaded the page from cache before refreshing, and the 'Please activate' message was still present long enough to fail the test.  It may be better to directly load the `baseURL/post/sitename` URL again rather than just hit the Back button.

bsessions85:
I think I've gotten it all figure out now. I didn't move the chunk @alisterscott mentioned should be pulled out since it isn't clear where to put it at this point. I added a check for the post being saved to the editor page so we can make sure we wait till it is saved to avoid the chrome page is not saved alert.

lib/pages/editor-page.js
```diff
@@ -152,4 +152,14 @@ export default class EditorPage extends BaseContainer {
 		const emailVerificationSelector = by.css( '.editor-ground-control__email-verification-notice' );
 		return driverHelper.isElementPresent( this.driver, emailVerificationSelector );
 	}
+
+	postSaved() {
+		const driver = this.driver;
+		const savedStatusSelector = by.css( '.editor-ground-control__save-status' );
+		return driver.wait( until.stalenessOf( driver.findElement( savedStatusSelector ) ), this.explicitWaitMS, 'The post should started to save by now' ).then( () => {
+			return driver.wait( until.elementLocated( savedStatusSelector ), this.explicitWaitMS, 'The post status should be saving by now' ).then( () => {
+				return driver.wait( until.elementTextIs( driver.findElement( savedStatusSelector ), 'Saved' ), this.explicitWaitMS, 'Post should have been saved by now' );
```
hoverduck: Can we trigger this off of a data attribute on the `savedStatusSelector` element rather than the text?  Looking for "Saved" won't work if we run these tests in another locale/language.

If there's no existing class name or attribute we can use we'll need to submit a PR to wp-calypso add one (like https://github.com/Automattic/wp-calypso/pull/15894 for example)

bsessions85: I put in a PR for wp-calypso to add it as an attribute [Automattic/wp-calypso#17603](https://github.com/Automattic/wp-calypso/pull/17603)

alisterscott:
> I added a check for the post being saved to the editor page so we can make sure we wait till it is saved to avoid the chrome page is not saved alert.

We have a helper method that does something similar to that, although it's a bit hacky but it works

`ensureSaved():`

https://github.com/Automattic/wp-e2e-tests/blob/master/lib/components/post-editor-toolbar-component.js#L18

hoverduck:
:man_facepalming: I completely forgot we had the helper method already, I was just focused on the locale issue.  Thanks @alisterscott.

Adding the data attribute to Calypso is probably better in the long run for completeness' sake, but for this PR I'd vote to just use `ensureSaved()` and move on.  If/when the attribute is merged into Calypso we can just update that method to take advantage of it.

bsessions85:
I removed the postSaved method and used ensureSaved. I will have to do more searching next time before adding one!  Thanks for all the feedback!

PR #682, opened 2017-08-29
Remove support for DISABLE_EMAIL flag

alisterscott:
Use SKIP_TEST_REGEX instead


alisterscott:
@rachelmcr @hoverduck just a heads up on this, I am removing DISABLE_EMAIL in favour of SKIP_TEST_REGEX

Issue #302, opened 2016-10-18
Add test for Featured Image dialog


hoverduck:
To catch this scenario - https://github.com/Automattic/wp-calypso/issues/8822

brbrr:
This will rely on https://github.com/Automattic/wp-calypso/pull/17565

brbrr:
Closed by #694

PR #694, opened 2017-08-31
Add Featured image test

brbrr:
as requested in #302

### DIFF
--- a/lib/components/post-editor-sidebar-component.js
+++ b/lib/components/post-editor-sidebar-component.js
@@ -48,6 +48,14 @@ export default class PostEditorSidebarComponent extends BaseContainer {
 		return this._expandOrCollapseSection( 'drawer__more-options', false );
 	}
 
+	expandFeaturedImage() {
+		return this._expandOrCollapseSection( 'featured-image', true );
+	}
+
+	closeFeaturedImage() {
+		return this._expandOrCollapseSection( 'featured-image', false );
+	}
+
 	addNewCategory( category ) {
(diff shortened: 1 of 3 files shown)


hoverduck:
Just adding a link to cross-reference this with the data-attribute PR: https://github.com/Automattic/wp-calypso/pull/17565

brbrr:
Hm, that 'dirty' flag doesn't look like expected behavior. Featured image was deleted, so there no actual changes to be saved. For my understanding is a bug, and removing unexpected alert is kinda workaround. A better approach would be to issue a bug in wp-calypso. If you agree - I'll create an issue in the wp-calypso repo.

rachelmcr:
> It's weird because adding/deleting images to the post by itself doesn't set the flag, but featured image does.

Uploading/deleting images in the site's media library from within the editor doesn't set the flag, but if you add images to the post content and then delete them it does. That's more analogous to adding/removing the featured image.

> Hm, that 'dirty' flag doesn't look like expected behavior. Featured image was deleted, so there no actual changes to be saved.

I tested other editor actions such as adding and deleting an image from the post content, adding and removing text in the post, and adding and removing tags/categories. In each case, the initial change sets the "dirty" flag but removing the change does not unset the "dirty" flag.

In that sense, the featured image behavior is consistent with how the AYS dialog works with other features in the editor, but I don't think it's ideal. Opening an issue in the wp-calypso repo to raise that seems like a good idea, and perhaps it will help shed light on the AYS bug in https://github.com/Automattic/wp-calypso/issues/9833.

brbrr:
Committed workaround of wp-calypso/#17701. It should be removed once the issue is resolved.

PR #699, opened 2017-09-04
Try a couple of waits for Jetpack specs

alisterscott:


alisterscott:
Already fixed by @hoverduck Closing now

PR #701, opened 2017-09-05
Try clearing cart through WP.com Rest API

alisterscott:
**Work in progress - not currently working as expected**


alisterscott:
I want to use the wpcom library instead for the rest stuff

See: https://github.com/Automattic/wp-e2e-tests/issues/702

alisterscott:
@rachelmcr @hoverduck do you know what I am doing wrong here? I can't get the service call to happen at all in my test to empty the cart

rachelmcr:
Just a small thing — in the API call `shopping-cart` needs to use bracket notation:

```
let res = service.sites['${site}']['shopping-cart'].clear.POST( {
	params: { site: account[2] }
} );
```

And then in the next line the response code should be 200 rather than 444.

That should do the trick (along with making sure the blog address or ID is defined for that account in the config), but I'm happy to look again if there's any more trouble.

hoverduck:
This is awesome!

Should we move it into a helper function somewhere?  We also call the existing `shoppingCartWidgetComponent.empty()`function in the `specs-jetpack-calypso/wp-jetpack-plans-spec.js` spec

alisterscott:
Thanks for your help @rachelmcr 

> Just a small thing — in the API call shopping-cart needs to use bracket notation:

I updated it to this but I still can't see it working 🤔 

> And then in the next line the response code should be 200 rather than 444.

Yes, I was trying to get it to fail so I know it's working, rather than silently passing which it's currently doing 😕

> That should do the trick (along with making sure the blog address or ID is defined for that account in the config)

Yes, I added this, and it should be on CircleCI now also

rachelmcr:
I tried with your branch (no local changes) and the CircleCI config and it worked for me. (The test failed as expected and the cart was cleared with a 200 response from the API.) However, when I changed the expected response from 444 to 200 to get the expected passing behavior, the test passed but the cart wasn't cleared.

When I run the spec in bluecat's debug mode (using `BLUECAT_DEBUG_CONSOLE=true`) with either 444 or 200 as the expected response, I see the API request & 200 response and the cart is successfully cleared.

It looks to me like the test is ending before the API call finishes (so when we force a test failure or debug mode is enabled, it slows things down long enough for the API call to complete). When I remove the last `return true` in this test, the test passes and the cart is cleared — that might be the last piece to get this to actually pass instead of silently passing.

alisterscott:
@bsessions85 I know you did some API stuff recently - would you mind reviewing this to see if it makes sense (and if it works for you)

bsessions85:
I played with this for a while yesterday and couldn't get it to work. The test always passes for me, even  with the 444 in there. I played around with removing `return true` and some other things but couldn't get it to work properly. The call itself seems to be working and clearing the cart.

bsessions85:
I played with this a little more today and I have it working. The key is using the "done" call. I found it in the bluecat unit tests and it seems to work here too.

`				// Remove all items from basket for clean up
				test.it( 'Can empty cart for next test', ( done ) => {
					return service.run(function () {
						oauth.login(service, account);
						let res = service.sites['${site}']['shopping-cart'].clear.POST({
							params: {site: account[2]}
						});
						expect(res.data.statusCode).to.equal(200);
						console.log('blah');
						console.log(res.data.body);

						done();
					})`

alisterscott:
Thanks @bsessions85 !!!

alisterscott:
Now I'm getting `Could not capture mocha tests` 😠

hoverduck:
> Now I'm getting Could not capture mocha tests 

This was insanely tricky.  It boils down to a compile issue for a binary utility used by the bluecat package.  Something isn't working right on the Docker image we're using in CI when we do `npm install` the first time, but I added a step to clear out just the offending package and rebuild it in `run.sh`, and that seems to work.  It's not ideal, so I'll add a workaround ticket to try and fix it later, but it doesn't add a huge delay to the tests.

Here are my debugging steps for future reference:

I SSHed into the CircleCI container to manually run the debug command in the Magellan output.  I was also able to see the problem by running `circleci build` locally, but that requires a bit of extra work to get the NODE_CONFIG set up correctly.  Note that in order to configure node to run interactively you have to first run `source $HOME/.nvm/nvm.sh && nvm use`, which is otherwise handled by the `run.sh` setup.

Running that command I got an error with the `fibers` package required by `bluecat`, suggesting a rebuild (which didn't work due to a lack of node-gyp).  From the SSH session I was able to blow away the `node_modules` folder and re-run `npm install`, which worked fine.

I tried re-running the build without cache (which is just the `node_modules` dir), but it still failed.  Scratching my head, I added a step to run.sh for CI-only to blow away just the `fibers` directory and re-do the install.  That seems to work fine...I hoped that it would then have a valid file cached, so I tried removing those steps, but it still failed.

I tried a couple of other workarounds I found elsewhere ([#](https://github.com/webdriverio/wdio-mocha-framework/issues/30), [#](https://github.com/laverdet/node-fibers/issues/321#issuecomment-276758931)), but neither worked.

I also tried re-ordering a few things in the build process to try and save a valid cache, but nothing seemed to stick.  It's possible that this will still be cleared up by a proper cache once it's merged to master, but I'm not counting on it.  And I wouldn't want to rely on that anyway, because then the next time we make a change to `package.json` it will invalidate the cache and break this again.

hoverduck:
And I did all that work above before I realized that you're not actually adding `bluecat` here, that was done a while back.  So I have NO idea why it's suddenly breaking here.  And we could also probably fix it by getting away from bluecat/fibers and using wpcom instead.  I'll hold off on writing up a workaround issue until that decision is made.  I haven't worked with the API stuff directly so I don't know how much effort it would be to switch out.

bsessions85:
I took a quick look at what it would take to use wpcom instead and it wasn't clear if it was possible to make the "hidden" api calls with it.

alisterscott:
Thanks for the sleuthing @hoverduck !!!
I'll have a look sometime next week to see whether I can use wpcom instead

PR #716, opened 2017-09-13
Add checking for error messages on all pages

### LINKED ISSUE #711: Invalid primary site error during invite tests
The invalid site error is being shown during the invite acceptance tests. Introduced here: https://github.com/Automattic/wp-calypso/pull/17731

We need to:

1. Investigate why the invite sign ups are showing this error and whether it's an issue for users
2. Add some checking to our e2e tests to fail when they see such an error coming up

alisterscott:
This adds a check to every page to make sure that neither an error nor a warning appears:

screen shot 2017-09-13 at 1 18 52 pm

If it does, it gives us a Slack warning.

The problem is this seems to double to execution time in CircleCI.

@hoverduck @rachelmcr is this worth it for slowing things down drastically?

See #711 for some context (which ended up being a false alert)

### DIFF
--- a/lib/base-container.js
+++ b/lib/base-container.js
@@ -9,7 +9,7 @@ import * as slackNotifier from './slack-notifier';
 
 
 export default class BaseContainer {
-	constructor( driver, expectedElementSelector, visit = false, url = null, checkErrorsWarnings = true ) {
+	constructor( driver, expectedElementSelector, visit = false, url = null, checkErrors = true, checkWarnings = true ) {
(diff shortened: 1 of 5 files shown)


hoverduck:
Yeah, I don't think it's worth doubling the execution time.  What about setting the default value for `checkErrors` and `checkWarnings` to false, and only enabling it on certain pages?  Like maybe the Invites page and the Editor?  Any others?

alisterscott:
Decided to leave this as the double in execution time isn't worth it

PR #719, opened 2017-09-15
Check publicize defaults to empty message

alisterscott:
Change introduced here: https://github.com/Automattic/wp-calypso/pull/18028

We won't assert on the content of the placeholder as that is localised (we can't be assured of it being in English)


Tug:
Thank you for this @alisterscott!

PR #722, opened 2017-09-19
Fix URLs for calypso.live usage

alisterscott:
This PR fixes the URLs for calypso.live to logon etc


alisterscott:
@hoverduck would you mind reviewing this? The intention is to use this new warmup capability to make sure a live branch is active on calypso.live before running some e2e tests against it

hoverduck:
Is there a strong reason to do the warmup check internally in WebDriver?  I feel like it would make more sense to have a loop externally in `run.sh` and/or the CircleCI config that watches for the branch to be ready using something like

```bash
curl https://calypso.live/?list_branches 2>/dev/null | jq '.ready | contains( ["$BRANCHNAME"] )'
```

alisterscott:
> Is there a strong reason to do the warmup check internally in WebDriver? I feel like it would make more sense to have a loop externally in run.sh and/or the CircleCI config that watches for the branch to be ready using something like

@hoverduck: nope, great idea! 

Moving this idea over to https://github.com/Automattic/wp-e2e-tests-for-branches/pull/2 as it makes sense to do in the live branch wrapper repo

alisterscott:
This PR still contains some changes to how we determine live branch URLs which will be necessary for https://github.com/Automattic/wp-e2e-tests-for-branches/pull/2 to work correctly

@hoverduck do the remaining changes look okay?

hoverduck:
LGTM!

Just one comment...do you think it's worth adding a helper function to do `config.has( 'liveBranch' ) && config.get( 'liveBranch' )` to avoid repitition?

alisterscott:
>> Just one comment...do you think it's worth adding a helper function to do config.has( 'liveBranch' ) && config.get( 'liveBranch' ) to avoid repitition?

Thanks @hoverduck  and fixed now

PR #731, opened 2017-10-03
Workaround: Skip media upload test on mobile

hoverduck:
Calypso issue https://github.com/Automattic/wp-calypso/issues/18488 documents a problem with the media modal on mobile (say that 5 times fast), which is causing the media upload tests to fail.  This PR is skipping those tests until it's fixed (which is more complicated than it seems).  The fix is attempted in https://github.com/Automattic/wp-calypso/issues/18500

### DIFF
--- a/specs/wp-media-upload-spec.js
+++ b/specs/wp-media-upload-spec.js
@@ -22,146 +22,149 @@ test.before( function() {
 	driver = driverManager.startBrowser();
 } );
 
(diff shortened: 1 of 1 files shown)


stephanethomas:
@hoverduck: I've merged https://github.com/Automattic/wp-calypso/pull/18500, you may want to revert that workaround now.

PR #737, opened 2017-10-10
Ensure logged out as commenter before logging in to view comment

hoverduck:
The Reader commenting spec was consistently failing when running against calypso.localhost due to a timing issue in the way we were logging out and then back in.

@rachelmcr / @alisterscott were you seeing this problem, or just me?  Could be just me.


rachelmcr:
I saw that timing issue once last week while running against calypso.localhost but haven't seen it since then, so it wasn't a consistent problem for me but 👍  to preventing it.

PR #739, opened 2017-10-17
Update/update ensure saved

bsessions85:
The ensureSaved function was causing issues on e2e test run https://github.com/Automattic/wp-calypso/pull/18878 so I figured it was a good time to finish the changes I started during trial.

### DIFF
--- a/lib/components/post-editor-toolbar-component.js
+++ b/lib/components/post-editor-toolbar-component.js
@@ -24,7 +24,7 @@ export default class PostEditorToolbarComponent extends BaseContainer {
 		driver.wait( function() {
 			return driverHelper.isElementPresent( driver, saveSelector ).then( function( present ) {
 				return ! present;
-		 	} );
+			} );
 		}, this.explicitWaitMS, 'The save selector was still present when it should have disappeared with auto-save.' );
 
 	}
--- a/lib/components/post-editor-toolbar-component.js
+++ b/lib/components/post-editor-toolbar-component.js
@@ -15,13 +15,18 @@ export default class PostEditorToolbarComponent extends BaseContainer {
 		this.publishButtonSelector = By.css( '.editor-publish-button' );
 	}
 
-	ensureSaved() {
+	ensureSaved( clickSave = true ) {
+		const saveSelector = By.css( 'button.editor-ground-control__save' );
 		const savedSelector = By.css( 'span.editor-ground-control__save-status[data-e2e-status="Saved"]' );
 		const driver = this.driver;
 
+		if ( clickSave === true ) {
+			driverHelper.clickIfPresent( driver, saveSelector, 3 );
+		}
+
 		return driver.wait( function() {
 			return driverHelper.isElementPresent( driver, savedSelector ).then( function( present ) {
-				return ! present;
+				return present;
 			} );
 		}, this.explicitWaitMS, 'The saved selector was still not present when it should have been with auto-save.' );
 
--- a/lib/components/post-editor-toolbar-component.js
+++ b/lib/components/post-editor-toolbar-component.js
@@ -16,8 +16,7 @@ export default class PostEditorToolbarComponent extends BaseContainer {
 	}
 
 	ensureSaved() {
-		const saveSelector = By.css( 'button.editor-ground-control__save' );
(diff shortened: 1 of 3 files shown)


hoverduck:
I'm a bit confused.  This PR looks for the attribute `data-e2e-status="Saved"` and clicks it if present.  But if the attribute is "Saved" doesn't that mean it's already saved and doesn't need to be clicked?  I'm also looking at the changes you made in https://github.com/Automattic/wp-calypso/pull/17603 (which were actually refactored a bit a couple days ago - https://github.com/Automattic/wp-calypso/blob/master/client/post-editor/editor-ground-control/index.jsx) and now I notice there are only the two states represented, `Saved` and `Saving`.  Should there be a state that says `Unsaved` or something?

Or alternately, just to address it from the e2e side, we could keep the `const saveSelector` definition as-is (maybe rename it to `saveButtonSelector`) and say clickIfPresent, then wait for `data-e2e-status="Saved"` to be present to resolve the promise.

bsessions85:
@hoverduck  Thanks. The click was in there from before and if we're interested in making sure that auto save is working, it shouldn't be there at all. I removed it and now we're just waiting to make sure that the `data-e2e-status="Saved"` attribute shows up.

hoverduck:
Hmm...I like the thought that we're using this to verify that the autosave functionality is working, but I wonder if it's also important that we enforce that the post is saved to prevent spurious failures, or having to wait for autosave when it'd be faster to just click the button and move on.

Maybe we could have an optional parameter on the function that waits for autosave (if it's not already done), but we default to clicking Save (if necessary).  Then we could have a specific spec where we wait for Autosave, but for rest of the time we force it.

bsessions85:
OK, I added a param for it and set it false in one test. I also ended up removing one ensureSaved check because it wasn't relevant with the new revisioning stuff.

lib/components/post-editor-toolbar-component.js
```diff
@@ -15,24 +15,21 @@ export default class PostEditorToolbarComponent extends BaseContainer {
...
-			} );
-		}, this.explicitWaitMS, 'The save selector was still present when it should have disappeared with auto-save.' );
+		if ( clickSave === true ) {
+			driverHelper.clickIfPresent( driver, saveSelector, 3 );
+		}
 
 		return driver.wait( function() {
-			return driverHelper.isElementPresent( driver, savingSelector ).then( function( present ) {
-				return ! present;
+			return driverHelper.isElementPresent( driver, savedSelector ).then( function( present ) {
```
alisterscott: I think this whole block could be replaced with calling `driverHelper.waitTillPresentAndDisplayed(...)` (I think this existing code pre-dates that helper method)

specs/wp-post-editor-spec.js
```diff
@@ -863,7 +863,6 @@ test.describe( `[${host}] Editor: Posts (${screenSize})`, function() {
 						assert.equal( errorShown, false, 'There is an error shown on the editor page!' );
 					} );
 					this.postEditorToolbarComponent = new PostEditorToolbarComponent( driver );
-					this.postEditorToolbarComponent.ensureSaved();
```
alisterscott: Why does this work not calling this now? The reason we have ensureSaved is to avoid the "unsaved changes" dialogs which sometimes appear as a race condition - are we sure these won't show?

specs/wp-signup-spec.js
```diff
@@ -814,7 +814,7 @@ testDescribe( `[${host}] Sign Up  (${screenSize}, ${locale})`, function() {
 														this.editor.enterContent( blogPostQuote + '\n' );
 
 														this.postEditorToolbarComponent = new PostEditorToolbarComponent( driver );
-														this.postEditorToolbarComponent.ensureSaved();
+														this.postEditorToolbarComponent.ensureSaved( false );
```
alisterscott: If you change the method to use named parameters this will need to be something like: `.ensureSaved( { clickSave: false } );`

specs/wp-post-editor-spec.js
bsessions85: It works because it really wasn't needed. We're not trying to change the page or anything, We are clicking the update button which saves it for us.

PR #741, opened 2017-10-18
Match http or https when working with Jetpack sites

sirreal:
This PR updates the regex which trims the scheme from a Jetpack sitename to match `http` as well as `https` at the start of the string.

This is useful when working with "disposable" sites, such as those provided by http://poopy.life


hoverduck:
It's not really relevant to your changes here, but just something related to keep in mind - there are a few things that are different on http vs https sites, like post/site preview (http will launch in a new tab as opposed to a modal popup on https).  We differentiate our pre-configured hosts here - https://github.com/Automattic/wp-e2e-tests/blob/master/config/default.json#L58

PR #742, opened 2017-10-18
Removed Themes step from signup flows

hoverduck:
https://github.com/Automattic/wp-calypso/pull/18908 removed the Themes step from all the signup flows.  This PR updates the tests to reflect that.


hoverduck:
@bsessions85 It's less ugly than it would've been before #356 :smile:

hoverduck:
Bah, I misread the Calypso PR...I thought they went ahead and removed the Themes step from _every_ flow, but that's just an action for a later date.  Adding it back in where necessary...

PR #744, opened 2017-10-18
Try/chrome headless

alisterscott:


alisterscott:
It doesn't work on CircleCI yet because I think we need to upgrade Chrome.

@hoverduck: how I do upgrade Chrome in the docker containers for Circle? Is it under /docker-compose? Would you mind writing a P2 post about how to update this and where all the pieces sit (or point me to one if this already exists sorry)

bsessions85:
I did some testing and was able to run the tests successfully in the docker containers locally with 3.6.0 of the selenium docker images. CircleCI is using the image automattic/wp-e2e-tests:0.0.5 which is set up by the dockerfile in root. We'll just need to update that file, build the image, and push it to the docker hub with a new tag.  I found instruction on how to do that, but would probably need access to our hub.

hoverduck:
> We'll just need to update that file, build the image, and push it to the docker hub with a new tag

Yep, that should be it.  I only found out recently that we have an internal a8c Docker registry, so I'd like to switch to using that if we can (I need to follow up on it).  Otherwise for now I'm the only one with privileges on the Automattic repo on hub.docker.com.  If you all have accounts on that site I can add you, but for now I'll plan on testing/pushing the updates tomorrow.

hoverduck:
I made the version updates in a separate PR - #745 - so you should be able to rebase and pick that up here.

.circleci/config.yml
```diff
@@ -17,7 +17,7 @@ jobs:
             - /wp-e2e-tests/node_modules
           key: << checksum "package.json" >>
       - run: ./scripts/randomize.sh specs
-      - run: ./scripts/run-wrapper.sh
+      - run: env HEADLESS=1 ./scripts/run-wrapper.sh
```
hoverduck: Rather than doing this, why don't we just repurpose the `-x` argument to run.sh to set the envvar?  That way it'll be picked up by all the wrapper repos as well.  It'll also make that argument functional across OSes instead of just on Linux.

bsessions85:
Made the changes to use the -x flag as well.

bsessions85:
@hoverduck  Do these changes look like what you were thinking?

hoverduck:
:+1:

PR #750, opened 2017-10-24
Jetpack connect test using a jurassic.ninja site

seear:
A test for connecting a .org site running jetpack, starting from Calypso. Adds a library module `wporg-creator-page` that spins up a new .org site with jetpack installed per-test. This uses the `jurassic.ninja` service, an internal poopy.life clone.

Running a new .org site per test allows using a different set of features to the other jetpack tests (for example SSO turned off), and connecting/disconnecting without affecting the other tests.

These tests require a longer timeout, I'm running with 40 seconds, due to the time to create a .org site and the time to authorize the jetpack connection.

Also requires user `e2eflowtestingjetpackconnect`.

/cc @oskosk @sirreal

### DIFF
--- a/lib/pages/signup/pick-a-plan-page.js
+++ b/lib/pages/signup/pick-a-plan-page.js
@@ -30,6 +30,13 @@ export default class PickAPlanPage extends BaseContainer {
 
 		return this._selectPlan( 'free' );
 	}
+	// Explicitly select the free button on jetpack without needing `host` above.
+	selectFreePlanJetpack() {
+		return driverHelper.clickWhenClickable(
+			this.driver,
+			By.css( '.jetpack-connect__plans-nav-buttons button' )
+		);
+	}
 	selectPersonalPlan() {
 		return this._selectPlan( 'personal' );
 	}
--- a/lib/pages/signup/pick-a-plan-page.js
+++ b/lib/pages/signup/pick-a-plan-page.js
@@ -1,3 +1,4 @@
+import config from 'config';
 import { By, until } from 'selenium-webdriver';
 
 import BaseContainer from '../../base-container.js';
@@ -82,5 +83,12 @@ export default class PickAPlanPage extends BaseContainer {
 		const selector = By.css( 'div.business-bundle div.wpcom-plan-price span' );
 		this.driver.wait( until.elementLocated( selector ), this.explicitWaitMS, 'Could not locate the business plan price element' );
 		return this.driver.findElement( selector ).getText();
+	}
+	static waitForPage( driver ) {
+		driverHelper.waitTillPresentAndDisplayed(
+			driver,
+			By.css( '.plans-features-main__group' ),
+			config.get( 'explicitWaitMS' ) * 2
+		);
 	}
 }
--- a/lib/pages/signup/pick-a-plan-page.js
+++ b/lib/pages/signup/pick-a-plan-page.js
@@ -12,7 +12,13 @@ const screenSize = currentScreenSize();
 
 export default class PickAPlanPage extends BaseContainer {
 	constructor( driver ) {
-		super( driver, By.css( '.plans-features-main__group' ) );
+		super(
+			driver,
+			By.css( '.plans-features-main__group' ),
+			false,
+			null,
+			config.get( 'explicitWaitMS' ) * 2
+		);
 	}
 	_selectPlan( level ) {
 		let prefix = 'table';
@@ -83,12 +89,5 @@ export default class PickAPlanPage extends BaseContainer {
 		const selector = By.css( 'div.business-bundle div.wpcom-plan-price span' );
 		this.driver.wait( until.elementLocated( selector ), this.explicitWaitMS, 'Could not locate the business plan price element' );
 		return this.driver.findElement( selector ).getText();
-	}
-	static waitForPage( driver ) {
-		driverHelper.waitTillPresentAndDisplayed(
-			driver,
-			By.css( '.plans-features-main__group' ),
-			config.get( 'explicitWaitMS' ) * 2
-		);
 	}
 }
--- a/lib/pages/wporg-creator-page.js
+++ b/lib/pages/wporg-creator-page.js
@@ -6,6 +6,7 @@ import * as driverHelper from '../driver-helper';
 
 const TEMPLATE_URL = 'http://poopy.life/create?src=weary-duck&key=pVP6jBZR1C6wALZh';
 const PASSWORD_ELEMENT = by.css( '#tdr_password' );
+const URL_ELEMENT = by.css( '#tdr_url' );
 
 export default class WporgCreatorPage extends BaseContainer {
 	constructor( driver ) {
@@ -20,6 +21,11 @@ export default class WporgCreatorPage extends BaseContainer {
 			'Could not locate password element'
 		);
 		return this.driver.findElement( PASSWORD_ELEMENT ).getText();
+	}
+
+	getUrl() {
+		this.driver.wait( until.elementLocated( URL_ELEMENT ) );
+		return this.driver.findElement( URL_ELEMENT ).getText();
 	}
 
 	waitForWpadmin() {
--- a/lib/pages/wporg-creator-page.js
+++ b/lib/pages/wporg-creator-page.js
@@ -8,10 +8,12 @@ const TEMPLATE_URL = 'http://jurassic.ninja/create';
 const PASSWORD_ELEMENT = By.css( '#jurassic_password' );
 const URL_ELEMENT = By.css( '#jurassic_url' );
 const CONTINUE_LINK = By.linkText( 'The new WP is ready to go, visit it!' );
+const PROGRESS_MESSAGE = By.css( '#progress' );
 
 export default class WporgCreatorPage extends BaseContainer {
 	constructor( driver ) {
-		super( driver, CONTINUE_LINK, /* visit url */ true, TEMPLATE_URL );
+		super( driver, PROGRESS_MESSAGE, /* visit url */ true, TEMPLATE_URL );
+		driverHelper.waitTillPresentAndDisplayed( driver, CONTINUE_LINK, this.explicitWaitMS * 2 );
 		driverHelper.clickWhenClickable( driver, CONTINUE_LINK );
 	}
 
--- a/lib/pages/wporg-creator-page.js
+++ b/lib/pages/wporg-creator-page.js
@@ -1,0 +1,28 @@
+/** @format */
+import { By as by, until } from 'selenium-webdriver';
+import BaseContainer from '../base-container';
+
+import * as driverHelper from '../driver-helper';
+
+const TEMPLATE_URL = 'http://poopy.life/create?src=weary-duck&key=pVP6jBZR1C6wALZh';
+const PASSWORD_ELEMENT = by.css( '#tdr_password' );
+
+export default class WporgCreatorPage extends BaseContainer {
+	constructor( driver ) {
+		super( driver, by.css( '.continue-to-install' ), /* visit url */ true, TEMPLATE_URL );
+		driverHelper.clickWhenClickable( driver, by.css( '.continue-to-install' ) );
+	}
+
+	getPassword() {
+		this.driver.wait(
+			until.elementLocated( PASSWORD_ELEMENT ),
+			3000,
+			'Could not locate password element'
+		);
+		return this.driver.findElement( PASSWORD_ELEMENT ).getText();
+	}
+
+	waitForWpadmin() {
+		return driverHelper.waitTillPresentAndDisplayed( this.driver, PASSWORD_ELEMENT );
+	}
+}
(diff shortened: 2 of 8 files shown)


oskosk:
I'm down for this! BUT! We should ask @hoverduck for his opinion about running tests against this. 

Just a thought in the air, but I'd say that they are taking care more than me about the involved systems and having them be reliable for automated testing that goes into a development workflow.

specs-jetpack-calypso/wp-jetpack-connect-spec.js
```diff
@@ -0,0 +1,104 @@
...
+		} );
+
+		test.it( 'Can disconnect existing jetpack sites', done => {
+			const siteSwitcherSelector = By.css( '.current-site__switch-sites' );
+
+			const removeSites = () => {
+				driver.findElements( siteSwitcherSelector ).then( found => {
+					if ( ! found.length ) {
+						// only one site left
+						return done();
```
sirreal: Got it! 🎉

specs-jetpack-calypso/wp-jetpack-connect-spec.js
```diff
@@ -0,0 +1,121 @@
...
+			driver.getCurrentUrl().then( urlDisplayed => {
+				this.sidebarComponent.setABTestControlGroupsInLocalStorage( urlDisplayed );
+			} );
+		} );
+
+		test.it( 'Can disconnect existing jetpack sites', done => {
+			const siteSwitcherSelector = By.css( '.current-site__switch-sites' );
+			const siteSelector = By.css( '.site-selector .site a[aria-label*="ninja"]' );
+
+			const removeSites = () => {
```
seear: This whole function can move into `sidebar-component` as `removeAllJetpackSites()` and work by `.is-jetpack` instead of `poopy` or `ninja`.

specs-jetpack-calypso/wp-jetpack-connect-spec.js
```diff
@@ -0,0 +1,121 @@
...
+							driver.navigate().refresh();
+							return done();
+						}
+
+						this.sidebarComponent.searchForSite( 'ninja' );
+						this.sidebarComponent.selectSettings();
+						const settingsPage = new SettingsPage( driver );
+						settingsPage.manageConnection();
+						settingsPage.disconnectSite();
+						driverHelper.waitTillPresentAndDisplayed( driver, By.css( '.is-success' ) );
```
sirreal: 👍  Wait for confirm

specs-jetpack-calypso/wp-jetpack-connect-spec.js
sirreal: As long as there's a limited number of sites. If that site list grows beyond a threshold and sites are hidden, we would need a search phrase.

lib/pages/wporg-creator-page.js
```diff
@@ -0,0 +1,31 @@
+/** @format */
+import { By as by, until } from 'selenium-webdriver';
```
sirreal: Why the aliased import?

specs-jetpack-calypso/wp-jetpack-connect-spec.js
```diff
@@ -0,0 +1,121 @@
...
+	test.describe( 'Connect From Calypso:', function() {
+		this.bailSuite( true );
+
+		test.before( function() {
+			return driverManager.clearCookiesAndDeleteLocalStorage( driver );
+		} );
+
+		test.it( 'Can create wporg site', () => {
+			this.wporgCreator = new WporgCreatorPage( driver );
+			this.wporgCreator.waitForWpadmin();
```
sirreal: I'm not totally clear about tests like this. There's no return and it's not async via `done`.

Does the test runner wait for `waitForWpadmin`?

Should we be testing some sort of result, or does a lack of errors satisfy the test?

specs-jetpack-calypso/wp-jetpack-connect-spec.js
```diff
@@ -0,0 +1,121 @@
...
+		} );
+
+		test.it( 'Can get URL', () => {
+			this.wporgCreator.getUrl().then( url => {
+				this.url = url;
+			} );
+		} );
+
+		test.it( 'Can log in', () => {
```
sirreal: This (and subsequent tests) depend on getting a URL for a new site, part of a previous test. I had a quick look at how to do dependent tests and found that we can use a new `describe` that contains the dependent part with a `before` that will throw if our conditions aren't met ([#](https://stackoverflow.com/a/36003551/1432801)).

specs-jetpack-calypso/wp-jetpack-connect-spec.js
seear: > does a lack of errors satisfy the test?

I think so. Breaking things up into separate tests aids debugging. See for example https://github.com/Automattic/wp-e2e-tests/blob/master/scripts/jetpack/wp-jetpack-activate.js

specs-jetpack-calypso/wp-jetpack-connect-spec.js
```diff
@@ -0,0 +1,100 @@
...
+			removeSites();
+		} );
+
+		test.it( 'Can add new site', () => {
+			this.sidebarComponent.addNewSite( driver );
+			const addNewSitePage = new AddNewSitePage( driver );
+			addNewSitePage.addSiteUrl( this.url );
+		} );
+
+		test.it( 'Can click the free plan button', () => {
```
hoverduck: I was getting consistent timeouts on this step waiting for the plan selection to be visible.  Maybe need to increase the timeouts in the `selectFreePlanJetpack()` function?

specs-jetpack-calypso/wp-jetpack-connect-spec.js
```diff
@@ -0,0 +1,100 @@
...
+const screenSize = driverManager.currentScreenSize();
+
+var driver;
+
+test.before( function() {
+	this.timeout( startBrowserTimeoutMS );
+	driver = driverManager.startBrowser();
+} );
+
+test.describe( `Jetpack Connect: (${ screenSize }) @jetpack`, function() {
```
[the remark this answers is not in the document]
seear: Does logging in with the [explicit user here](https://github.com/Automattic/wp-e2e-tests/pull/750/files#diff-090b3a4ee1ff9de99f050ceae1ad6f45R51) give us enough protection?

specs-jetpack-calypso/wp-jetpack-connect-spec.js
```diff
@@ -0,0 +1,100 @@
...
+				this.url = url;
+			} );
+		} );
+
+		test.it( 'Can log in', () => {
+			const loginFlow = new LoginFlow( driver, 'jetpackConnectUser' );
+			loginFlow.loginAndSelectMySite();
+		} );
+
+		test.it( 'Can disconnect existing jetpack sites', () => {
```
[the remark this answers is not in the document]
seear: Are there setup/teardown locations we can do this that run before/after all the tests (not just before all the tests in one spec)?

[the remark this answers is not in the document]
hoverduck: We have `lib/after.js` that has our global teardown/processing.

specs-jetpack-calypso/wp-jetpack-connect-spec.js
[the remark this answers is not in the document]
hoverduck: Yep, that works :)

specs-jetpack-calypso/wp-jetpack-connect-spec.js
seear: Timeouts now fixed.

seear:
@hoverduck on the matter of site disconnections interfering with other tests: currently this is the only test using the `e2eflowtestingjetpackconnect` account, so we haven't hit that problem yet.

It might make sense address this when we add the next test, just so we can tell that we have a working solution?

hoverduck:
> @hoverduck on the matter of site disconnections interfering with other tests: currently this is the only test using the e2eflowtestingjetpackconnect account, so we haven't hit that problem yet.
> 
> It might make sense address this when we add the next test, just so we can tell that we have a working solution?

While this is the only test using that account, it could be run in multiple contexts at the same time.  And now that I'm thinking about it more I take back what I said earlier about having the gate for running the test just be the account itself being sufficient.

As it's written now this test will run (with the `e2eflowtestingjetpackconnect` account) even if the rest of the tests are being executed against our pre-configured BLUEHOST or PRESSABLE sites.  

What are your thoughts for when this test should run?  Should it be included with the CI tests run when a change is merged into Jetpack/master?  That's the only existing place that makes sense.  We could also create a new category of tests entirely if necessary.

lib/pages/signup/pick-a-plan-page.js
```diff
@@ -83,4 +84,11 @@ export default class PickAPlanPage extends BaseContainer {
 		this.driver.wait( until.elementLocated( selector ), this.explicitWaitMS, 'Could not locate the business plan price element' );
 		return this.driver.findElement( selector ).getText();
 	}
+	static waitForPage( driver ) {
```
hoverduck: Rather than override the BaseContainer function here I'd rather see it updated to take an optional parameter in case there are other places where we need to modify the default.  But I'm also fine just recording that in an issue labeled `task` and leaving this as-is if you don't have time to fix it now.

seear: > I'd rather see it updated to take an optional parameter in case there are other places where we need to modify the default.

Yeah good plan, I'll do that now.

seear:
> What are your thoughts for when this test should run? Should it be included with the CI tests run when a change is merged into Jetpack/master?

This test and the ones that follow are testing calypso rather than the Jetpack plugin, so it will make more sense to run them after a merge to wp-calypso/master.

> While this is the only test using that account, it could be run in multiple contexts at the same time. And now that I'm thinking about it more I take back what I said earlier about having the gate for running the test just be the account itself being sufficient.

Mm you are right. For example a dev run could clash with a CI run. The nature of these jetpack connection tests means that sites will constantly be being added and removed. How about we have one account per context, something like `e2eflowtestingjetpackconnectCI`, `e2eflowtestingjetpackconnectDEV`?

seear:
_Ideally_ we would create a new wpcom account, maybe without a site, for each test, but I don't think that is currently possible.

seear:
Mobile sidebar menu problems fixed in 9e00e83.

Still an odd problem entering text in an input on mobile, investigating now.

seear:
Now passing at mobile width 🎉

hoverduck:
> Ideally we would create a new wpcom account, maybe without a site, for each test, but I don't think that is currently possible.

We have an entire suite of tests that create new wpcom accounts :smile: - https://github.com/Automattic/wp-e2e-tests/blob/master/specs/wp-signup-spec.js

Although all of the existing flows create a site along with the account, and they're wpcom-centric.  We should probably explore a Jetpack-centric flow that starts with the plugin installation and goes through account creation along with activation.

hoverduck:
I stumbled into an issue when the `JETPACKHOST` variable is set, which caused the getAccountConfig function to break down.  Since that envvar will be set to CI in the nominal use case (even though it won't be attached to _that_ dynamic site) I went ahead and just hard-coded a check in for `jetpackConnectUser`.

sirreal:
> _Ideally_ we would create a new wpcom account, maybe without a site, for each test, but I don't think that is currently possible.

@seear Are you referring to creating a user during testing, or the siteless signup? Siteless signup [exists, but is rather hidden](https://wordpress.com/start/account/user).

sirreal:
Just a note that this will need to be updated as soon as https://github.com/Automattic/wp-calypso/pull/19509 lands due to changes in the disconnect flow:

```
FAILED: Taking screenshot of: 'http://calypso.localhost:3000/settings/disconnect-site/foolish-fiordland.jurassic.ninja'

Jetpack Connect: (desktop) @jetpack Connect From Calypso: Can disconnect existing jetpack sites:
NoSuchElementError: no such element: Unable to locate element: {"method":"css selector","selector":".is-primary.is-scary"}
```

seear:
> Are you referring to creating a user during testing, or the siteless signup?

@sirreal I was thinking of creating a new user per-test to avoid the need to cleanup connected sites, but now you mention it, a site-less user should probably be one of the test cases anyway.

hoverduck:
> Just a note that this will need to be updated as soon as Automattic/wp-calypso#19509 lands due to changes in the disconnect flow:

Note that @brbrr fixed this in our _other_ disconnect script here - https://github.com/Automattic/wp-e2e-tests/pull/773

And of course I realize we should consolidate these at some point.

sirreal:
New AB test which affects this flow to take into account. Running the test I see:

```
Found an AB Testing key in local storage that isn't known: 'jetpackHidePlanIconsOnMobile'. This may cause inconsistent A/B test behaviour, please check this is okay and add it to 'knownABTestKeys' in default.config
```

See https://github.com/Automattic/wp-calypso/pull/19335

seear:
Rebased.

seear:
> Note that @brbrr fixed this in our other disconnect script here - #773

> And of course I realize we should consolidate these at some point.

Just rebased this PR, and it automatically takes advantage of @brbrr's change because it calls `SettingsPage.disconnectSite()`. Nice work all round 👍

PR #753, opened 2017-10-25
Remove unused pages

alisterscott:
Remove some pages which weren't referenced anywhere. Checked against both Jetpack and WPCom test suites - no issues


alisterscott:
@hoverduck just a heads-up in case this breaks anything Vis-diff related - it shouldn't though

hoverduck:
Thanks for cleaning these up.  We may end up returning some of the wp-admin pages with upcoming Jetpack work, but we can do that on a case-by-case basis

Issue #732, opened 2017-10-06
Add e2e test for domain only site


alisterscott:
A domain only site behaves differently in Calypso as most menus are hidden etc.

As these are often 'forgotten' ee could add an e2e tests that makes sure these still work correctly and don't act like a full site

PR #763, opened 2017-10-30
Adds a domain only signup

### LINKED ISSUE #759: Add e2e tests to the `/domains` flow
https://github.com/Automattic/wp-calypso/pull/18557 broke the `/domains` flow today - it was not until @hewsut reported (p1509110055000391-slack-domains-dev) it that we noticed the regression and @aidvu fixed it in https://github.com/Automattic/wp-calypso/pull/19226.

We plead thee e2e gods to shine upon this flow with your evertesting light 🙇

The simplest flow (similar to the regular domain purchase flow, you just don't pick the domain, themes... anything):
* Go to https://wordpress.com/start/domain-first/site-or-domain?new=random-domain.blog
* Click `Just buy a domain`.
* You are redirected to checkout with the domain in cart (note, however, that it's `https://wordpress.com/checkout/no-site` - meaning, at this point, it's not tied to any site - not sure if it changes anything for you, but wanted to point it out in case it does).

alisterscott:
Fixes #759

Depends upon data attributes in https://github.com/Automattic/wp-calypso/pull/19264

This adds a domain only signup scenario using a .live domain sandbox method

Also, this completes the business plan scenario using the same .live domain sandbox method

Should we add some tasks to the end of sign up to ensure calypso shows a domain only site? This would fix this issue: #732

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -56,10 +56,11 @@
     }
   },
   "httpsHosts": [ "WPCOM", "PRESSABLE" ],
-  "knownABTestKeys": [ "multiDomainRegistrationV1", "signupSurveyStep", "presaleChatButton", "businessPlanDescriptionAT", "newSiteWithJetpack", "postPublishConfirmation", "readerIntroIllustration", "paymentShowPaypalLogo", "jetpackConnectPlansCopyChanges", "postSignupUpgradeScreen", "skipThemesSelectionModal", "recommendShortestDomain", "checkoutPaymentMethodTabs", "unlimitedThemeNudge" ],
+  "knownABTestKeys": [ "multiDomainRegistrationV1", "signupSurveyStep", "presaleChatButton", "businessPlanDescriptionAT", "newSiteWithJetpack", "postPublishConfirmation", "readerIntroIllustration", "paymentShowPaypalLogo", "jetpackConnectPlansCopyChanges", "postSignupUpgradeScreen", "skipThemesSelectionModal", "recommendShortestDomain", "checkoutPaymentMethodTabs", "unlimitedThemeNudge", "gsuiteUpsell" ],
   "overrideABTests": [
 	[ "signupSurveyStep_20170329", "hideSurveyStep" ],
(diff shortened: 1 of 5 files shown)


alisterscott:
@hoverduck  so by using the cookie and a .live domain in this test it will be okay to complete the purchase? I checked out e2eflowtesting1509342157854858.live and it doesn't seem to be a live domain which is good ✅

alisterscott:
I think this fails on CircleCI as the purchase takes a realllllly long time to complete - maybe because it's a test domain?

klimeryk:
AFAIK, the actual payment step _is_ that long, see p99Zz8-4W-p2, for example:
> Payment is processing… for an excruciating minute and 49 seconds, why does paying with a credit card take so long?

Similar report in p99Zz8-51-p2:
> The only real issue that I noticed was that the “transaction” took about 25-30 seconds even while using credits.

Both reports are from production, too.

Of course, the payment step is also when the domain is provisioned (actually registered, etc.), so domains still could be the culprit. Maybe we could run a profiler (or just some simple debug prints) to figure out what's taking so long?

@rralian, do you know if there were reports of such long delays in the payments step? Maybe there's a way to speed things up, like using a different payment provider, etc.?

alisterscott:
@klimeryk I updated our business plan spec to include registering a .live domain using the same sandbox method and it completes the purchase, whereas using the domain flow I can't even get the test credit card payment to finish no matter how long I set it to wait. So maybe there is an issue with test payments for domains only.

alisterscott:
It looks like this causes issues on our back end systems, so please DO NOT merge or even run these until we know what's going on - see `p2-p2MSmN-6ev`

hoverduck:
The backend issue has been resolved (funds were added to our test environment, along with a cron to periodically watch for it to get low).  So this is no longer blocked on that.

klimeryk:
Yup, sorry about that - of course, it _had_ to happen just as we recommended using this test env to you 😅

PR #764, opened 2017-10-31
Node: Upgrade to v8 LTS carbon

sirreal:
Node 8 will be the [should become current LTS](https://github.com/nodejs/Release#release-schedule1) today. I expect [wp-calypso](https://github.com/Automattic/wp-calypso) will likely make the switch soon and it would be good for e2e tests to be ready as well.

✅  ~**BEWARE**: Consider this a task and in progress. Removing `node_modules` and running `npm install` breaks on the existing `postinstall` script.~


.nvmrc
```diff
@@ -1 +1 @@
-v6.11.5
+v8.9.1
```
sirreal: Need to move to `v8.9.3` now

package.json
```diff
@@ -10,7 +10,7 @@
   "main": "greetings.js",
   "defaultTestArgs": "-g -s $BROWSERSIZE",
   "scripts": {
-    "postinstall": "npm pack lib/reporter && npm install ./spec-xunit-reporter-0.0.1.tgz",
+    "postinstall": "npm pack ./lib/reporter && npm install ./spec-xunit-reporter-0.0.1.tgz",
```
sirreal: ❤️  Thanks! Never got around to figuring out why install was broken

alisterscott:
This PR looks good now and passes on CircleCI, however, when I run `npm install` I get unmodified files: package.json and package-lock.json. Is this expected? Should we commit these to this branch?

sirreal:
I've committed those changes. It looks like Node 8 decided to alphabetize dependencies, which seems nice to me 🙂 

Looking at b1910c72be004bee2f217a6b6e649c2e04abb697, I'd say it's 👌 

I think if this is working as expected now, we can make the move to node 8 🎉

alisterscott:
I'll investigate why this is now failing in Circle. Maybe it needs the `.` after all

alisterscott:
With the new npm version our post install script was adding the `spec-xunit-reporter` generated library to `package.json` which would cause clean installs to fail (since it hasn't been generated yet)

 ``` "scripts": {
    "postinstall": "npm pack ./lib/reporter && npm install ./spec-xunit-reporter-0.0.1.tgz",```




so I had to update the postinstall task to:


 ``` "scripts": {
    "postinstall": "npm pack ./lib/reporter && npm install ./spec-xunit-reporter-0.0.1.tgz --no-save",```

as npm now saves to package.json by default and this avoids it, so clean installs will now work and CircleCI should pass

That was a bit tricky to work out

alisterscott:
I've going to merge this as it's now working on CI

Issue #628, opened 2017-07-18
Docs: Split up the README into separate files


hoverduck:
The README is getting to be pretty unwieldy.  We should split it apart into separate files in a `/docs` directory with details on running the different types of tests and just link to them from the main README.

bsessions85:
Closed in #768

PR #768, opened 2017-11-01
Update/breakdown readme

bsessions85:
Split out the README to be more manageable.  #628 

I'm definitely open to suggestions on ways to make it better.

@rachelmcr @alisterscott @brbrr @hoverduck


alisterscott:
Looks good to me @bsessions85 

I like how all the hyperlinks work into the broken down files 💯

PR #772, opened 2017-11-02
Update the save selector for editor

alisterscott:
When running locally (calypso.localhost) and revision history is enabled the old selector doesn't work as expected


alisterscott:
@bsessions85 would you mind reviewing this (against a local calypso environment)?

bsessions85:
This looks good to me, I can watch it pass, even when the history link is there.

PR #776, opened 2017-11-03
Update node to 6.11.5

alisterscott:
To match `wp-calypso`


alisterscott:
Is this enough to refresh the docker image? It uses nvm right?

zinigor:
Should be good, and also should fix the Jetpack build.

Issue #690, opened 2017-08-30
Delete dynamically created Jetpack sites after test failures


hoverduck:
The Jetpack CI tests are built using the CircleCI Workflows capability with three tasks:

1) Build - Create / connect the site
2) Test - Execute the e2e tests
3) Destroy - Disconnect / delete the site

It works great as long as the tests are successful, but if they fail the Destroy task is never started.  Workflows lacks a capability to run a task even on failure.

This issue is to track the problem and manage the site deletion steps elsewhere, but keep in mind that the Test task is run on desktop/mobile in parallel, so we can't delete the site until both pieces are done.

brbrr:
I've done some investigation on this:

- CircleCI does not support any sort of setup/teardown flows. 
- Magellan does have global [teardown](https://github.com/TestArmada/magellan#setup-and-teardown) support, with which we could solve this issue. 

I'll move forward and try this out

brbrr:
This should be fixed in #911

alisterscott:
@brbrr Can we close this now?

brbrr:
With pleasure :)

PR #784, opened 2017-11-06
Try/jetpack failed tests cleanup magellan

brbrr:
Switching from CircleCI workflows to Magellan setup/teardown flow #690 

This is WIP because this should be run against JetPack tests


alisterscott:
@brbrr are you planning to continue with this one?

brbrr:
Since we planning to switch to jurassic.ninja -this could be closed

PR #791, opened 2017-11-08
AB Tests: Add known test jetpackHidePlanIconsOnMobile to default config

sirreal:
`jetpackHidePlanIconsOnMobile` was added in https://github.com/Automattic/wp-calypso/pull/19335

This PR adds it to the known tests.

It also formats and alphabetizes the list of known tests to make future additions and deletions less invasive (smaller diffs).


sirreal:
The CI failure can be ignored and is because this repo is supposed to run in a special context with wp-calypso, is that correct?

PR #796, opened 2017-11-09
Introduce a test for Jetpack Connect starting from wp-admin

tyxla:
This PR adds a test for connecting a .org site running jetpack, starting from wp-admin, which is the general flow according to p7rd6c-132-p2. 

Uses the recently introduced `wporg-creator-page` to create a .org site with Jetpack installed (using the jurassic.ninja service). Right now it creates a new site, but we might want to consider creating just 1 site and running all tests against it.

Requires the user `e2eflowtestingjetpackconnect` for logging into WordPress.com.

### DIFF
--- a/lib/flows/login-flow.js
+++ b/lib/flows/login-flow.js
@@ -174,7 +174,7 @@ export default class LoginFlow {
 	loginUsingExistingForm() {
 		let testUserName, testPassword;
 
-		const accountInfo = LoginFlow.getAccountConfig( this.account );
+		const accountInfo = dataHelper.getAccountConfig( this.account );
 
 		if ( accountInfo !== undefined ) {
 			testUserName = accountInfo[0];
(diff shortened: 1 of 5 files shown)


seear:
>  Right now it creates a new site, but we might want to consider creating just 1 site and running all tests against it.

Multiple sites is best I think, because ideally we will mark these @parallel so they can run more often.

seear:
> The one thing I don't understand is why the CI test run has so many failures on this branch

I've updated `NODE_CONFIG` to add `jetpackConnectUser`, and tried rebuilding without the cache. Still seeing failures.

seear:
Rebased!

tyxla:
Woot, thank you @seear 💯 👍

tyxla:
This one looks ready from mine and Steve's perspective, could you have a look before we merge it @hoverduck? Thanks!

tyxla:
Thanks @seear and @hoverduck 👍

PR #800, opened 2017-11-09
Introduce test for Jetpack Connect with Jetpack not installed

tyxla:
This PR introduces a test for Jetpack Connect with Jetpack not installed.

Relies on #806 to spin up a jurassic.ninja site without Jetpack preinstalled.


seear:
rebased

seear:
I've tweaked this to make the `addNewSite()` method more robust so that it now works with any number of existing sites. This means the new test is not relying on previous tests, and we will be able to move to parallel testing more easily.

tyxla:
Nice one @seear. Your change looks good to me, and works well. Thanks 👍

PR #803, opened 2017-11-13
Fix Hello Dolly selector and account config selector for connect flow

hoverduck:


alisterscott:
Thanks for the change

PR #805, opened 2017-11-14
Try simplifying login

alisterscott:
Let's see if this works


alisterscott:
@hoverduck can you see any issues with removing this? I'm guessing the magellan auto-retry will capture any issues we have and I haven't seen this message is Slack recently

PR #809, opened 2017-11-16
Jetpack: Fix unreliable site disconnects in the Jetpack connect tests

seear:
Move site disconnects into a **separate test that disconnects any expired sites**. Disconnecting broken sites can be done from the site switcher and should be more reliable than going through the settings page.

Theoretically this change should allow all the jetpack connect tests in this spec to be run in parallel, because there is no danger of a live site being disconnected.

**Testing**
There should never be any expired sites in the `jetpackConnectUser` account because on the master branch all sites are disconnected in one of the tests. This PR can be tested, slowly, by using a different account, adding some `jurassic.ninja/create?shortlived` sites, and waiting an hour+ for them to expire.

/cc @sirreal

### DIFF
--- a/specs-jetpack-calypso/wp-jetpack-connect-spec.js
+++ b/specs-jetpack-calypso/wp-jetpack-connect-spec.js
@@ -65,7 +65,6 @@ test.describe( `Jetpack Connect: (${ screenSize }) @jetpack`, function() {
 						driver,
 						By.css( '.notice.is-dismissable .notice__dismiss' )
 					);
-					console.log( 'going to remove sites again' );
 					removeSites();
 				} );
 			};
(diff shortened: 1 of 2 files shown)


sirreal:
I tested this with 1 expired site and 0 expired sites, both work well. This should be a nice improvement.

I'm waiting for some more sites to expire, I'll give more feedback for n expired sites soon.

specs-jetpack-calypso/wp-jetpack-connect-spec.js
```diff
@@ -35,6 +35,45 @@ test.before( function() {
...
+					// seems like it is not waiting for this
+					driverHelper.waitTillPresentAndDisplayed(
+						driver,
+						By.css( '.notice.is-success.is-dismissable' )
+					);
+					driverHelper.clickWhenClickable(
+						driver,
+						By.css( '.notice.is-dismissable .notice__dismiss' )
+					);
+					console.log( 'going to remove sites again' );
```
sirreal: Need to clean up

tyxla:
Nice one @seear 👍

PR #814, opened 2017-11-21
Run jetpack connect tests in parallel

seear:
Now that they no longer depend on disconnecting sites before adding new sites, they can run in parallel.

### DIFF
--- a/specs-jetpack-calypso/wp-jetpack-connect-spec.js
+++ b/specs-jetpack-calypso/wp-jetpack-connect-spec.js
@@ -32,10 +32,10 @@ test.before( function() {
 	driver = driverManager.startBrowser();
 } );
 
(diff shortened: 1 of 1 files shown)


seear:
> the @parallel tag isn't needed here

@hoverduck does this PR have any effect then, or would these tests have been split across the workers already?

seear:
As I understand it, with the `@jetpack` tag moved down to the individual tests, these are now run separately and in parallel, whereas before they were run all together one after the other.

Keeping the `@parallel` tag because we may want to run these tests with the other Calypso tests at some point.

hoverduck:
> As I understand it, with the @jetpack tag moved down to the individual tests, these are now run separately and in parallel, whereas before they were run all together one after the other.

That's right.

Issue #702, opened 2017-09-05
Replace bluecat with WP


alisterscott:
We currently use this third party library for REST API calls:

https://www.npmjs.com/package/bluecat

We should be using our own node client to do the REST API stuff: 

https://www.npmjs.com/package/wpcom

bsessions85:
Done in PR #815

PR #815, opened 2017-11-21
#702 Using wpcom instead of bluecat for api calls

bsessions85:
I updated this to use wpcom instead of bluecat for api calls. It is still using bluecat to get the oauth token.
#702


bsessions85:
@rachelmcr or @alisterscott can you please take a look at this one?

PR #817, opened 2017-11-23
Remove email verification test from signup spec

rachelmcr:
This PR removes the email verification test from the signup spec (originally added in #426).

The verification email is no longer consistently sent immediately after signup which is causing the test to fail. This change in behavior is part of an a/b test, but not a Calypso a/b test — I can't find a way to avoid the a/b test within our signup spec, and I was advised that some delay is likely to be adopted (ref: p91bmw-da-p2, p1511453214000236-slack-e2e-testing-discuss). We can revisit if that doesn't happen.

I left the test that confirms the user can't publish before email verification, as that test isn't affected by these changes. If we want to test the ability to publish after email verification, we'll need to revisit it and handle the delay once that delay is finalized. cc @hoverduck

### DIFF
--- a/specs/wp-signup-spec.js
+++ b/specs/wp-signup-spec.js
@@ -205,7 +205,8 @@ testDescribe( `[${host}] Sign Up  (${screenSize}, ${locale})`, function() {
 
 										test.it( 'Can see a the magic link email', function() {
 											return this.emailClient.pollEmailsByRecipient( emailAddress ).then( function( emails ) {
-												assert.equal( emails.length, 2, 'The number of newly registered emails is not equal to 2 (activation and magic link)' );
+												//Disabled due to a/b test on activation email. See https://github.com/Automattic/wp-e2e-tests/issues/819
+												//assert.equal( emails.length, 2, 'The number of newly registered emails is not equal to 2 (activation and magic link)' );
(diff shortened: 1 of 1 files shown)


rachelmcr:
The `Sign up for a free site and log in via a magic link` test also needs to be updated, as it checks for 2 emails (activation and magic link) before proceeding.

rachelmcr:
I'm going to go ahead and merge this to address the failing tests, but I can revisit if there are any concerns about these changes or better ways to handle this a/b test while it's still running.

hoverduck:
Thanks for handling this @rachelmcr.  Sounds like removing the test altogether is the way to go.

What happens if we signup, attempt to publish (and are blocked)...is there a prompt there to "resend" the Activation e-mail (which may not have actually been sent at all yet)?  We could maybe use that as the mechanism to force the e-mail to send early.

rachelmcr:
>What happens if we signup, attempt to publish (and are blocked)...is there a prompt there to "resend" the Activation e-mail (which may not have actually been sent at all yet)?

There is some discussion on the a/b test about the various actions that are blocked before email verification, and a suggestion to trigger the email when you hit one of these actions. I'm following that discussion and once it's resolved I'll circle back to this test to see what makes the most sense.

PR #823, opened 2017-11-28
Write url to screenshot and post it on Slack

akirk:
With this change the current URL is written as metadata to the screenshot. That makes it possible to output it as a comment when posting it to Slack.

### DIFF
--- a/lib/after.js
+++ b/lib/after.js
@@ -35,27 +35,38 @@ test.afterEach( function() {
 	const locale = driverManager.currentLocale().toUpperCase();
 
 	if ( this.currentTest.state === 'failed' ) {
-		const prefix = `FAILED-${locale}-${screenSize}-${shortTestFileName}`;
-
 		let neverSaveScreenshots = config.get( 'neverSaveScreenshots' );
 		if ( neverSaveScreenshots ) {
 			return null;
 		}
 
-		driver.getCurrentUrl().then( ( url ) => console.log( `FAILED: Taking screenshot of: '${url}'` ), ( err ) => {
+		driver.getCurrentUrl().then( url => console.log( `FAILED: Taking screenshot of: '${url}'` ), ( err ) => {
 			slackNotifier.warn( `Could not capture the URL when taking a screenshot: '${err}'` );
 		} );
 
+
 		return driver.takeScreenshot().then( ( data ) => {
-			mediaHelper.writeScreenshot( data, prefix );
+			driver.getCurrentUrl().then( url => {
+				mediaHelper.writeScreenshot( data, function( date ) {
+					return `FAILED-${locale}-${screenSize}-${shortTestFileName}-${date}`;
+				}, {
+					url
+				} );
+			} );
 		}, ( err ) => {
 			slackNotifier.warn( `Could not take screenshot due to error: '${err}'` );
 		} );
 	}
+
 	if ( config.get( 'saveAllScreenshots' ) === true ) {
-		const prefix = [ `${locale}-${screenSize}-${shortDescribeFileName}`, `${shortTestFileName}` ];
 		return driver.takeScreenshot().then( ( data ) => {
-			mediaHelper.writeScreenshot( data, prefix, true );
+			driver.getCurrentUrl().then( url => {
+				mediaHelper.writeScreenshot( data, function( date ) {
+					return `${locale}-${screenSize}-${shortDescribeFileName}-${date}-${shortTestFileName}.png`;
+				}, {
+					url
+				} );
+			} );
 		}, ( err ) => {
 			slackNotifier.warn( `Could not take screenshot due to error: '${err}'` );
 		} );
--- a/lib/after.js
+++ b/lib/after.js
@@ -20,7 +20,7 @@ test.afterEach( function() {
 } );
 
 // Take Screenshot
-test.afterEach( function() {
+test.afterEach( function( done ) {
 	this.timeout( afterHookTimeoutMS );
 	if ( ! this.currentTest ) {
 		return;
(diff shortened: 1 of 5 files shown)


alisterscott:
Sorry I didn't get to reviewing this today - and I think @hoverduck is AFK too
@brbrr and/or @bsessions85: would you be able to review this please?

lib/reporter/magellan-reporter.js
```diff
@@ -142,19 +143,23 @@ Reporter.prototype.listenTo = function( testRun, test, source ) {
...
-									title: `${screenshotPath} - # ${process.env.CIRCLE_BUILD_NUM}`,
-									channels: config.get( 'slackChannelForScreenshots' )
-								}, function( err ) {
-									if ( err ) {
-										slackClient.send( {
-											icon_emoji: ':a8c:',
-											text: `Build <https://circleci.com/gh/${process.env.CIRCLE_PROJECT_USERNAME}/${process.env.CIRCLE_PROJECT_REPONAME}/${process.env.CIRCLE_BUILD_NUM}|#${process.env.CIRCLE_BUILD_NUM}> Upload to slack failed: '${err}'`,
-											username: 'e2e Test Runner'
+								fs.createReadStream( `${screenshotDir}/${screenshotPath}` )
+									.pipe( pngitxt.get( 'url', function ( url ) {
```
akirk: I wonder if this could be solved better but I'll need some input on this.

bsessions85: I don't have any better suggestions, but perhaps @brbrr or @alisterscott does?

bsessions85:
This looks fine to me, but I agree that the code mentioned above should be cleaned up if anyone has any ideas.

lib/reporter/magellan-reporter.js
ramonjd: I can't see any obvious ways to improve this either. 

It looks to be going against the '__spirit__' of the file's contents in general, but maybe the solution is just to abstract it a bit, and make it more readable (?)

```
function slackUploadFileCallback( err, data ) {
    if ( err ) {
        slackClient.send( {
            ...args
        } );
    }
}
function getStreamCallback() {
    return pngitxt.get( 'url', function ( url ) {
        slackUpload.uploadFile( { ...args }, slackUploadFileCallback );
    } );
}

fs.createReadStream( `${screenshotDir}/${screenshotPath}` )
    .pipe( getStreamCallback() );


```
None of the library methods returns a promise, I don't think that would make life easier in this case anyway.

lib/after.js
```diff
@@ -35,27 +35,38 @@ test.afterEach( function() {
...
 		} );
 	}
+
 	if ( config.get( 'saveAllScreenshots' ) === true ) {
-		const prefix = [ `${locale}-${screenSize}-${shortDescribeFileName}`, `${shortTestFileName}` ];
 		return driver.takeScreenshot().then( ( data ) => {
-			mediaHelper.writeScreenshot( data, prefix, true );
+			driver.getCurrentUrl().then( url => {
+				mediaHelper.writeScreenshot( data, function( date ) {
+					return `${locale}-${screenSize}-${shortDescribeFileName}-${date}-${shortTestFileName}`;
```
brbrr: If `${date}` is used in both file names (here & on line 51) maybe it worth to put it right after other "static" values, so it would be easier to sort them etc?

lib/reporter/magellan-reporter.js
```diff
@@ -142,19 +143,23 @@ Reporter.prototype.listenTo = function( testRun, test, source ) {
...
+											initialComment: url,
+											channels: config.get( 'slackChannelForScreenshots' )
+										}, function( err ) {
+											if ( err ) {
+												slackClient.send( {
+													icon_emoji: ':a8c:',
+													text: `Build <https://circleci.com/gh/${process.env.CIRCLE_PROJECT_USERNAME}/${process.env.CIRCLE_PROJECT_REPONAME}/${process.env.CIRCLE_BUILD_NUM}|#${process.env.CIRCLE_BUILD_NUM}> Upload to slack failed: '${err}'`,
+													username: 'e2e Test Runner'
+												} );
+											}
```
brbrr: Using guard case together with arrow function can make it look a bit nicer:

```
}, ( err ) => {
	if ( !err ) return
	slackClient.send( {
		icon_emoji: ':a8c:',
		text: `Build <https://circleci.com/gh/${process.env.CIRCLE_PROJECT_USERNAME}/${process.env.CIRCLE_PROJECT_REPONAME}/${process.env.CIRCLE_BUILD_NUM}|#${process.env.CIRCLE_BUILD_NUM}> Upload to slack failed: '${err}'`,
		username: 'e2e Test Runner'
	} );
} );
```

lib/after.js
```diff
@@ -33,33 +33,29 @@ test.afterEach( function() {
...
-		} );
-	}
-	if ( config.get( 'saveAllScreenshots' ) === true ) {
-		const prefix = [ `${locale}-${screenSize}-${shortDescribeFileName}`, `${shortTestFileName}` ];
-		return driver.takeScreenshot().then( ( data ) => {
-			mediaHelper.writeScreenshot( data, prefix, true );
-		}, ( err ) => {
-			slackNotifier.warn( `Could not take screenshot due to error: '${err}'` );
-		} );
+		filenameCallback = () => `FAILED-${locale}-${screenSize}-${shortTestFileName}-${date}`;
```
sirbrillig: Since this variable is never defined with `let`, `var`, or `const`, I think it becomes a global variable, which is probably not what you want.

akirk: Indeed, thanks! (I just quickly pushed this WIP file up for review)

akirk:
@brbrr I changed this according to your suggestion, is it ok now?

brbrr:
Thanks, @akirk!

PR #824, opened 2017-11-28
Cancel domain purchases

alisterscott:
Even though the e2e tests use sandboxed domains it's still a good idea to cancel these.

This adds (lots of) steps to calypso to cancel these out


klimeryk:
I'm guessing the tests are failing right now because of https://github.com/Automattic/wp-calypso/issues/20274?

alisterscott:
It was actually failing due to guided tours blocking the success message - I now dismiss guided tours if it appears and this now works 👍

alisterscott:
Pushed a fix for the survey which now requires answers: https://github.com/Automattic/wp-calypso/pull/20320

PR #833, opened 2017-12-05
Added AB test signupSiteSegmentStep

hoverduck:


alisterscott:
Introduced here: https://github.com/Automattic/wp-calypso/pull/20369

PR #839, opened 2017-12-07
Skip the domain purchase/cancellation steps

hoverduck:
The domain purchases are timing out due to issues with the test environment (potentially we haven't been cancelling the domains like we thought).  Disabling this part of the test until we figure out what's going on.


hoverduck:
Ref p1512666701000812-slack-domains-dev

PR #845, opened 2017-12-12
New way to detect infinite list loading

hoverduck:
@alisterscott Do you mind taking a look at this when you get back from AFK?  I'm going to go ahead and merge it since it appears to be working fine, but I'm curious for your opinion.

The driver was the new condensed posts list, which changed the selector for the placeholder elements we were watching for to ensure the list was loaded.  And now the list loads in a way such that we _always_ see a placeholder element at the very bottom, offscreen.

So this PR adds a new helper function that just takes a selector and looks for a certain number of that element to be present.  It defaults to 10 elements, which seems to be enough to say that the list is populated.

### DIFF
--- a/lib/pages/posts-page.js
+++ b/lib/pages/posts-page.js
@@ -1,4 +1,4 @@
-import webdriver, { By } from 'selenium-webdriver';
+import { By } from 'selenium-webdriver';
 import BaseContainer from '../base-container.js';
 
 import * as driverHelper from '../driver-helper';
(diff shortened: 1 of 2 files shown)


lib/pages/posts-page.js
```diff
@@ -10,12 +10,8 @@ export default class PostsPage extends BaseContainer {
 
 	waitForPosts() {
 		const driver = this.driver;
-		const resultsLoadingSelector = By.css( '.posts__list .is-placeholder:not(.post-image)' );
-		driver.wait( function() {
-			return driverHelper.isElementPresent( driver, resultsLoadingSelector ).then( function( present ) {
-				return ! present;
-			} );
-		}, this.explicitWaitMS, 'The post results loading element was still present when it should have disappeared by now.' );
```
nylen: We lose the timeout and the custom message here.  How important are those?

alisterscott:
@hoverduck I had a look at this. I can't really think of a way to deterministically wait for the placeholder not to appear because as you pointed out there's always one at the bottom if there are more posts in the list. It's also an issue when there's only one or two posts as the placeholder doesn't seem to load.

I think your solution is neat but I think it will have a problem when we periodically clear out all the posts there won't be 10 posts to wait for (until we get to 10 which _should_ happen pretty quickly) but we will have potentially confusing results until we get there.

With the condensed list it seems faster so I'm not sure we even need the waiting for the placeholders to disappear anymore. I've put together a PR that just waits for the post title (which we know and doesn't change for different cultures) to appear in the list - I think this should work 🤞 

https://github.com/Automattic/wp-e2e-tests/pull/857

PR #851, opened 2017-12-20
Use about page instead of design type in signup tests

markryall:
The new about page step has been declared a winner in a signup a/b test as an alternative to the design type step.

Currently, 100% of en users are seeing this step instead of design type.

This change ensures the signup tests will be prepared for the likely permanent replacement of the design type step with this new step.


lib/pages/signup/about-page.js
```diff
@@ -0,0 +1,15 @@
+import { By } from 'selenium-webdriver';
+import * as driverHelper from '../../driver-helper.js';
+import * as driverManager from '../../driver-manager.js';
```
alisterscott: Is this line necessary?

alisterscott:
hotdog

PR #855, opened 2017-12-21
Add more cases for passwordless login

Tug:
This PR adds a simple test case for passwordless login on WordPress.com.

It also introduces a new format for defining and using test accounts.
The main idea is to have the ability to define multiple accounts for the same "feature". For instance, here we'd like to test passwordless login, we can then define multiple accounts that have have the `"passworless"` feature:
```json
{
  "accounts": [
	{
	  "email": "passwordless1@gmail.com",
	  "features": [ "passwordless" ]
	},
	{
	  "email": "passwordless2@gmail.com",
	  "features": [ "passwordless" ]
	},
	{
	  "email": "passwordless3@gmail.com",
	  "features": [ "passwordless" ]
	},
	{
	  "email": "passwordless4@gmail.com",
	  "features": [ "passwordless" ]
	}
  ]
}
```

When running the passwordless tests, one of the accounts will be used at a time (most probably) so we should avoid triggering 2 magic link emails for the same account at the same time.

### Testing Instructions
- Apply this patch and use pb/1cc8a for your `config/local-development.json`
- Boot wp-calypso locally
- Run the testsuite: `env BROWSERSIZE=mobile ./node_modules/.bin/mocha specs/wp-log-in-out-spec.js`
- Assert that all the tests pass

### Reviews
- [ ] Product
- [ ] Code

### DIFF
--- a/lib/data-helper.js
+++ b/lib/data-helper.js
@@ -1,6 +1,6 @@
 import phrase from 'asana-phrase';
 import config from 'config';
-import { map } from 'lodash';
+import { difference, map } from 'lodash';
 
 String.prototype.toProperCase = function() {
 	return this.replace( /\w\S*/g, function( txt ) {
@@ -86,6 +86,31 @@ export function getAccountConfig( account ) {
 	}
 
 	return localConfig[ account ];
+}
+
+export function getAllAccountsWithFeatures( features = [] ) {
+	let allAccounts = [];
+
+	if ( config.has( 'accounts' ) ) {
+		allAccounts = config.get( 'accounts' );
+	} else if ( process.env.ACCOUNT_INFO ) {
+		allAccounts = JSON.parse( process.env.ACCOUNT_INFO );
+	}
+
+	return allAccounts.filter( function( account ) {
+		// return accounts that have all the features requested
+		return difference( features, account.features || [] ).length === 0;
+	} );
+}
+
+export function getRandomAccountWithFeatures( features = [] ) {
+	if ( ! Array.isArray( features ) ) {
+		features = [ features ];
+	}
+
+	const accounts = getAllAccountsWithFeatures( features );
+
+	return accounts[ Math.floor( Math.random() * accounts.length ) ];
 }
 
 export function getJetpackSiteName() {
--- a/lib/data-helper.js
+++ b/lib/data-helper.js
@@ -111,14 +111,28 @@ export function hasAccountWithFeatures( features = [] ) {
 	return getAllAccountsWithFeatures( features ).length > 0;
 }
 
-export function getRandomAccountWithFeatures( features = [] ) {
+export function pickRandomAccountWithFeatures( features = [] ) {
 	if ( ! Array.isArray( features ) ) {
 		features = [ features ];
 	}
 
-	const accounts = getAllAccountsWithFeatures( features );
+	const accounts = getAllAccountsWithFeatures( features ).filter( function( account ) {
+		return ! account.inUse;
+	} );
 
-	return accounts[ Math.floor( Math.random() * accounts.length ) ];
+	if ( accounts.length === 0 ) {
+		return null;
+	}
+
+	const account = accounts[ Math.floor( Math.random() * accounts.length ) ];
+	account.inUse = true;
+
+	return account;
+}
+
+
+export function releaseAccount( account ) {
+	delete account.inUse;
 }
 
 export function getJetpackSiteName() {
--- a/lib/data-helper.js
+++ b/lib/data-helper.js
@@ -103,6 +103,14 @@ export function getAllAccountsWithFeatures( features = [] ) {
 	} );
 }
 
+export function hasAccountWithFeatures( features = [] ) {
+	if ( ! Array.isArray( features ) ) {
+		features = [ features ];
+	}
+
+	return getAllAccountsWithFeatures( features ).length > 0;
+}
+
 export function getRandomAccountWithFeatures( features = [] ) {
 	if ( ! Array.isArray( features ) ) {
 		features = [ features ];
--- a/lib/email-client.js
+++ b/lib/email-client.js
@@ -6,7 +6,7 @@ const emailWaitMS = config.get( 'emailWaitMS' );
 export default class EmailClient {
 	constructor( mailboxId ) {
 		const apiKey = config.get( 'mailosaurAPIKey' );
-		const Mailosaur = require( 'mailosaur' )( apiKey, 'https://www.mailosaur.com/api/' );
+		const Mailosaur = require( 'mailosaur' )( apiKey, 'https://mailosaur.com/api/' );
 		this.mailbox = new Mailosaur.Mailbox( mailboxId );
 	}
 
--- a/lib/email-client.js
+++ b/lib/email-client.js
@@ -70,4 +70,19 @@ export default class EmailClient {
 
 		return d.promise;
 	}
+
+	getNewEmailsByRecipient( emailAddress ) {
+		var self = this;
+		var d = webdriver.promise.defer();
+		this.pollEmailsByRecipient( emailAddress ).then( function( emails ) {
+			self.deleteAllEmail().then( function() {
+				d.fulfill( emails );
+			}, function( err ) {
+				d.reject( err );
+			} );
+		}, function( err ) {
+			d.reject( err );
+		} );
+		return d.promise;
+	}
 }
--- a/lib/email-client.js
+++ b/lib/email-client.js
@@ -70,19 +70,4 @@ export default class EmailClient {
 
 		return d.promise;
 	}
-
-	getNewEmailsByRecipient( emailAddress ) {
-		var self = this;
-		var d = webdriver.promise.defer();
-		this.pollEmailsByRecipient( emailAddress ).then( function( emails ) {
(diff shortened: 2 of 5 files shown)


hoverduck:
I worry that even with the approach you're taking here to routinely clear out the inbox on every query we'll run into timing races when this test is run in parallel.  At a minimum we have a potential for 2 runs at the same time since we do both mobile and desktop screen width, and even beyond that we frequently have multiple builds on CI at the same time.

That's why we opted for the original magic link test to be done on signup, since that's the only way we could guarantee a unique e-mail address that would only receive a single message.  https://github.com/Automattic/wp-e2e-tests/blob/master/specs/wp-signup-spec.js#L65

Is running this test here going to add anything over testing magic links on the signup spec?

alisterscott:
> That's why we opted for the original magic link test to be done on signup, since that's the only way we could guarantee a unique e-mail address that would only receive a single message.

Could we add some meta-data to the magic link email that is sent from WordPress.com that has a GUID or something that we also hide in the Calypso log-in page, so we poll the emails to find the email with the meta-data that we expect which is unique so we don't get clashes even using the same account? Or now I think about that approach do subsequent magic email links immediately invalidate the previous ones?

Tug:
I see, that's a challenging problem!

> Could we add some meta-data to the magic link email that is sent from WordPress.com that has a GUID or something that we also hide in the Calypso log-in page

That would work but I feel it adds quite a lot of complexity. I guess we would need to add it to the email header so it doesn't appear even for text emails...

> Or now I think about that approach do subsequent magic email links immediately invalidate the previous ones?

No they don't invalidate, but they have a validity limit anyway.

How about having multiple test accounts and choose one randomly at the beginning of the test?

Tug:
> Is running this test here going to add anything over testing magic links on the signup spec?

Yes, we'd like to test all login cases, make sure the `redirect_to` parameter is honoured (for jetpack connect for instance), 2fa screen is shown for 2fa users, oauth2 flows etc.

lib/flows/login-flow.js
```diff
@@ -12,28 +12,26 @@ import * as dataHelper from '../data-helper';
...
 
 export default class LoginFlow {
-	constructor( driver, account ) {
+	constructor( driver, accountOrFeatures ) {
 		this.driver = driver;
-
-		if ( account ) {
-			this.account = account;
+		accountOrFeatures = accountOrFeatures || 'defaultUser';
+		if ( typeof accountOrFeatures === 'string' ) {
```
Tug: I don't want to break backward compatibility with existing config so we should make sure any additional code works for both types of account config.

lib/data-helper.js
```diff
@@ -88,6 +88,39 @@ export function getAccountConfig( account ) {
...
+
+	if ( config.has( 'accounts' ) ) {
+		allAccounts = config.get( 'accounts' );
+	} else if ( process.env.ACCOUNT_INFO ) {
+		allAccounts = JSON.parse( process.env.ACCOUNT_INFO );
+	}
+
+	return allAccounts.filter( function( account ) {
+		// return accounts that have all the features requested
+		return difference( account.features || [], features ).length === 0;
```
yurynix: Is that what you wanted here?
```js
_.difference(['passwordless', '2fa'], ['passwordless']); // => [ '2fa' ]
```

shouldn't `features` be a subset of `account.feature` ?

Tug: Nice catch!, thx!

lib/data-helper.js
```diff
@@ -88,6 +88,39 @@ export function getAccountConfig( account ) {
...
+}
+
+export function getRandomAccountWithFeatures( features = [] ) {
+	if ( ! Array.isArray( features ) ) {
+		features = [ features ];
+	}
+
+	const accounts = getAllAccountsWithFeatures( features );
+
+	return accounts[ Math.floor( Math.random() * ( accounts.length -1 ) ) ];
```
yurynix: Shouldn't -1 here, since [Math.random()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random) max is 0.9999999
> The Math.random() function returns a floating-point, pseudo-random number in the range 0 to 1 — that is, from 0 (inclusive) up to but not including 1 (exclusive)

so, for [1,2] you'll get `Math.floor( 0.999999 * 1 )`, never 1

lib/email-client.js
```diff
@@ -6,7 +6,7 @@ const emailWaitMS = config.get( 'emailWaitMS' );
 export default class EmailClient {
 	constructor( mailboxId ) {
 		const apiKey = config.get( 'mailosaurAPIKey' );
-		const Mailosaur = require( 'mailosaur' )( apiKey, 'https://www.mailosaur.com/api/' );
+		const Mailosaur = require( 'mailosaur' )( apiKey, 'https://mailosaur.com/api/' );
```
yurynix: Why this change necessary? Wasn't it working before?

lib/data-helper.js
Tug: You're right

lib/email-client.js
Tug: Surprisingly, delete requests returns a 301 redirect and thus fail for `https://www.mailosaur.com/api/` (they were not used in the code until now). Using `https://www.mailosaur.com/api/` it works fine.

lib/email-client.js
```diff
@@ -70,4 +70,19 @@ export default class EmailClient {
 
 		return d.promise;
 	}
+
+	getNewEmailsByRecipient( emailAddress ) {
```
yurynix: Super minor, feel free to ignore how about something like that:
```js
getNewEmailsByRecipient( emailAddress ) {
	var self = this;
	return this.pollEmailsByRecipient( emailAddress )
		.then( function( emails ) {
			return self.deleteAllEmail().then( function () { 
				return emails;
			} );
	         } );
} );
```

or we don't care about delete errors:

```js
getNewEmailsByRecipient( emailAddress ) {
	var self = this;
	return this.pollEmailsByRecipient( emailAddress )
		.then( function( emails ) {
		      self.deleteAllEmail();
                      return emails;
	         } );
} );
```

also, why can't we use arrow functions here? we seems to use the ES6 export / import syntax here?

specs/wp-log-in-out-spec.js
```diff
@@ -83,8 +87,63 @@ test.describe( `[${host}] Authentication: (${screenSize}) @parallel @jetpack`, f
...
+				let magicLoginLink, loginFlow, magicLinkEmail, emailClient;
+				test.before( function () {
+					loginFlow = new LoginFlow( driver, [ 'passwordless' ] );
+					emailClient = new EmailClient( get( loginFlow.account, 'mailosaur.inboxId' ) );
+					return loginFlow.login();
+				} );
+
+				test.it( 'Can find the magic link in the email received', function() {
+					return emailClient.pollEmailsByRecipient( loginFlow.account.email ).then( function( emails ) {
+						for ( let email of emails ) {
```
yurynix: Can we maybe
```js
const selectedEmail = emails.find( email.subject.indexOf( 'WordPress.com' ) > -1 ); // or lodash
magicLinkEmail = selectedEmail;
magicLoginLink = selectedEmail.html.links[0].href;
```

yurynix:
Thank you for working on that @Tug ! 👍 

@hoverduck mentioned:
>At a minimum we have a potential for 2 runs at the same time since we do both mobile and desktop screen width, and even beyond that we frequently have multiple builds on CI at the same time.

From that I understand that multiple test runners might run at the same time on different machines / processes, so the locking and random picking mechanism you've made at `pickRandomAccountWithFeatures()` won't going to solve any collision issues and just makes this unnecessary complicated, but I'm really a noob at this code base, so I might be really mistaken here.

In case I'm not, maybe we should go with a different approach, to assign each testing driver a uuid on the user agent:
```
const selenium = require('selenium-webdriver');
const chrome = require('selenium-webdriver/chrome');
const uuidv4 = require('uuid/v4');

const opts = new chrome.Options();
opts.addArguments([`user-agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/63.0.3239.84 Safari/537.36 e2e_id/${ uuidv4() }"`]);

var driver = new selenium.Builder().
    withCapabilities(opts.toCapabilities()).
    build();
```
This example is for chrome, but similar can be get done [for FF](https://stackoverflow.com/questions/29916054/change-user-agent-for-selenium-driver)

Then we'll get this `e2e_id` on the API and include it in a mail header, which we'll later extract to identify which email we should use.

That's quite complicated, so I would've liked to avoid that if possible, first maybe we should really verify we have really a problem with collisions, maybe start with merging this PR and log to slack channel when test really failed?

What's your thoughts on that @hoverduck @alisterscott ?

hoverduck:
>  I understand that multiple test runners might run at the same time on different machines / processes

That's correct.  So yeah, the account.inUse locking mechanism here won't really come into play.  Ideally I would like to have a truly robust solution in place, like the e2e_id in the e-mail header...but that's also probably unneeded complexity.

Just relying on pure randomness to avoid collisions will _probably_ be okay.  The tests are auto-retried once, so we'd actually have to have a collision two separate times in order to fail the whole suite, so the odds are probably in our favor.

PR #857, opened 2017-12-22
Try waiting for post title in list of posts

alisterscott:
This should simply waiting for the post in the posts list.

### DIFF
--- a/lib/pages/posts-page.js
+++ b/lib/pages/posts-page.js
@@ -8,22 +8,18 @@ export default class PostsPage extends BaseContainer {
 		super( driver, By.css( '.posts' ) );
 	}
 
-	waitForPosts() {
-		const driver = this.driver;
-		const postSelector = By.css( '.post-item__panel:not( .is-placeholder )' );
-		return driverHelper.waitForInfiniteListLoad(
-			driver,
-			postSelector,
-			{ numElements: 1 }
-		);
+	waitForPostTitled( title ) {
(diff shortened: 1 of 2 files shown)


alisterscott:
Seemed to pass in CI - I'll run it again to see 👍

lib/pages/posts-page.js
```diff
@@ -8,22 +8,18 @@ export default class PostsPage extends BaseContainer {
...
+		return driverHelper.isElementPresent( this.driver, PostsPage._getPostLinkSelector( title ) );
 	}
 	editPostWithTitle( title ) {
-		const postTitleSelector = By.linkText( `${ title }` );
-		return driverHelper.clickWhenClickable( this.driver, postTitleSelector );
+		return driverHelper.clickWhenClickable( this.driver, PostsPage._getPostLinkSelector( title ) );
+	}
+
+	static _getPostLinkSelector( title ) {
+		return By.linkText( `${ title }` );
```
hoverduck: Do we need to be any more specific here than just link text to ensure that it's a link in the Posts list and not elsewhere on the screen?  I'm thinking if a bug leaves the post as a draft instead of publishing it, maybe that link text shows up in the "Continue editing..." prompt in the masterbar, or if we bring back the drafts list to the right of Posts.

alisterscott: Thanks, I think we'd need to use an xpath to select by class and by text, so I'm moving that into a separate enhancement here: https://github.com/Automattic/wp-e2e-tests/issues/862

PR #858, opened 2017-12-22
Rename G suite upsell test.

taggon:
The previous G Suite upsell has dropped because of some issues. Instead, the new version of the test `gsuiteUpsellTake2` will be launched soon.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -59,7 +59,7 @@
   "knownABTestKeys": [
     "businessPlanDescriptionAT",
     "domainsCheckoutLocalizedAddresses",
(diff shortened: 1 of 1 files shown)


fditrapani:
LGTM 👍

alisterscott:
Hi @taggon 

> The previous G Suite upsell has dropped because of some issues

Are you sure it's dropped? I am seeing this still in `master`:

screen shot 2017-12-27 at 4 04 45 pm

This is causing warnings for our e2e tests as they no longer know about this AB test (which this e2e PR removed)

PR #867, opened 2018-01-15
Adding the paymentMethodsOnPlans abtest

yoavf:
See https://github.com/Automattic/wp-calypso/pull/21442


alisterscott:
Thanks for this update - the error I will fix separately - not related.

PR #868, opened 2018-01-15
Add tests for 2fa login

Tug:
This PR adds 4 simple tests for 2fa login:
- email/password login with 2fa code via application (otp)
- email/password login with 2fa code via sms
- passwordless login with 2fa code via application (otp)
- passwordless login with 2fa code via sms

This required the addition of 2 new features:
- `lib/xmpp-client.js` which is used to listen to new sms for an account via xmpp.
- [`speakeasy`](https://github.com/speakeasyjs/speakeasy) dependency, which is a lib for OTP.

The code is a bit messy and could be refactored. I suggest we do this is a follow up PR.

### Testing Instructions
- `git checkout update/2fa-login`
- Create/Modify `config/local-test.json`  to add the new accounts information (see pb/1ce96). You can also use `local-development.json` if you have calypso running locally and `public-api.wordpress.com` sandboxed with the right filter to rewrite magic link urls).
- Run it for 1 device: `NODE_ENV=test BROWSERSIZE=mobile ./node_modules/.bin/mocha specs/wp-log-in-out-spec.js`
- Try multiple devices at once: `NODE_ENV=test ./node_modules/.bin/mocha specs/wp-log-in-out-spec.js`

### Reviews
- [x] Code
- [ ] Product


stephanethomas:
> I wasn't able to run the test suite on my virtual machine running Ubuntu

With some help from @Tug, I was able to fix that. The new tests run smoothly.

lib/pages/login-page.js
```diff
@@ -43,6 +43,40 @@ export default class LoginPage extends BaseContainer {
...
+		} else if ( twoFAMethod === 'backup' ) {
+			actionSelector = By.linkText( 'I can\'t access my phone' );
+		}
+
+		if ( actionSelector ) {
+			return driverHelper.isElementPresent( this.driver, actionSelector ).then( ( actionAvailable ) => {
+				if ( actionAvailable ) {
+					return driverHelper.clickWhenClickable( this.driver, actionSelector );
+				} else {
+					return Promise.resolve();
```
yurynix: can just `return;` here

lib/pages/login-page.js
```diff
@@ -43,6 +43,40 @@ export default class LoginPage extends BaseContainer {
...
+				}
+			} )
+		} else {
+			return Promise.resolve();
```
yurynix: can just `return;` here

lib/pages/login-page.js
yurynix: Now that I think about it, shouldn't we actually reject here?

lib/xmpp-client.js
```diff
@@ -0,0 +1,42 @@
...
+		if ( ! stanza.is( 'message' ) || stanza.getAttr( 'from' ) !== 'wpcom@im.wordpress.com' ) {
+			return;
+		}
+		const body = stanza.getChild( 'body' );
+		if ( ! body ) {
+			return;
+		}
+		const messageText = body.getText();
+		const smsContentRegex = /^SMS received from (\+?\d+): (.*)$/.exec( messageText );
+		if ( ! smsContentRegex || ! smsContentRegex[2] ) {
```
yurynix: should we check that `smsContentRegex.length > 2` here instead directly accessing 3rd index ?

specs/wp-log-in-out-spec.js
```diff
@@ -88,12 +91,76 @@ test.describe( `[${host}] Authentication: (${screenSize}) @parallel @jetpack`, f
...
+			test.it( 'Should be on the /log-in/sms page', function() {
+				return twoFALoginPage.displayed().then( function( displayed ) {
+					return driver.getCurrentUrl().then( ( urlDisplayed ) => {
+						assert( urlDisplayed.indexOf( '/log-in/sms' ) !== -1, 'The 2fa sms page is not displayed after log in' );
+					} );
+				} );
+			} );
+
+			test.it( 'Enter the 2fa code and we\'re logged in', function() {
+				return twoFALoginPage.enter2FACode( twoFACode );
```
yurynix: How do we assert that we're actually logged-in? From what I understand, and I'm fairly new to that code-base, this only current checks if the 2FA dialog disappears, but nothing more, should we maybe look for user's name or something like that to assert we're indeed correctly logged in?

lib/xmpp-client.js
Tug: 👍 I negated this condition (to return early) without thinking.

specs/wp-log-in-out-spec.js
Tug: we check that we can see the reader page, this is not part of the test yet i think

lib/xmpp-client.js
```diff
@@ -0,0 +1,42 @@
...
+	client.on( 'stanza', stanza => {
+		if ( ! stanza.is( 'message' ) || stanza.getAttr( 'from' ) !== 'wpcom@im.wordpress.com' ) {
+			return;
+		}
+		const body = stanza.getChild( 'body' );
+		if ( ! body ) {
+			return;
+		}
+		const messageText = body.getText();
+		const smsContentRegex = /^SMS received from (\+?\d+): (.*)$/.exec( messageText );
```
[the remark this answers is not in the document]
yurynix: Since we're controlling the webhook, why not send it in JSON format and just `JSON.parse()` it ?

specs/wp-log-in-out-spec.js
```diff
@@ -136,6 +203,140 @@ test.describe( `[${host}] Authentication: (${screenSize}) @parallel @jetpack`, f
...
+						return true;
+					} );
+				} );
+
+				test.describe( 'Can use the magic link and the code received via sms to log in', function() {
+					let magicLoginPage, twoFALoginPage, twoFACode;
+					test.before( function( done ) {
+						driver.get( magicLoginLink );
+						magicLoginPage = new MagicLoginPage( driver );
+						// make sure we listen for SMS before we trigger any
```
yurynix: > before we trigger any [what]

Looks like that comment misses something at it's end ?

specs/wp-log-in-out-spec.js
```diff
@@ -136,6 +203,140 @@ test.describe( `[${host}] Authentication: (${screenSize}) @parallel @jetpack`, f
...
+						const xmppClient = listenForSMS( loginFlow.account );
+						xmppClient.once( 'online', () => {
+							// send sms now!
+							magicLoginPage.finishLogin();
+							twoFALoginPage = new LoginPage( driver );
+							twoFALoginPage.use2FAMethod( 'sms' );
```
yurynix: Because we chose an account that has 2FA method set to sms,
the screen that will be displayed after successful magic link usage
will be the right one for the 2FA method...

So do we actually need to call it and if we do, do we need to tell it what method to use?

lib/pages/login-page.js
```diff
@@ -43,6 +43,40 @@ export default class LoginPage extends BaseContainer {
 		} );
 	}
 
+	use2FAMethod( twoFAMethod ) {
```
yurynix: On a "normal flow" test, do we even need to call that?
Since the user will be redirected to whatever method of 2FA she has, isn't she?
What am I missing here?

yurynix:
Thank you for working on that @Tug , that looks great ! ❤️ 
I think it's also quite doable to handle the push case in the same manner you did SMS, but lets keep it for another PR.

yurynix:
When running got that:
1516546641672-failed-en-mobile-enter-the-2fa-code-and-we-re-logged-in

screen shot 2018-01-21 at 16 58 20

yurynix:
Second run everything passed:
screen shot 2018-01-21 at 17 03 17

lib/pages/login-page.js
Tug: No, users might have all 2fa methods enabled and might see the push screen once and the sms screen another time, I created this method so we would be sure to be on the right page, and if we're not that we are offered the options to switch to the right page.

lib/pages/login-page.js
Tug: Yes we should 👍 thanks for noticing!

lib/xmpp-client.js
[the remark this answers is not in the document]
Tug: This channel isn't meant to be used for computer to computer communication (ie as an api), other types of data can flow through the wire so I thought a "human" format would make more senses here.

specs/wp-log-in-out-spec.js
Tug: We might want to pick an account that has multiple types of 2fa enabled, also it's not reliable which 2fa screen is displayed, if you had a failure once you will be displayed with a different screen on your next attempt (see my other comment)

lib/pages/login-page.js
Tug: Actually no, we should resolve here.
If the action is not available we're assuming that we're already on the right page.
The case where the option is not available is not tested right now!

specs/wp-log-in-out-spec.js
Tug: any sms?
I'll try to rephrase it as it doesn't seem very clear right now

yurynix:
Works as described, LGTM 🚢

lib/pages/login-page.js
```diff
@@ -42,6 +42,36 @@ export default class LoginPage extends BaseContainer {
...
+		let actionSelector;
+
+		if ( twoFAMethod === 'sms' ) {
+			actionSelector = By.linkText( 'Code via text message' );
```
hoverduck: These selectors will have a problem if/when we run these tests in other languages.  Can we add a data attribute in wp-calypso like [this](https://github.com/Automattic/wp-calypso/pull/21825) instead?

If that's not a viable option, we can leave this as-is for now.  But I'd recommend writing an issue in the wp-e2e-tests repo tagged with `workaround` just to document that these strings are here and a better solution should be found at some point.

Tug: 👍 I'll create an issue as soon as it is merged!

PR #871, opened 2018-01-21
Removing the paymentMethodsOnPlans abtest

yoavf:
Test has wrapped up  ~this should not be merged before https://github.com/Automattic/wp-calypso/pull/21713 is merged~


rachelmcr:
Thanks for this PR! Another change to the AB tests snuck in, though — could you rebase? And then this should be good to go.

PR #875, opened 2018-01-23
fix failing jetpack tests

brbrr:


specs/wp-post-editor-spec.js
```diff
@@ -27,6 +27,7 @@ const startBrowserTimeoutMS = config.get( 'startBrowserTimeoutMS' );
 const screenSize = driverManager.currentScreenSize();
 const host = dataHelper.getJetpackHost();
 const httpsHost = config.get( 'httpsHosts' ).indexOf( host ) !== -1;
+const siteName = dataHelper.getJetpackSiteName();
```
alisterscott: That needs to be conditional

specs/wp-post-editor-spec.js
```diff
@@ -834,7 +835,7 @@ test.describe( `[${host}] Editor: Posts (${screenSize})`, function() {
 					this.readerPage = new ReaderPage( driver, true );
 					this.navbarComponent = new NavbarComponent( driver );
 					this.navbarComponent.clickMySites();
-					this.sidebarComponent = new SidebarComponent( driver );
+					this.sidebarComponent = new SidebarComponent( driver, siteName );
```
alisterscott: Here too probably

PR #882, opened 2018-01-24
Add abtest setup for plans page

sirreal:
Ensure that AB test overrides are properly applied to plans pages.

Should prevent the following warnings and correctly apply the supplied AB test value:

```
Found an AB Testing key in local storage that isn't known: 'promoteYearlyJetpackPlanSavings'. This may cause inconsistent A/B test behaviour, please check this is okay and add it to 'knownABTestKeys' in default.config - […]
```

### DIFF
--- a/lib/pages/signup/pick-a-plan-page.js
+++ b/lib/pages/signup/pick-a-plan-page.js
@@ -19,6 +19,7 @@ export default class PickAPlanPage extends BaseContainer {
 			null,
 			config.get( 'explicitWaitMS' ) * 2
 		);
+		this.setABTestControlGroupsInLocalStorage();
 	}
 	_selectPlan( level ) {
 		let prefix = 'table';


lib/pages/signup/pick-a-plan-page.js
```diff
@@ -19,6 +19,7 @@ export default class PickAPlanPage extends BaseContainer {
 			null,
 			config.get( 'explicitWaitMS' ) * 2
 		);
+		this.setABTestControlGroupsInLocalStorage();
```
sirreal: This accepts a `url` param with no default. It's not clear to me whether that's required or not. I'm not sure what would make sense there.

https://github.com/Automattic/wp-e2e-tests/blob/98ce98063d9be8a6f7640e6a5251218a4d225b57/lib/base-container.js#L71-L82

rachelmcr:
@sirreal I'm not seeing that AB test warning anymore; is there a specific flow that's still triggering it? The Plans page in the signup flow should already have the AB tests in local storage, as it's applied earlier in the signup flow: https://github.com/Automattic/wp-e2e-tests/blob/98ce98063d9be8a6f7640e6a5251218a4d225b57/lib/pages/signup/start-page.js#L41

sirreal:
Thanks @rachelmcr, I think I was observing the warning from multiple sources and this is likely unneeded as you note. Closing.

PR #883, opened 2018-01-24
Remove site segment test

fditrapani:
We are removing the siteSegment test over here: https://github.com/Automattic/wp-calypso/pull/21797

This PR removes the siteSegment test from our e2e tests. 

@alisterscott and @taggon can you please review?

@alisterscott can you explain the order of operations once this PR has been approved? Do we merge and deploy the other PR in wp-calypso before this one? or do we deploy and merge this one first?

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -65,7 +65,6 @@
     "skipThemesSelectionModal",
     "unlimitedThemeNudge",
     "jetpackHidePlanIconsForAllDevices",
-    "signupSiteSegmentStep",
     "checklistThankYouForFreeUser",
     "checklistThankYouForPaidUser",
     "paymentMethodsOnPlans",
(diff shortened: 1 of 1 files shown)


markryall:
This looks like it should be ok.

I would have thought that these tests should pass since 100% of users are now allocated to the same variant being selected by these tests.

markryall:
Actually I think there are still some steps in signup tests (removing the design type) that will need to be changed to handle all the changes in https://github.com/Automattic/wp-calypso/pull/21797

alisterscott:
This will be good to merge after Automattic/wp-calypso#21797 - doesn't need to be immediately following it as nothing is critical - just clean up

fditrapani:
Thanks everyone!

PR #884, opened 2018-01-24
Fix Login flow

Tug:
My last PR changed the format for `this.account` in `LoginFlow` but did not update all the methods using it. This does.


seear:
Tested locally by using the latest full config and running:
```
./node_modules/.bin/mocha specs-jetpack-calypso/wp-jetpack-connect-spec.js
```
Result:
```
26 passing (4m)
```

Issue #411, opened 2017-03-09
Add missing e2e test around mapping an existing domain


alisterscott:
This is a missing e2e test

Could be during sign up or via an existing site

PR #885, opened 2018-01-25
Add e2e test: domain mapping via existing site

juliaamosova:
This PR adds e2e test for mapping an existing domain via an existing site ( to address this issue: https://github.com/Automattic/wp-e2e-tests/issues/411). 

**NEW FILES CREATED:**

New Spec created: 

* `wp-manage-domains-map-spec.js`

New Components created:

* `enter-a-domain-component.js`
* `find-a-domain-component.js`
* `map-a-domain-component.js`

New Pages created:

* `domain-map-page.js`
* `domain-map-checkout-page.js`
* `domain-my-own-page.js`

**TEST STEPS:**

The test consists of the following steps:

Map a domain to an existing site
      ✓ Log In and Select Domains 
      Can find domain mapping section and enter a domain to map
        ✓ Can choose add a domain 
        ✓ Can see the domain search component 
        ✓ Can select to use an existing domain 
        ✓ Can see use my own domain page 
        ✓ Can select to manually connect existing domain component 
        ✓ Can see enter a domain component
        ✓ Can enter the domain name 
        ✓ Can add domain to the cart 
        Can pay for domain mapping
          ✓ Can see checkout page 
          ✓ Can empty cart for next test 

**TEST GIFS:**

Desktop Test Gif: http://recordit.co/PcVkjWQ6D0

Tablet Test Gif: http://recordit.co/WcPq3wtixJ

Mobile Test Gif: http://recordit.co/FTO73BVJNF

### DIFF
--- a/lib/components/enter-a-domain-component.js
+++ b/lib/components/enter-a-domain-component.js
@@ -1,0 +1,21 @@
+import { By } from 'selenium-webdriver';
+
+import BaseContainer from '../base-container.js';
+import * as driverHelper from '../driver-helper';
+
+export default class EnterADomainComponent extends BaseContainer {
+	constructor( driver ) {
+		super( driver, By.css( '.map-domain-step__add-domain' ) );
+	}
+
+	enterADomain() {
+		const blogName = 'myawesomedomain.com';
+		const entryInputSelector = By.className( 'map-domain-step__external-domain' );
+		driverHelper.setWhenSettable( this.driver, entryInputSelector, blogName );
+	}
+
+	clickonAddButtonToAddDomainToTheCart() {
+		const addDomainSelector = By.css( '.map-domain-step__go.button.is-primary' );
+		return driverHelper.clickWhenClickable( this.driver, addDomainSelector, this.explicitWaitMS );
+	}
+}
--- a/lib/components/enter-a-domain-component.js
+++ b/lib/components/enter-a-domain-component.js
@@ -8,14 +8,13 @@ export default class EnterADomainComponent extends BaseContainer {
 		super( driver, By.css( '.map-domain-step__add-domain' ) );
 	}
 
-	enterADomain() {
-		const blogName = 'myawesomedomain.com';
+	enterADomain( blogName ) {
 		const entryInputSelector = By.className( 'map-domain-step__external-domain' );
 		driverHelper.setWhenSettable( this.driver, entryInputSelector, blogName );
 	}
 
 	clickonAddButtonToAddDomainToTheCart() {
 		const addDomainSelector = By.css( '.map-domain-step__go.button.is-primary' );
-		return driverHelper.clickWhenClickable( this.driver, addDomainSelector, this.explicitWaitMS );
+		return driverHelper.clickWhenClickable( this.driver, addDomainSelector );
 	}
 }
--- a/lib/components/find-a-domain-component.js
+++ b/lib/components/find-a-domain-component.js
@@ -35,7 +35,7 @@ export default class FindADomainComponent extends BaseContainer {
 	checkAndRetryForFreeBlogAddresses( expectedBlogAddresses, blogName ) {
 		const self = this;
 		self.freeBlogAddress().then( ( actualAddress ) => {
-			if (expectedBlogAddresses.indexOf( actualAddress ) < 0) {
+			if ( expectedBlogAddresses.indexOf( actualAddress ) < 0 ) {
 				let message = `The displayed free blog address: '${actualAddress}' was not in the expected addresses: '${expectedBlogAddresses}'. Re-searching for '${blogName}' now.`;
 				slackNotifier.warn( message );
 				self.searchForBlogNameAndWaitForResults( blogName );
--- a/lib/components/find-a-domain-component.js
+++ b/lib/components/find-a-domain-component.js
@@ -10,40 +10,46 @@ export default class FindADomainComponent extends BaseContainer {
 		super( driver, By.css( '.register-domain-step' ) );
 		this.declineGoogleAppsLinkSelector = By.className( 'google-apps-dialog__cancel-link' );
 	}
+
 	waitForResults() {
 		const driver = this.driver;
 		const resultsLoadingSelector = By.css( '.domain-suggestion.is-placeholder' );
 		driver.wait( function() {
 			return driverHelper.isElementPresent( driver, resultsLoadingSelector ).then( function( present ) {
-				return ! present;
+				return !present;
 			} );
 		}, this.explicitWaitMS * 2, 'The domain results loading element was still present when it should have disappeared by now.' );
 		return this.checkForUnknownABTestKeys();
 	}
+
 	waitForGoogleApps() {
 		return this.driver.wait( until.elementLocated( this.declineGoogleAppsLinkSelector ), this.explicitWaitMS, 'Could not locate the link to decline google apps.' );
 	}
+
 	searchForBlogNameAndWaitForResults( blogName ) {
 		const searchInputSelector = By.className( 'search__input' );
 		driverHelper.setWhenSettable( this.driver, searchInputSelector, blogName );
 		return this.waitForResults();
 	}
+
 	checkAndRetryForFreeBlogAddresses( expectedBlogAddresses, blogName ) {
 		const self = this;
 		self.freeBlogAddress().then( ( actualAddress ) => {
-			if ( expectedBlogAddresses.indexOf( actualAddress ) < 0 ) {
+			if (expectedBlogAddresses.indexOf( actualAddress ) < 0) {
 				let message = `The displayed free blog address: '${actualAddress}' was not in the expected addresses: '${expectedBlogAddresses}'. Re-searching for '${blogName}' now.`;
 				slackNotifier.warn( message );
 				self.searchForBlogNameAndWaitForResults( blogName );
 			}
 		} );
 	}
+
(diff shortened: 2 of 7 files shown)


lib/components/find-a-domain-component.js
```diff
@@ -10,64 +10,80 @@ export default class FindADomainComponent extends BaseContainer {
...
 	}
+
 	selectMapOwnDomain() {
 		const mapOwnSelector = By.css( 'div.domain-mapping-suggestion button' );
 		return driverHelper.clickWhenClickable( this.driver, mapOwnSelector, this.explicitWaitMS );
-	};
+	}
+
+	selectUseOwnDomain() {
+		const useOwnDomain = By.css( '.domain-suggestion.card.is-compact.is-clickable.is-visible.domain-transfer-suggestion' );
```
alisterscott: This works, but could be just `.domain-suggestion.card.domain-transfer-suggestion` which is less likely to change (I think) and still unique on the page:

screen shot 2018-01-25 at 4 08 36 pm

lib/components/find-a-domain-component.js
```diff
@@ -10,64 +10,80 @@ export default class FindADomainComponent extends BaseContainer {
...
 		this.driver.wait( until.elementIsEnabled( element ), this.explicitWaitMS, 'The paid address button for ' + dotComAddress + ' does not appear to be enabled to click' );
 		return driverHelper.clickWhenClickable( this.driver, selector, this.explicitWaitMS );
 	}
+
 	selectFreeAddress() {
 		const freeAddressSelector = By.css( '.domain-suggestion.is-clickable' );
 		return driverHelper.clickWhenClickable( this.driver, freeAddressSelector, this.explicitWaitMS );
 	}
+
 	selectMapOwnDomain() {
```
alisterscott: It looks like `selectMapOwnDomain()` is unused, so you could possibly update and use this, or remove it?

lib/components/find-a-domain-component.js
```diff
@@ -10,64 +10,80 @@ export default class FindADomainComponent extends BaseContainer {
...
 		return driverHelper.clickWhenClickable( this.driver, mapOwnSelector, this.explicitWaitMS );
-	};
+	}
+
+	selectUseOwnDomain() {
+		const useOwnDomain = By.css( '.domain-suggestion.card.is-compact.is-clickable.is-visible.domain-transfer-suggestion' );
+		return driverHelper.clickWhenClickable( this.driver, useOwnDomain, this.explicitWaitMS );
+	}
+
 	waitForOwnDomainMapping() {
```
alisterscott: `waitForOwnDomainMapping()` also seems unused - not related directly to this PR but it could be removed

lib/components/find-a-domain-component.js
```diff
@@ -10,64 +10,80 @@ export default class FindADomainComponent extends BaseContainer {
...
 		const searchInputSelector = By.className( 'search__input' );
 		driverHelper.setWhenSettable( this.driver, searchInputSelector, blogName );
 		return this.waitForResults();
 	}
+
 	checkAndRetryForFreeBlogAddresses( expectedBlogAddresses, blogName ) {
 		const self = this;
 		self.freeBlogAddress().then( ( actualAddress ) => {
-			if ( expectedBlogAddresses.indexOf( actualAddress ) < 0 ) {
+			if (expectedBlogAddresses.indexOf( actualAddress ) < 0) {
```
alisterscott: These extra spaces are actually required here to pass `eslint`, you can install an eslint plugin for your IDE which will give you real time eslint feedback, like I use WebStorm:

screen shot 2018-01-25 at 4 13 37 pm

lib/components/map-a-domain-component.js
```diff
@@ -0,0 +1,15 @@
...
+import BaseContainer from '../base-container.js';
+import * as driverHelper from '../driver-helper';
+
+export default class MapADomainComponent extends BaseContainer {
+	constructor( driver ) {
+		super( driver, By.css( '.transfer-domain-step' ) );
+	}
+
+	selectManuallyConnectExistingDomain() {
+		const manuallyConnectSelector = By.css( '#primary > main > span > div > div:nth-child(2) > div > div > p > a:nth-child(1)' );
```
alisterscott: We try to avoid CSS selectors like this because of the nested chain they're fragile and easy to break, since if someone changes one element in the chain it will no longer work.

I think something like `.card.transfer-domain-step__map-option a[href="#"]` will work, which finds the `a` (link) with a link of # which is a child of the card. I use Chrome to inspect elements and use command-f in the elements window where you can type a CSS selector to make sure it works (and is unique):

screen shot 2018-01-25 at 4 19 04 pm

specs/wp-manage-domains-map-spec.js
```diff
@@ -0,0 +1,128 @@
...
+				const mapADomainComponent = new MapADomainComponent( driver );
+				mapADomainComponent.selectManuallyConnectExistingDomain();
+			} );
+
+			test.it( 'Can see enter a domain component', () => {
+				const mapADomainPage = new MapADomainPage( driver );
+				mapADomainPage.displayed().then( ( displayed ) => {
+					assert.equal( displayed, true, 'Could not see map a domain page' );
+				} );
+
```
alisterscott: Just an eslint error on unnecessary new line

lib/components/enter-a-domain-component.js
```diff
@@ -0,0 +1,21 @@
...
+import BaseContainer from '../base-container.js';
+import * as driverHelper from '../driver-helper';
+
+export default class EnterADomainComponent extends BaseContainer {
+	constructor( driver ) {
+		super( driver, By.css( '.map-domain-step__add-domain' ) );
+	}
+
+	enterADomain() {
+		const blogName = 'myawesomedomain.com';
```
alisterscott: I think this const needs to be removed and added to the `enterADomain() `line as a variable like `enterADomain( blogName )`

lib/components/enter-a-domain-component.js
```diff
@@ -0,0 +1,21 @@
...
+		const entryInputSelector = By.className( 'map-domain-step__external-domain' );
+		driverHelper.setWhenSettable( this.driver, entryInputSelector, blogName );
+	}
+
+	clickonAddButtonToAddDomainToTheCart() {
+		const addDomainSelector = By.css( '.map-domain-step__go.button.is-primary' );
+		return driverHelper.clickWhenClickable( this.driver, addDomainSelector, this.explicitWaitMS );
```
alisterscott: It doesn't hurt, but you don't need to explicitly pass `this.explicitWaitMS` here since it has a default value (in earlier code samples it may not have had a default value, so there are examples like this throughout the code still)

lib/components/enter-a-domain-component.js
juliaamosova: Thanks! I should have known this myself. Fixed.

lib/components/enter-a-domain-component.js
juliaamosova: I was wondering about `this.explicitWaitMS` when writing this function but decided to leave it _as is_ thinking I might miss some cases where it is needed (that I am not aware of) if I remove it. Will ask clarifying questions next time. Fixed.

lib/components/find-a-domain-component.js
juliaamosova: Thanks for this! I also use WebStorm and do have `eslint` plugin installed. The highlight color is too pale in my environment which makes me miss some notifications. I'll look into it. Fixed the spacing here.

lib/components/find-a-domain-component.js
juliaamosova: I removed `selectMapOwnDomain()` here because I think it makes more sense. If I am not mistaken, prior to the introduction of Incoming Domain Transfers, the Domain Mapping section was on the same page where you would enter a new domain name if you were looking to register one. Now, there is no mention of the Domain Mapping immediately on the page where you would register a new domain name so I think it is best to keep all domain mapping related functions in a separate component dedicated to domain mapping specifically.

lib/components/find-a-domain-component.js
juliaamosova: Removed.

specs/wp-manage-domains-map-spec.js
juliaamosova: Thanks! Fixed.

lib/components/find-a-domain-component.js
juliaamosova: Thanks! It also looks more elegant. Fixed.

lib/components/map-a-domain-component.js
juliaamosova: Thanks for the explanation and the tip here! Also thanks for finding the right selector. Fixed.

alisterscott:
Hi @juliaamosova: you may need to rebase or merge changes from master into your branch, as there was a couple of updates for other tests which may fail on your PR.

juliaamosova:
> Hi @juliaamosova: you may need to rebase or merge changes from master into your branch, as there was a couple of updates for other tests which may fail on your PR.

@alisterscott did that just now and all checks have passed. Please review requested changes when get a chance. Thanks! 

cc @hoverduck

juliaamosova:
Thanks @alisterscott!

hoverduck:
Sorry I didn't get around to reviewing it while I was traveling, but happy to see it merged!  :+1:

juliaamosova:
Thanks @hoverduck!

Issue #862, opened 2018-01-03
Update post item selector to only be on the list of posts


alisterscott:
The method `static _getPostLinkSelector( title ) {` on the `lib/pages/posts-page.js` page currently just matches on the post title which can be anywhere in the DOM.

This should be more specific to be just on the posts list. 

It currently uses a `linkText` selector which doesn't support having a parent specified also, so we'd need to use an xpath selector I believe since css selectors can't select by text value.

Raised by @hoverduck in #857

PR #903, opened 2018-02-01
Update post item selector to only be on the list of posts #862

igladun:


alisterscott:
Fixes https://github.com/Automattic/wp-e2e-tests/issues/862

Issue #835, opened 2017-12-07
Show a warning when NODE_ENV is not set


alisterscott:

bsessions85:
Implemented in #907

PR #907, opened 2018-02-02
Showing a warning when NODE_ENV is not set.

bsessions85:
#835 Added a warning when NODE_ENV is not set. It will show a warning if you run using run.sh or using mocha to run a specific test file. 

@alisterscott  Can you check this out to make sure it is what you were thinking?


alisterscott:
The warning worked well for me - should we exit out with a non-zero exit code at that point?

bsessions85:
Yeah, I think we should. It is much clearer what the problem is. I updated to exit right away

Issue #787, opened 2017-11-07
Add tests for Customizer updates in WP 4.9


hoverduck:
WordPress 4.9 brings some updates to the customizer that it would be nice to cover in our flows.  For starters there's the ability to save a draft of your changes and preview them via custom URL.

There have also been some changes to the nav menus - https://make.wordpress.org/core/2017/11/07/nav-menu-improvements-in-the-customizer-in-4-9/

Others?

brandonpayton:
Here is another change that added publishing options for Customizer changesets:
https://make.wordpress.org/core/2017/11/03/new-features-and-enhancements-with-customizer-changesets-in-4-9/

A Settings button was added to the Publish button and opens a view that allows you to save a changeset as a draft or schedule it to be published in the future.

alisterscott:
Our forward focus for the e2e tests is on Calypso so we won't duplicate testing core WordPress customizer functionality like saving drafts of changes within our e2e tests - closing.

PR #908, opened 2018-02-05
E2E test for case "Save a draft of menu changes and preview them via …

igladun:
…custom URL"#787

### DIFF
--- a/lib/base-container.js
+++ b/lib/base-container.js
@@ -40,8 +40,9 @@ export default class BaseContainer {
 	}
 
 	waitRequestsToComplete() {
-		//method which will implement waiting while all AJAX requests are completed
-		return this.driver.sleep( 5000 );
+		return this.driver.wait( () => {
+			return this.driver.executeScript( 'return jQuery.active === 0' );
+		}, 10000, 'Timed out waiting while AJAX loads' );
 	}
 
 	displayed() {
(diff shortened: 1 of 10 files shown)


specs/wp-post-nux-spec.js
```diff
@@ -66,18 +66,6 @@ test.describe( `[${host}] Post-NUX Flows (${screenSize}) @parallel`, function()
 					} );
 				} );
 
-				test.it( 'Can update and view site tagline', function() {
```
hoverduck: Is there a reason you took this out?

lib/base-container.js
```diff
@@ -39,6 +39,12 @@ export default class BaseContainer {
 		return driverHelper.waitTillPresentAndDisplayed( this.driver, this.expectedElementSelector, this.explicitWaitMS );
 	}
 
+	waitRequestsToComplete() {
```
alisterscott: this will only work on the customizer itself which is the only place where jQuery will be available, so it would be best to move this into the `lib/pages/customizer-page.js` to make sure it's not called from other pages.

lib/base-container.js
```diff
@@ -39,6 +39,12 @@ export default class BaseContainer {
...
+		return this.driver.wait( () => {
+			return this.driver.executeScript( 'return jQuery.active === 0' );
```
alisterscott: This method should also check to make sure that jQuery is loaded also because sometimes switching iFrames there are delays (which is why we have the sleeps to begin with)

lib/base-container.js
```diff
@@ -39,6 +39,12 @@ export default class BaseContainer {
...
+		}, 10000, 'Timed out waiting while AJAX loads' );
```
alisterscott: Where did this number come from? How did you decide upon it?
It should be declared at least as a constant and preferably as a config value. You could probably use `this.explicitWaitMs` which I believe has the same intention

igladun:
Thanks for the review. I fixed all remarks. The test was failing sometimes on mobile because the theme of test account is changing. Now should be OK.
@hoverduck I moved the test to separate file, because I think we will have more tests for the Customizer and to don't load the  `wp-post-nux`.

alisterscott:
Thanks for the updates @igladun 

I think this looks good now - my only concern is whether these menus will build up over time - what do you think @hoverduck?

hoverduck:
> my only concern is whether these menus will build up over time 

Yeah, now that you mention it I worry about that too.  I wasn't paying attention to the fact that that's a persistent change beyond the scope of the Customizer.  We could update the cleanup cron to also wipe out the menu, but it's probably better to have this test just modify a setting like the site title, or add a widget or something that only changes the front-end.

igladun:
After discussion with @hoverduck , closing this PR and creating a new one

Issue #906, opened 2018-02-02
Use a meta selector to find 2fa links on the 2fa login page


Tug:
See this discussion
https://github.com/Automattic/wp-e2e-tests/pull/868#discussion_r165448322

We should add different css classes to the various types of links on the 2fa page and get those links using a query selector for instance.

PR #909, opened 2018-02-05
Select 2fa options using the new data-e2e-link attributes

Tug:
Addresses https://github.com/Automattic/wp-e2e-tests/issues/906

This PR updates our test to use the new `data-e2e-link` attributes added to the login 2fa screens in https://github.com/Automattic/wp-calypso/pull/22088.
It also fixes a race condition when clearing the cookies before each login tests.

### Testing Instructions
- Check that the login suite passes
`NODE_ENV=test BROWSERSIZE=mobile ./node_modules/.bin/mocha specs/wp-log-in-out-spec.js`

### Reviews
- [x] Code
- [ ] Product

### DIFF
--- a/lib/pages/login-page.js
+++ b/lib/pages/login-page.js
@@ -46,11 +46,11 @@ export default class LoginPage extends BaseContainer {
 		let actionSelector;
 
 		if ( twoFAMethod === 'sms' ) {
-			actionSelector = By.linkText( 'Code via text message' );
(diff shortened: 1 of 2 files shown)


stephanethomas:
Changes looks good to me. I'm unable to launch the test suite at the moment because of limited Internet connectivity, but otherwise this should be good to go.

PR #912, opened 2018-02-06
Login: Add test for push 2fa case

yurynix:
Add a case for a user logging in with a push 2FA enabled.

This test case imitates the Android [app](https://play.google.com/store/apps/details?id=org.wordpress.android) getting a push with a token via [GCM](https://developers.google.com/cloud-messaging/) and then "approving" it via
WPCOM rest API. It then asserts the user is indeed logged in.

#### Testing instructions
  
1. Run `git checkout add/login-2fa-push`
2 Create/Modify `config/local-test.json`  to add the new accounts information (see pb/1d081). You can also use `local-development.json`.
- Run it for 1 device: `NODE_ENV=test BROWSERSIZE=mobile ./node_modules/.bin/mocha specs/wp-log-in-out-spec.js`

#### Reviews
  
- [ ] Code
- [ ] Product


lib/push-client.js
```diff
@@ -0,0 +1,79 @@
...
+ * Listens to push notifications with push token
+ * @param {Object} pushConfig GCM keys and tokens
+ * @param {Function} callback a function to call when we get a push token
+ */
+export const subscribeToPush = ( pushConfig, callback ) => {
+	const persistentIds = [];
+	let connection;
+	listen(
+		{
+			...pushConfig,
```
Tug: Why not `listen( pushConfig, ... )`?

Tug:
I think there might be a problem with the `push-receiver` dependency, it's not getting installed with `npm install`. Maybe switch to `git+https` ? That would look like

Other than that I couldn't get the test to pass,
I edited my `local-test.json` config and ran `env NODE_ENV=test BROWSERSIZE=mobile ./node_modules/.bin/mocha specs/wp-log-in-out-spec.js` but in the end I got:
```
  16 passing (8m)
  2 failing

  1) [WPCOM] Authentication: (mobile) @parallel @jetpack @visdiff Can Log in on a 2fa account "before all" hook: ret:
     Error: Timeout of 160000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.


  2) [WPCOM] Authentication: (mobile) @parallel @jetpack @visdiff Can Log in on a passwordless account with 2fa using sms Can request a magic link email by entering the email of an account which does not have a password defined Can use the magic link and the code received via sms to log in "before all" hook: ret for "Can find the magic link in the email received":
     Error: Timeout of 160000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

yurynix:
What fails is not the push test, but another test, which is kinda weird.

It's stuck on this screen:
screen shot 2018-02-08 at 9 19 20

screen shot 2018-02-08 at 9 22 21
( highlighted in green is the test added by this PR )

Tug:
LGTM 👍 

Could you fix the eslint issues mentioned in CircleCI, then it should be good to go!

Tug:
🚢

PR #893, opened 2018-01-30
Add linting pre-commit hook

brbrr:
This will prevent maintainers to push code with linting errors. In addition - pre-commit hook will lint only changed files.

PR #918, opened 2018-02-08
Add linting pre-commit hook take 2

brbrr:
That's 2d attempt of #893. In this PR I use different npm package ([husky](https://github.com/typicode/husky)) which did not create issues when switching between branched which are outdated from master

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ screenshots-i18n/
 spec-xunit-reporter-0.0.1.tgz
 *xunit_*.xml
 .DS_Store
+npm-debug.log
 
 # Ignore all local config files, but keep the example there.
 config/local-*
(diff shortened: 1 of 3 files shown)


alisterscott:
I added and committed a file with eslint errors without any warnings or errors -> not sure what I am doing wrong.

Also, running `npm install` for this branch generated changed files `package.json` and `package-lock.json`: 

screen shot 2018-02-09 at 3 19 02 pm

hoverduck:
@alisterscott This wasn't working for you or me because we had previously manually set a pre-commit hook linked to `scripts/git-pre-commit-circleci-validate` for validating the CircleCI config file.  @rachelmcr you're probably the only other one who might have that configured.  You'll need to manually remove that link and then re-run `npm install` (probably after blowing away your node_modules dir) to get husky to work.

I've updated this PR to also run that CircleCI validate script, so it's win/win.  This is a much better approach than what I'd had previously, since it doesn't require manual configuration.

This Slack mention was the only place I could find where I "announced" the feature, so that's why it was never configured for any of the newer contributors: p1506535003000456-slack-e2e-testing-discuss

brbrr:
I'm going to merge this PR since CI errors not actually related to these changes

PR #919, opened 2018-02-08
Use encrypted config to store credentials

### LINKED ISSUE #467: A secure way to store common credentials
Encrypt in git? Or another way? How does this work with CircleCI

brbrr:
Fixes #467 

This PR adds encrypted config file which is checked in to git. That config file is a compilation of all target specific config files. How will we handle config values which are different across targets?

- Config file has its "root" values which are same for all targets e.g. `testSiteForInvites`
- Config file includes `TARGET` values, which basically overrides from specific targets. e.g. `JETPACK` target has its own `slackHook` 
- `TARGET` should be defined as env variable in circleCI config file for all targets except `WP_E2E_TESTS` & `WP_E2E_TESTS_FOR_BRANCHES`

Documentation is pending

- [ ] Move (or add) `NODE_ENV` env vars into CircleCI config files
- [x] Review all the configs, and compile single file including all the differences and separate overriding values like `slackHook` etc
- [ ] Add `TARGET` env var into CircleCi config to be able to differentiate similar targets from each other(mostly apply for targets with `$JETPACKHOST=WPCOM`)
- [ ] Add internal documentation


alisterscott:
This looks good, it'll be useful to have just one config easily accessible.
One thing that might need addressing is the values for Slack notifications `slackChannelForScreenshots` and `slackHook` are currently different across projects/configs and with this change I believe they'll all go into the main notifications channel.
Also, I believe `i18nScreenshotUser` is missing from the encryped file - I'm not sure how many other of these there are when consolidating all the files

brbrr:
I'll just put it here(we have **9** test targets!!):

```
WP_E2E_TESTS
https://circleci.com/gh/Automattic/wp-e2e-tests
https://github.com/Automattic/wp-e2e-tests
NODE_ENV=test

WP_E2E_TESTS_CANARY
https://circleci.com/gh/Automattic/wp-e2e-tests-canary
https://github.com/Automattic/wp-e2e-tests-canary
NODE_ENV=test

WP_E2E_TESTS_JETPACK
https://circleci.com/gh/Automattic/wp-e2e-tests-jetpack
https://github.com/Automattic/wp-e2e-tests-jetpack
NODE_ENV=test
JETPACKHOST=PRESSABLE

WP_E2E_TESTS_FOR_BRANCHES
https://circleci.com/gh/Automattic/wp-e2e-tests-for-branches
https://github.com/Automattic/wp-e2e-tests-for-branches
NODE_ENV=test

JETPACK
https://circleci.com/gh/Automattic/jetpack
https://github.com/Automattic/jetpack
NODE_ENV=test

WP_E2E_TESTS_VISDIFF
https://circleci.com/gh/Automattic/wp-e2e-tests-visdiff
https://github.com/Automattic/wp-e2e-tests-visdiff
NODE_ENV NOT-DEFINED

WP_E2E_TESTS_IE11
https://circleci.com/gh/Automattic/wp-e2e-tests-ie11
https://github.com/Automattic/wp-e2e-tests-ie11
NODE_ENV=test

WP_E2E_TESTS_WOO
https://circleci.com/gh/Automattic/wp-e2e-tests-woocommerce
https://github.com/Automattic/wp-e2e-tests-woocommerce
NODE_ENV=test

WP_E2E_TESTS_i18N
https://circleci.com/gh/Automattic/wp-e2e-tests-i18n
https://github.com/Automattic/wp-e2e-tests-i18n
NODE_ENV=test
```

alisterscott:
> WP_E2E_TESTS_IE11
> https://circleci.com/gh/Automattic/wp-e2e-tests-ie11
> https://github.com/Automattic/wp-e2e-tests-ie11
> NODE_ENV=test

This isn't used and can be removed

alisterscott:
What's the difference between JETPACK and WP_E2E_TESTS_JETPACK

brbrr:
They are using different hosting - WP_E2E_TESTS_JETPACK uses PRESSABLE, while JETPACK - self-hosted sites created via ServerPilot(and that target will switch to Jurassic.ninja)

.circleci/config.yml
```diff
@@ -7,12 +7,14 @@ jobs:
...
         environment:
                 NODE_ENV: test
+                NODE_CONFIG: {}
     steps:
       - checkout
       - restore_cache:
           key: << checksum "package.json" >>
       - run: source $HOME/.nvm/nvm.sh && npm install
       - run: source $HOME/.nvm/nvm.sh && npm run lint
+      - run: source $HOME/.nvm/nvm.sh && openssl enc -aes-256-cbc -d -k $CONFIG_KEY -in ./config/encrypted.enc -out ./config/local-test.json
```
hoverduck: I think a better way would be `openssl enc -aes-256-cbc -d -pass env:CONFIG_KEY -in ./config/encrypted.enc -out ./config/local-test.json`, since that won't risk accidentally dumping the config key to the screen (and CircleCI UI) if the command fails.

hoverduck:
Another option for managing the project-unique values like Slack hooks is to set the `NODE_APP_INSTANCE` variable in CircleCI and have separate config files for each project with just the overrides - https://github.com/lorenwest/node-config/wiki/Configuration-Files#file-load-order

brbrr:
> have separate config files for each project

Yeah, that would be easier solution, but still, we would need to maintain multiple config files, which isn't great at all, and I think we should try to have as much config files as possible

alisterscott:
I have updated

wp-e2e-tests-for-branches

and

wp-e2e-tests-woocommerce

to both use new config and all working great ✅

PR #924, opened 2018-02-12
Remove: G Suite Upsell V2 test from the configuration.

taggon:
As the test ends(see https://github.com/Automattic/wp-calypso/pull/21787), it should be removed from the e2e tests default configuration too.


dzver:
LGTM

hoverduck:
Thanks for the update @taggon, I've finished out the necessary changes and will merge it now! :tada:

PR #925, opened 2018-02-12
New AB test and fixed domain cancel flow

hoverduck:


hoverduck:
Customizer tests failed on CI, but passed locally.  I'm merging this and will troubleshoot if they continue to fail on CI

PR #928, opened 2018-02-13
Test for previewing title changes through preview URL

igladun:
Simplified version of changes preview test

### DIFF
--- a/lib/base-container.js
+++ b/lib/base-container.js
@@ -96,6 +96,6 @@ export default class BaseContainer {
 	}
 
 	navigateBack() {
-		this.driver.navigate().back();
+		return this.driver.navigate().back();
 	}
 }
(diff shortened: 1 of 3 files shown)


igladun:
@hoverduck @alisterscott, sorry, forget to set you as reviewers.

lib/base-container.js
```diff
@@ -94,4 +94,8 @@ export default class BaseContainer {
 
 		return this.waitForPage();
 	}
+
+	navigateBack() {
```
hoverduck: I think this would make more sense to go in `driver-helper` rather than `base-container`

alisterscott:
I've had to revert this in #942 because for some reason it doesn't work when running against calypso.localhost:3000, the back button remains on the front end of the site and the customizer step then fails.

@igladun: would you mind having a look at why this is the case on calypso.localhost:3000?

igladun:
@alisterscott , I will check

PR #949, opened 2018-02-17
Test for preview title changes

igladun:
Hi @hoverduck  and @alisterscott . This issue with driver.back() inside calypso is tricky and I am investigating it. I re-wrote the test using open the preview in a new window.

### DIFF
--- a/specs/wp-post-nux-spec.js
+++ b/specs/wp-post-nux-spec.js
@@ -10,6 +10,7 @@ import * as mediaHelper from '../lib/media-helper';
 import LoginFlow from '../lib/flows/login-flow';
 import CustomizerPage from '../lib/pages/customizer-page';
 import SidebarComponent from '../lib/components/sidebar-component';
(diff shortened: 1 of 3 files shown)


specs/wp-post-nux-spec.js
```diff
@@ -59,13 +61,50 @@ test.describe( `[${host}] Post-NUX Flows (${screenSize}) @parallel`, function()
...
+
+				test.it( 'Can see the correct title displayed', function() {
+					switchToWindow( driver, 1 );
+					return this.customizerPage.title().then( ( title ) => {
+						return assert( title.includes( newSiteTitle ), 'The title is correctly displayed on the preview' );
+					} );
+				} );
+
+				test.it( 'Can revert the site title', function() {
+					switchToWindow( driver, 0 );
```
alisterscott: This leaves the new window open which isn't ideal, especially if you had a subsequent test that opens a new window

PR #953, opened 2018-02-20
Set the domain suggestion group

alisterscott:
This is causing inconsistent behaviour

Uses group_0

Introduced here: https://github.com/Automattic/wp-calypso/pull/22463/


alisterscott:
I'm going to merge this one since it's causing inconsistent test results on master

PR #954, opened 2018-02-20
Override AB test groups when logged in

### LINKED ISSUE #929: AB tests not overriden for logged in users
Lots of warnings in Slack

Seems to be only logged in users

Possibly related to https://href.li/?https://github.com/Automattic/wp-calypso/pull/22149

alisterscott:
Fixes #929

### DIFF
--- a/lib/flows/login-flow.js
+++ b/lib/flows/login-flow.js
@@ -77,9 +77,9 @@ export default class LoginFlow {
 
 		this.editorPage = new EditorPage( this.driver );
 
-		// this.driver.getCurrentUrl().then( ( urlDisplayed ) => {
-		// 	return this.editorPage.setABTestControlGroupsInLocalStorage( urlDisplayed );
-		// } );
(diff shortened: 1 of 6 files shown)


hoverduck:
@alisterscott The change looks good, but for some reason the Trash a Post spec is failing on this branch...it's dropping the user back on the Posts list rather than the Reader (which was the old behavior a while back).  I don't see any AB tests that look like they'd be causing that, but I must be missing something...

alisterscott:
@hoverduck thanks, I know what that is. If we don't override the groups the it remembers the screen before the editor so it returns to there, however, when we override the groups we refresh the page so it loses the return path, so it returns us to the Reader (root) - I'll update the spec

alisterscott:
I'm going to merge this one as we're seeing inconsistent behaviour for the logged in domain test

failed-en-desktop-can-select-the--com-search-result-and-decline-google-apps-for-email-1519169529215

PR #962, opened 2018-02-21
Try chromedriver 2.35.0

alisterscott:
Try upgrading Chromedriver


alisterscott:
> session not created exception: Chrome version must be >= 62.0.3202.0

PR #967, opened 2018-02-22
Add/jetpack pressable nux flow

### LINKED ISSUE #932: Jetpack: Add e2e test for Pressable NUX flow
Add a test for the Pressable NUX flow, to catch regressions like this: https://github.com/Automattic/wp-calypso/issues/22378

Related (map of this flow): p6jOYE-1wo-p2

brbrr:
Fixes #932


rachelmcr:
One error I'm seeing is on the Activity Log page, for the `Can add site credentials` step:

```
Timed out waiting for element with css selector of '.activity-log-day' to be present and displayed
```

I noticed that's happening on the Activity Log page in Calypso (looks like it's happening because we don't necessarily have recorded activity yet at this stage, so that class doesn't show up):

failed-en-desktop-can-add-site-credentials-1519325520606

However, that shouldn't be the point in the flow where users add their credentials. During the Pressable NUX flow, after approving the connection on the authorization page users should land on `/start/pressable-nux/creds-permission` with a prompt to share credentials:

screenshot 2018-02-22 19 04 07

brbrr:
Tests are failing because of the expected bug: https://github.com/Automattic/wp-calypso/issues/22378

This PR is ready to review, but not ready to merge

specs-jetpack-calypso/wp-pressable-nux-spec.js
```diff
@@ -0,0 +1,92 @@
...
+
+		test.it( 'Can approve login with WordPress', function() {
+			this.pressableApprovePage = new PressableApprovePage( driver );
+			return this.pressableApprovePage.approve();
+		} );
+
+		test.it( 'Can create new site', function() {
+			this.siteName = dataHelper.getNewBlogName();
+			this.pressableSitesPage = new PressableSitesPage( driver );
+			return this.pressableSitesPage.addNewSite( this.siteName );
```
rachelmcr: My test run failed at this point because the Pressable test account already has the maximum allowed sites (5 sites), so at least one site will need to be deleted to create a new one. A couple ideas:

- We could check for that before adding a new site, and delete one if needed.
- We could clean up at the end of the flow and delete the new site we created.

rachelmcr: I see now that you do clean up at the end of the flow, so I guess we also need a check just in case that doesn't work as expected and the account gets filled up.

specs-jetpack-calypso/wp-pressable-nux-spec.js
```diff
@@ -0,0 +1,92 @@
...
+			this.jetpackAuthorizePage = new JetpackAuthorizePage( driver );
+			return this.jetpackAuthorizePage.approveConnection();
+		} );
+
+		test.it( 'Can proceed with Pressable NUX flow', function() {
+			this.nuxFlow = new PressableNUXFlow( driver );
+			return this.nuxFlow.addSiteCredentials();
+		} );
+
+		test.it( 'Can navigate to the Jetpack dashboard', function() {
```
rachelmcr: Could we change this to something more specific to the JITM we're testing for here? I was confused at first when this step failed, as I saw the Jetpack dashboard load but the step had failed. Maybe `Can see expected content in the Jetpack dashboard` or something like that?

specs-jetpack-calypso/wp-pressable-nux-spec.js
```diff
@@ -0,0 +1,92 @@
...
+		} );
+
+		test.it( 'Can approve connection on the authorization page', function() {
+			this.jetpackAuthorizePage = new JetpackAuthorizePage( driver );
+			return this.jetpackAuthorizePage.approveConnection();
+		} );
+
+		test.it( 'Can proceed with Pressable NUX flow', function() {
+			this.nuxFlow = new PressableNUXFlow( driver );
+			return this.nuxFlow.addSiteCredentials();
```
rachelmcr: Should we also wait and check for the success message after adding our site credentials? That way we know if adding the site credentials failed or not, and we can fail at this step if the site credentials aren't successfully added. (And in the next step we're only checking if the Rewind JITM shows up errantly after successfully adding the site credentials, rather than using that as a check for whether the site credentials were added.)

lib/pages/pressable/pressable-sites-page.js
```diff
@@ -1,8 +1,50 @@
...
+
+const explicitWaitMS = config.get( 'explicitWaitMS' );
 
 export default class PressableSitesPage extends BaseContainer {
-	constructor( driver ) {
-		super( driver, by.css( 'form[action="/sites"]' ) );
+	constructor( driver, visit = false ) {
+		const url = 'https://my.pressable.com/sites';
+
+		super( driver, By.css( 'form[action="/sites"]' ), visit, url );
```
rachelmcr: This form only appears when the account hasn't reached its maximum number of sites. If the account has reached its maximum sites, and you try to use this page (e.g. to delete a site), it fails on the `waitForPage()` method:

`TimeoutError: Timed out waiting for element with css selector of 'form[action="/sites"]' to be present and displayed`

It looks like the class `site-show-sections` may be specific to this page and could be used here instead.

PR #968, opened 2018-02-22
add TARGET env var to run-wrapper

brbrr:
Depending on the branch name our PRs can trigger Jetpack / Woo specs. I've added `TARGET` env vars accordingly to make them comply with our new config structure

### DIFF
--- a/scripts/run-wrapper.sh
+++ b/scripts/run-wrapper.sh
@@ -31,8 +31,10 @@ if [ "$RUN_SPECIFIED" == "true" ]; then
   TESTARGS=$RUN_ARGS
 elif [[ "$CIRCLE_BRANCH" =~ .*[Jj]etpack.*|.*[Jj][Pp].* ]]; then
   export JETPACKHOST=PRESSABLE
+	export TARGET=JETPACK
   TESTARGS="-R -j" # Execute Jetpack tests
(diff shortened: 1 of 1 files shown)


scripts/run-wrapper.sh
```diff
@@ -31,8 +31,10 @@ if [ "$RUN_SPECIFIED" == "true" ]; then
   TESTARGS=$RUN_ARGS
 elif [[ "$CIRCLE_BRANCH" =~ .*[Jj]etpack.*|.*[Jj][Pp].* ]]; then
   export JETPACKHOST=PRESSABLE
+	export TARGET=JETPACK
```
alisterscott: this is a tab when expecting spaces

scripts/run-wrapper.sh
```diff
@@ -31,8 +31,10 @@ if [ "$RUN_SPECIFIED" == "true" ]; then
...
   TESTARGS="-R -j" # Execute Jetpack tests
 elif [[ "$CIRCLE_BRANCH" =~ .*[Ww][Oo][Oo].* ]]; then
+	export TARGET=WOO
```
alisterscott: same

scripts/run-wrapper.sh
brbrr: Thanks for noting this! VSCode just makes it looks nice whenever it space or tab indented. Will try to solve this.

PR #969, opened 2018-02-22
Add indentation rules for .sh and .yaml files

brbrr:


alisterscott:
Works well - thanks for the fix 👍 

Before

before

After

after

PR #970, opened 2018-02-22
Force the correct version of Chromedriver + Chrome

hoverduck:
PR #965 forced WebDriver to load the correct version of Chromedriver installed via npm, but this version is incompatible with the version of Chrome running on CircleCI 1.0 wrapper-repos.

Until we determine the long term solution (likely getting rid of wrapper repos altogether), this PR will force those CI 1.0 containers to download the current stable version of Chrome.

### DIFF
--- a/lib/driver-manager.js
+++ b/lib/driver-manager.js
@@ -121,6 +121,10 @@ export function startBrowser( { useCustomUA = true, resizeBrowserWindow = true }
 					options.addArguments( '--headless' );
 					global.isHeadless = true;
 				}
+
+				const service = new chrome.ServiceBuilder( chromedriver.path ).build();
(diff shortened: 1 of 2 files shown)


alisterscott:
Merging - let's see how how deep this rabbit hole goes

alisterscott:
>  It would probably be good to have an issue to remind us to remove it after we've got everything switched over.

Raised as #973

alisterscott:
An issue on mobile screens for payment buttons
Pending fix here: https://github.com/Automattic/wp-e2e-tests/pull/972

alisterscott:
Removed in #1175 as we're no longer on CircleCI 1.0

PR #971, opened 2018-02-22
Ignore failed clicks on IE

hoverduck:
The signup tests were failing on IE with inconsistent behavior.  I tracked this down (mostly) to what seems to be a timing race on clicking the Continue button at the end of the signup flow, where the IE driver isn't giving correct status back that it completed the click.  Sometimes the click is successful, but the operation still fails and returns false.

This PR adds a check to make it always assume success on IE.


alisterscott:
Looks good, but are we running these tests anywhere at the moment? Will Edge be more reliable (and still find bugs)?

hoverduck:
Bah, I rebased this branch from master and ended up with a cluttered commit history, so this is showing a bunch of unnecessary changes.

alisterscott:
> Bah, I rebased this branch from master and ended up with a cluttered commit history, so this is showing a bunch of unnecessary changes.

Fixed it

Issue #978, opened 2018-02-26
Workaround for Jetpack Disconnect


alisterscott:
The confirmation dialog is no longer appearing. Raised here:

https://github.com/Automattic/wp-calypso/issues/22794

![disconnect site](https://user-images.githubusercontent.com/128826/36648782-77eb10e0-1ae3-11e8-94d8-46e3d9f67b98.gif)

alisterscott:
Fixed, not needed

PR #979, opened 2018-02-26
Update the jetpack disconnect spec

alisterscott:
Put a workaround in place for https://github.com/Automattic/wp-e2e-tests/issues/978


alisterscott:
@brbrr Do you mind having a look at this? Works for me locally but fails on CircleCI. There's an issue with the disconnect notice not appearing.

brbrr:
Can we close this as not needed?

PR #981, opened 2018-02-26
Update npm dependencies for github repos

alisterscott:
The newer version of npm fails otherwise

Testing:
1. Remove node_modules
2. npm install

Make sure it works


alisterscott:
Still failing after deleting node_modules

alisterscott:
@hoverduck I can't get to work on a clean install - any ideas?

You're planning on moving these dependencies anyway?

hoverduck:
This was double-trouble:

* We started using Node v8.9.4 which includes npm v5.6.0, which made minor changes to the way `npm pack` works, which we're using to generate the custom reporter package in a `postinstall` step.  Since that was run after all the other packages were installed, we ran into (I think) this bug:
* https://github.com/npm/npm/issues/17379 - This documents issues with certain packages pulling from GitHub specifically, although even when I removed that bit from our package in testing I still hit problems.

But I discovered that I can skip the `npm pack` step and just install the reporter directly in the dependencies list as

```
    "spec-xunit-reporter": "./lib/reporter",
```

I'm not sure if that's a new capability or if we just didn't know it was possible that way before.  /shrug

`npm ls` still throws a lot of warnings about missing dependencies, but as best I can tell everything is working alright.  There aren't any more "invalid package" errors at least.

alisterscott:
Thanks for your investigation and effort @hoverduck 

I tested this on a fresh checkout in a new directory and it works well

It looks like that npm bug is fixed, but I think it's better to just merge this than try to force an update to npm everywhere

alisterscott:
It looks like this kills our wrapper repos: 

```cd wp-e2e-tests && npm install returned exit code 254
Action failed: cd wp-e2e-tests && npm install```

I'm trying re-running the wrapper repos without cache to see if this fixes things

alisterscott:
Forcing a cache refresh on the wrapper repos fixes the issue - I'll try to do as many as I can find

Issue #754, opened 2017-10-25
Investigate timeouts with chrome --headless option


hoverduck:
I tried updating the tests running on the Jetpack repo were updated with https://github.com/Automattic/jetpack/pull/8049 to use the `--headless` option after we merged #744 on wp-e2e-tests, but for some reason the tests were completely imploding with random failures - https://circleci.com/gh/Automattic/jetpack/1077

I switched back to forcing `xvfb-run` instead of the `-x` option on run.sh, and the tests ran cleanly.  We need to figure out what's different about the Jetpack test environment that's causing such havoc.

hoverduck:
Seems like this was affecting the wpcom tests as well, not just Jetpack.  Although it only started happening here in the last hour or two...so it could still be something unrelated.

alisterscott:
We haven't actually seen this for some time now - closing

PR #982, opened 2018-02-26
Use generic CircleCI container

hoverduck:
Since we're controlling the Chromedriver version via npm now the only real reason left that we're using a custom Docker container is the Chrome version, which doesn't really buy us that much.

This PR keeps us on CircleCI 2.0, but uses one of their provided container images (which currently has Chrome v64, for the record).  This should greatly simplify changes down the line since we won't need to build/push a new image just to change versions.  And from what I've seen so far it looks like CircleCI will likely continue updating Chrome on newer images as new versions of Node come out, so that should keep us current.

I want to run this through a few times before merging it to watch out for timeouts (https://github.com/Automattic/wp-e2e-tests/issues/754).  And we'll need to merge a similar change into the Jetpack repo as well, as it expects to retain the `nvm` steps that I've removed here.  I'll give that end a try tomorrow, as I'm about to sign out for the day :smile: 

Also note that I based this branch off of #981, since we needed the updated `npm install` setup, and I didn't merge that one yet.


alisterscott:
This looks good - we should update our wrapper repos to use the same CircleCI 2.0 containers as CircleCI 1.0 is being deprecated come October

alisterscott:
> I want to run this through a few times before merging it to watch out for timeouts

Is there any difference between running headless chrome, or non-headless chrome under xvfb?

alisterscott:
The caching wasn't working on re-runs in CircleCI so I updated the syntax a bit and it seems to be working now

hoverduck:
> The caching wasn't working on re-runs in CircleCI so I updated the syntax a bit and it seems to be working now

Thanks :smile:

> Is there any difference between running headless chrome, or non-headless chrome under xvfb?

Effectively there shouldn't be.  I would hope that headless chrome would potentially have lower resource utilization since it's not actually drawing the screens, but I don't know that for sure.  It could just as easily use _more_ resources depending on how it's implemented behind the scenes.

I just know that when we turned this on the first time we ended up seeing a lot of unexplained timeouts, and I'm hoping that's been resolved with a newer version of Chrome. :crossed_fingers:

hoverduck:
I've got the Jetpack PR ready here - https://github.com/Automattic/jetpack/pull/8942

They'll both need to be merged at the same time

PR #991, opened 2018-02-28
Move mocha and magellan plugin to Automattic repos

hoverduck:
I've transferred the repos from my personal account into the Automattic organization.  GitHub handles this behind the scenes with redirects (as long as I don't create another repo with the same name), so it's pretty seamless.  But this PR updates the dependencies to pull directly from Automattic anyway.

I've also updated the mocha repo from v3.2.0 to v5.0.1 and  the magellan-mocha-plugin from v7.0.1 to v9.0.1, so this PR reflects the new commit hashes for that.

Note that the updated mocha does give this warning, which I haven't attempted fixing yet:
`DeprecationWarning: "--compilers" will be removed in a future version of Mocha; see https://git.io/vdcSr for more info`

cc: @Automattic/flowpatrol


alisterscott:
> Note that the updated mocha does give this warning, which I haven't attempted fixing yet:
DeprecationWarning: "--compilers" will be removed in a future version of Mocha; see https://git.io/vdcSr for more info

Fixed here: #994

PR #992, opened 2018-02-28
Update Chromedriver to v2.35.0

hoverduck:
I saw one generic `timeout` failure on this build: https://circleci.com/gh/Automattic/wp-e2e-tests/17522

I think this is what we were seeing in bulk with the HEADLESS option previously (#754).  I'm just updating Chromedriver to the latest version to rule that out as a potential cause if the issue escalates.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "babel-runtime": "^6.3.19",
     "bluecat": "^1.1.6",
     "chai": "^3.4.1",
-    "chromedriver": "2.33.2",
+    "chromedriver": "2.35.0",
     "config": "1.16.0",
     "creditcard-generator": "0.0.7",
     "esformatter-braces": "^1.2.1",


alisterscott:
Does not using the headless option (and using xvfb) still work? Does this result in less potential timeout errors? 

https://github.com/Automattic/wp-e2e-tests/blob/baf54e193e8d8bbbaf21d29a370eac9739763205/run.sh#L173

hoverduck:
> Does not using the headless option (and using xvfb) still work? Does this result in less potential timeout errors?

It doesn't work by default since the `circleci/node:8.9.4-browsers` image doesn't have xvfb installed.  If there does end up being an issue with `--headless` we'll have to add a setup step to install it.

alisterscott:
I think this was missing the package lock file so this means npm install makes changes to this file - I've created a new PR for this #993

hoverduck:
Yeah, my bad.  I was jumping around between a bunch of branches today and lost track of where I'd run `npm install`.  I re-ran everything and diffed your updated package-lock file in that PR and it looks good.

Coincidentally, one of the changes in npm 5.6.0 is that they fixed a bug that prevented package-lock from being generated the same on all platforms, and also added a --package-lock-only option to avoid having to do a full install

PR #996, opened 2018-03-01
Try to fix headless

alisterscott:


alisterscott:
Doesn't make any difference

PR #998, opened 2018-03-01
New signout selector

alisterscott:
Remove the old selector


alisterscott:
This doesn't seem to make any difference :/

PR #999, opened 2018-03-01
Add retry to signout

hoverduck:
Recently I've seen a lot of failures on logout at the mobile screen width.  My assumption is that it's because the sidebar is sliding in with an animation, and the Sign Out button isn't catching the click.  This PR just adds a retry to clicking Sign Out.  I tried adding a better wait status to the `_closeProfileViewOnMobile()` function to ensure the sidebar was fully loaded, but I still saw the problem in 1/10 runs.


alisterscott:
Thanks for this fix 👍

PR #1010, opened 2018-03-07
Do deeper check on ABtest values so that order doesn't matter

bsessions85:
Used lodash to compare the A/B test values so that it doesn't fail if the order is different

Also updated error message to include curly brackets so it doesn't lead people down the wrong path if it fails.


bsessions85:
Merging this one in to reduce slack noise

alisterscott:
Thanks for doing this @bsessions85

Issue #329, opened 2016-11-25
Should be aware of control flow deprecation


manoj9788:
Just raising this, work on changing parts where control flow is used.  If you guys are already aware and have plans to  change we can close this.
https://github.com/SeleniumHQ/selenium/issues/2969

hoverduck:
Thanks @manoj9788....Selenium 3.0.0 is on my radar, but I hadn't actually spent much time looking at the new async stuff yet.  We'll leave this issue open to make sure it's on our minds as we move forward.

cc: @alisterscott

manoj9788:
@hoverduck Sure, am happy to help. I will soon put a new repository on this showcasing a demo.

alisterscott:
We probably need to prioritize this now that we're on Node 8.9.x

alisterscott:
We should move to using async/wait: https://github.com/SeleniumHQ/selenium/wiki/WebDriverJs

PR #1013, opened 2018-03-08
Try to make #changeToNewRole a bit more reliable

brbrr:
There are 2 buttons with the same locators in the EditTeamMemberPage. 'Update' button is in `disabled` state it role was not changed.

This PR makes `#changeToNewRole` click only on active 'Update' button, also it chains the driver-related calls to guarantee synchronous execution

### DIFF
--- a/lib/pages/edit-team-member-page.js
+++ b/lib/pages/edit-team-member-page.js
@@ -19,9 +19,15 @@ export default class EditTeamMemberPage extends BaseContainer {
 	}
 
 	changeToNewRole( roleName ) {
-		DriverHelper.clickWhenClickable( this.driver, By.css( `select#roles option[value=${roleName}]` ) );
-		DriverHelper.clickWhenClickable( this.driver, By.css( 'button.form-button.is-primary' ) );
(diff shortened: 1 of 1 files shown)


brbrr:
It is different :)

So while I tried to debug CircleCI failures I noticed that Invites test fails after that method call (https://github.com/Automattic/wp-e2e-tests/blob/master/specs/wp-invite-users-spec.js#L706)
Not sure why, but the local run was clicking on the disabled button(before it selects 'Author' from dropdown)

So here I changed button locator since there 2 buttons located with `button.form-button.is-primary` and it should click only not disabled button.

> I don't really understand why you'd need to chain these actions here?

In my understanding, all(or most) of the `driver` and `driverHelper` methods return a promise. And in my experience, WebDriverJS promise manager isn't reliable enough, that's why I prefer to chain all these promise calls whenever it's needed or not. As a side benefit - it would be much easier to convert `then` chains into `await` calls when we'll move to `async/await` style (https://github.com/Automattic/wp-e2e-tests/issues/329)

brbrr:
Its not needed

PR #1018, opened 2018-03-08
Add overrides for all known A/B tests

rachelmcr:
While reviewing visual regression tests results I noticed that some a/b tests were causing failures in Applitools due to testing visual (but not functional) changes.

Previously we only added A/B test overrides for tests that made functional changes. Adding them for all A/B tests will guard against unnecessary visdiff failures or noise.

Related update to Calypso docs: https://github.com/Automattic/wp-calypso/pull/23123

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -70,8 +70,8 @@
     "redesignedSidebarBanner"
   ],
   "overrideABTests": [
-	[ "domainSuggestionTestV6_20180301", "group_0" ],
-	[ "upgradePricingDisplayV2_20180305", "original" ],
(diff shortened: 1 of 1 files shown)


alisterscott:
Oh, I didn't even think about this.
So any new AB test we should add an override because even if it's a new label or UI element it can affect the visual tests.
Thanks for doing this - I'll keep this in mind now. Should we post this to a P2?

rachelmcr:
Good idea about posting to a p2 — I guess the documentation update only helps people unfamiliar with the existing process. :) Internal ref: p4TIVU-8R8-p2

PR #1021, opened 2018-03-09
Switch JN tests to use stable Jetpack

brbrr:
Workaround for https://github.com/Automattic/jetpack/issues/9020

Actually, Jetpack stable was used after fffa9b0e4a1a985c5d70ef76f72099bf64442395 (overlooked while refactoring `JetpackConnectFlow`). This PR make just makes it clearly visible, and fixing one more issue along the way.


alisterscott:
LGTM

Issue #1019, opened 2018-03-08
Visdiffs: Fix a/b test overrides (logged in and out)


rachelmcr:
I'm seeing two problems in a/b test behavior in our visual regression tests:

1) Logged out: Even when the a/b test overrides are set in the signup flow, it isn't always changing the a/b test behavior. From the results I'm seeing this likely only affects the first page in the signup flow, and because it's just testing a visual change it isn't affecting our functional tests.

On the first page in the signup flow, the `siteGoalsShuffle` test still shows a test variant when the control behavior should be set. I see this in our daily test runs (with CircleCI/Sauce Labs) but also in local test runs and manual tests:

![screenshot 2018-03-08 14 10 39](https://user-images.githubusercontent.com/8658164/37157412-b60f537c-22e0-11e8-81d6-af104a823817.png)

2) Logged in: The change made in https://github.com/Automattic/wp-e2e-tests/pull/954 isn't working for some reason when running the visdiff tests on Sauce Labs; the overrides aren't set in the browser. (Local test runs handle the overrides fine.) Error:

```
The localstorage value for AB tests wasn't set correctly.
Expected value is:
'{"domainSuggestionTestV6_20180301":"group_0","upgradePricingDisplayV2_20180305":"original","businessPlanDescriptionAT_20170605":"original","skipThemesSelectionModal_20170904":"show","checklistThankYouForFreeUser_20171204":"hide","checklistThankYouForPaidUser_20171204":"hide","minimizeFreePlan_20180219":"original","minimizedFreePlanForSignedUser_20180308":"original","minimizedFreePlanForNonSignedUser_20180308":"original","siteGoalsShuffle_20180214":"control","redesignedSidebarBanner":"oldBanner"}'
Actual value is:
'{}'
```

You can see this error by setting `{ "sauce": true, "sauceConfig": "osx-chrome" }` in your local config and running the post editor test spec: `./node_modules/.bin/mocha specs/wp-post-editor-spec.js -g '@visdiff'`

rachelmcr:
@alisterscott Do you have any ideas especially for the second (logged in) issue? I'm wondering if it's a timing issue or something else but I'm not sure the best way to debug it.

I wouldn't spend more than 1-2 hours on it (per Lance's suggestion, given that it's low priority at the moment) but figured I'd ask since you worked on that override PR recently.

alisterscott:
Hi @rachelmcr 

[...]

PR #1023, opened 2018-03-09
Try setting AB tests before the start page

alisterscott:
This sets the AB test overrides by visiting /log-in before then visiting the start page when they'll already be set.

Fixes the first part of #1019


rachelmcr:
Good idea! I like the simplicity of https://github.com/Automattic/wp-e2e-tests/pull/1024 so let's try that first. If any issues arise from that approach we can give this a try.

PR #1028, opened 2018-03-09
Updating test to change case of Remove text

bsessions85:
The remove case changed. updating test


bsessions85:
Merging this to get tests passing

alisterscott:
Was this an expected change? Could we have case insensitive match to stop this happening?

alisterscott:
Now I think about it we probably need some form of data attribute as this will only work in English?

bsessions85:
That is a good point.  I'll look in to it this morning and either fix or put in issue to fix later depending on how complicated it is.

PR #1029, opened 2018-03-09
Trying to skip through happychat prompt

bsessions85:
There is now a modal asking if you want to try happy chat before you cancel. Added an extra next button click to skip it.


alisterscott:
It looks like this only shows when happychat is available, which means that it sometimes won't show and then this test will fail...

alisterscott:
Do you know when and why this started happening? A link to a wp-calypso PR?

PR #1032, opened 2018-03-12
Adds a check around Happychat being available

alisterscott:
https://github.com/Automattic/wp-e2e-tests/pull/1029 introduced an additional step which doesn't always appear

This makes it conditional

I'm still not sure why we recently only started seeing this.

screen shot 2018-03-12 at 1 44 43 pm

### DIFF
--- a/lib/pages/cancel-purchase-page.js
+++ b/lib/pages/cancel-purchase-page.js
@@ -19,12 +19,14 @@ export default class CancelPurchasePage extends BaseContainer {
 		const e2eReason = 'e2e testing';
 		const dialogClass = '.cancel-purchase__button-warning-dialog';
 		const buttonDialogClass = '.dialog__action-buttons';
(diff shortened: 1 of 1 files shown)


alisterscott:
@Automattic/flowpatrol just a FYI I had to merge this as it was a fix for failing tests. I'm still not sure why we haven't seen this prompt before a few days ago - it's been in the wp-calypso codebase for some time

rachelmcr:
Happiness often has controls to enable support/chat in different areas, without having to make code changes. I searched and pre-cancellation chat was enabled on March 9 (internal ref p7DVsv-4pq-p2, more details p7DVsv-4hm-p2) which is why this only started appearing now.

bsessions85:
Good to know, thanks @rachelmcr! And thanks for fixing @alisterscott!

PR #1038, opened 2018-03-13
Remove viewers before adding new viewer

brbrr:
This PR fixes the `Inviting New User as a Viewer of a Private Site` test which is failing from time to time (due to viewers didn't get deleted after test)

### DIFF
--- a/specs/wp-invite-users-spec.js
+++ b/specs/wp-invite-users-spec.js
@@ -404,6 +404,12 @@ testDescribe( `[${host}] Invites:  (${screenSize})`, function() {
 					} );
 				} );
 
+				test.it( 'Can remove all viewers', function() {
+					return this.peoplePage.selectViewers()
+					.then( () => this.peoplePage.emptyUsers() )
+					.then( () => this.peoplePage.selectTeam() );
+				} );
+
 				test.it( 'Can invite a new user as an viewer', function() {
 					this.peoplePage.inviteUser();
 					this.invitePeoplePage = new InvitePeoplePage( driver );


alisterscott:
I'm curious what would happen if this removed viewers presently being added by another test happening in parallel (like another CircleCI build) - this could cause some non-deterministic results if the viewer was removed halfway through a test

brbrr:
That's good catch! What approach you think would be better? remove the user in `after` step?

I'm not sure why, but some users did not get deleted, and they stack up

alisterscott:
🤔 I _think_ this might be the reason for this failure: https://circleci.com/gh/Automattic/wp-e2e-tests/18008

alisterscott:
> I'm not sure why, but some users did not get deleted, and they stack up

Probably just failed test runs that are bailed and never get to the step to remove the viewer

> remove the user in after step?

We already do that in the scenario:

```
test.describe( 'As the original user, I can remove the new user added to site', function() {
										test.it( 'Can remove the team member from the site', function() {
											this.peoplePage.removeUserByName( newUserName );
											this.peoplePage.waitForSearchResults();
											this.peoplePage.viewerDisplayed( newUserName ).then( ( displayed ) => {
												assert.equal( displayed, false, `The username of '${newUserName}' was still displayed as a site viewer` );
											} );
										} );
```

> What approach you think would be better?

Probably a wp.com batch job that runs say once a week on the server outside of hours that gets rid of them all from the site - similar to what we do for the other site's content

alisterscott:
Added an issue to track this #1049 
And a PR to revert this #1048

brbrr:
> We already do that in the scenario:

I mean 'after' hook which would be executed test failed or passed

alisterscott:
> I mean 'after' hook which would be executed test failed or passed

Yes, that would probably be good, just need to make sure the state is persisted across blocks (using this. probably) and cater for what happens if it fails before creating the account

PR #1039, opened 2018-03-13
Use a selector to verify and remove viewers

bsessions85:
We were using a text match which broke when case changed and will not work with other translations. Using selectors now instead.

### DIFF
--- a/lib/pages/people-page.js
+++ b/lib/pages/people-page.js
@@ -77,11 +77,11 @@ export default class PeoplePage extends BaseContainer {
 	}
 
 	viewerDisplayed( username ) {
-		let viewers = this.driver.findElements( By.css( '.people-list-item' ) );
+		let viewers = this.driver.findElements( By.css( '.people-list-item .people-profile__username' ) );
 
 		return webdriver.promise.filter( viewers, function( viewer ) {
 			return viewer.getText().then( function( text ) {
-				return text === `${username}\n@${username}\nRemove`;
+				return text === username;
 			} );
 		} ).then( function( filteredViewers ) {
 			return ( filteredViewers.length === 1 );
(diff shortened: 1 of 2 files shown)


lib/pages/people-page.js
```diff
@@ -94,19 +94,17 @@ export default class PeoplePage extends BaseContainer {
 	 * @param {bool} useSince Whether to include the "SINCE" line in the pattern match (for Followers, not Viewers)
```
brbrr: I think this should be removed as not needed

lib/pages/people-page.js
```diff
@@ -94,19 +94,17 @@ export default class PeoplePage extends BaseContainer {
...
-				let re = new RegExp( `${username}\n@${username}\nRemove` );
-
-				if ( useSince ) {
-					re = new RegExp( `${username}\n@${username}\nSINCE.*\nREMOVE` );
-				}
-
-				return text.match( re );
-			} );
+			return viewer.findElement( By.css( '.people-profile__username' ) )
+				.then( ( profileUsernameElement ) => {
```
brbrr: what about to convert it into a one-liner `( profileUsernameElement => profileUsernameElement.getText() )?

lib/pages/people-page.js
```diff
@@ -94,19 +94,17 @@ export default class PeoplePage extends BaseContainer {
...
+					return profileUsernameElement.getText();
+				} )
+				.then( function( text ) {
```
brbrr: one-liners seems to look a bit clearer in such cases IMO

lib/pages/people-page.js
bsessions85: Thanks! updated

lib/pages/people-page.js
bsessions85: Done! Thanks!

bsessions85:
Thanks @brbrr! Changes made

PR #1043, opened 2018-03-14
Updated class selector to reflect new component in PR#23253

ramonjd:
In Calypso we've made [changes](https://github.com/Automattic/wp-calypso/pull/23253#issuecomment-372873699)  to the date picker, affecting the post editor e2e tests:

`(36 / 38) <-- Worker 2 FAIL (will retry).  Spent 38068 msec [WPCOM] Editor: Posts (mobile) Schedule Basic Public Post @parallel @jetpack @chrome`

This PR updates `nextMonthSelector` in `PostEditorSidebarComponent.chooseFutureDate` to use the [new class name](https://github.com/Automattic/wp-calypso/blob/ffde314eea1102b73180ad900ce3e8f6e8d90205/client/components/date-picker/style.scss#L67).

### DIFF
--- a/lib/components/post-editor-sidebar-component.js
+++ b/lib/components/post-editor-sidebar-component.js
@@ -228,7 +228,7 @@ export default class PostEditorSidebarComponent extends BaseContainer {
 
 	// Selects the first day of the second week of next month - to (hopefully) always select a future date on the calendar
 	chooseFutureDate() {
-		const nextMonthSelector = By.css( '.DayPicker-NavButton--next' );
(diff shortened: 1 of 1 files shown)


alisterscott:
@ramonjd Thanks for the update! This works well, but do you want to use `.DayPicker-NavButton--next,.date-picker__next-month` so it works on both versions for now and we can merge it, and then we can clean this up in a subsequent PR to just use the new selector? What do you think?

ramonjd:
> This works well, but do you want to use .DayPicker-NavButton--next,.date-picker__next-month

I like that idea!! I'll update. Thanks :)

Issue #1044, opened 2018-03-14
Investigate CircleCI errors: "Too long with no output (exceeded 10m0s)"


alisterscott:
We started seeing these around the same time we moved to using the [CircleCI prebuilt docker browser containers](https://hub.docker.com/r/circleci/node/) (over our own docker container)

Example build: https://circleci.com/gh/Automattic/wp-e2e-tests/17949

```
#!/bin/bash -eo pipefail
./scripts/run-wrapper.sh

> wp-e2e-tests@0.0.1 test /home/circleci/wp-e2e-tests
> ./run.sh ${TESTARGS:-$npm_package_defaultTestArgs}

Parallel execution details:
mobile=0, desktop=1, node=0, total=2
Executing tests at mobile screen width
Magellan 8.8.6
Will try to load configuration from default of ./magellan.json
Loaded configuration from:  /home/circleci/wp-e2e-tests/magellan.json
Magellan is creating temporary files at: /home/circleci/wp-e2e-tests/temp
Sauce configuration OK
The $NODE_CONFIG environment variable is malformed JSON
[INFO] [Mocha Plugin] Scanning test files for tests ...
[INFO] [Mocha Plugin] Found 38 tests in test files

Running 38 tests with 3 workers with chrome

--> Worker 3, mock port:12006, running test: [WPCOM] View site from sidebar: (mobile) @parallel @jetpack @visdiff @chrome 
--> Worker 2, mock port:12003, running test: [WPCOM] Plans: (mobile) @parallel @jetpack @chrome 
--> Worker 1, mock port:12000, running test: [WPCOM] Managing Domains: (mobile) @parallel @chrome 
(1 / 38) <-- Worker 2 PASS  [WPCOM] Plans: (mobile) @parallel @jetpack @chrome  
--> Worker 2, mock port:12009, running test: [WPCOM] Managing Domain Mapping: (mobile) @parallel @chrome 
(2 / 38) <-- Worker 1 FAIL (will retry).  Spent 72765 msec [WPCOM] Managing Domains: (mobile) @parallel @chrome  
--> Worker 1, mock port:12012, running test: [WPCOM] Editor: Media Upload (mobile) @parallel @jetpack @chrome 
(3 / 38) <-- Worker 2 PASS  [WPCOM] Managing Domain Mapping: (mobile) @parallel @chrome  
[...]

PR #1045, opened 2018-03-14
Try removing no-sandbox option for chrome

alisterscott:
Seeing whether this helps with #1044


alisterscott:
I've run this a number of times but the absence of the error doesn't mean this has indeed fixed it

brbrr:
We can try to merge it and see how it goes. Btw why do we use that option? 

In my past job I've used `--no-sandbox` to make chrome with `xvfb` work on openVZ containers

alisterscott:
> Btw why do we use that option?

I think it was there for legacy when we were using xvfb on older versions of chrome etc. We just never removed it

Issue #923, opened 2018-02-09
Create Style Guide for our e2e test framework


brbrr:
Together with a code style, we can use some kind of high-level style guide for our components/tests.

The most notable thing which differs the most - is the way we define element locators. There are other things that might be standardized. 

This ticket is mostly the place to discuss what things we can standardize.

brbrr:
- Passing arrow functions (“lambdas”) to Mocha is discouraged. Lambdas lexically bind this and cannot access the Mocha context. https://mochajs.org/#arrow-functions

alisterscott:
We now have a style guide, at least a start, feel free to add to it via PRs and examples 👍 

docs/style-guide.md

PR #1050, opened 2018-03-15
Clean up sign up spec & add start of the style guide

alisterscott:
Cleans up the sign up spec

Adds a domain registration details test data helper

Adds the beginning of a style guide for writing tests

Adds filling in the "about" page in one of the sign up scenarios (we didn't ever do this previously)

Includes some parts of #923


alisterscott:
@Automattic/flowpatrol this is a HUGE PR, but I wanted to clean things up in this spec, hopefully it will make the sign up tests faster and determistic, and also easier to read.

I've started a style guide as recommended by @brbrr  in #923

lib/pages/signup/about-page.js
```diff
@@ -9,6 +9,26 @@ export default class AboutPage extends BaseContainer {
 	}
 
 	submitForm() {
-		return driverHelper.clickWhenClickable( this.driver, By.css( '.about__submit-wrapper button.is-primary' ), this.explicitWaitMS );
+		return driverHelper.clickWhenClickable( this.driver, By.css( '.about__submit-wrapper button.is-primary' ) );
+	}
+
+	enterSiteDetails( siteTile, siteTopic, { showcase = false, share = false, sell = false, educate = false, promote = false } = {} ) {
```
rachelmcr: Small typo in this and the next line: `SiteTile` -> `SiteTitle`.

specs/wp-signup-spec.js
```diff
@@ -271,96 +229,86 @@ testDescribe( `[${host}] Sign Up  (${screenSize}, ${locale})`, function() {
...
 		} );
 
 		test.describe( `Step ${stepNum}: About Page`, function() {
 			stepNum++;
 
 			test.it( 'Can see the about page', function() {
-				this.startPage = new StartPage( driver, { visit: true, culture: locale } );
-				this.aboutPage = new AboutPage( driver );
-				return this.aboutPage.displayed().then( ( displayed ) => {
+				new StartPage( driver, { visit: true, culture: locale } ).displayed();
```
rachelmcr: Should we also break this out into its own step, as in the previous test (lines 87-95)?

specs/wp-signup-spec.js
```diff
@@ -505,38 +431,27 @@ testDescribe( `[${host}] Sign Up  (${screenSize}, ${locale})`, function() {
 								} );
 							} );
 
-							test.it( 'The sign up processing page will finish and show a \'Continue\' button', function() {
-								return this.signupProcessingPage.waitForContinueButtonToBeEnabled();
-							} );
-
-							test.it( 'Clicking the \'Continue\' button continues the process', function() {
-								return this.signupProcessingPage.continueAlong();
+							test.it( 'Can then see the sign up processing page and it will finish and show a \'Continue\' button, which is clicked', function() {
```
rachelmcr: With this change it looks like you can remove lines 427-432 above (starting with `test.it( 'Can then see the sign up processing page', function() {` ).

rachelmcr:
The style guide looks great! I started to review the signup spec but only got about halfway through. I skimmed through the rest of the changes but it would be great for another reviewer to take a look at this, too.

lib/pages/signup/about-page.js
alisterscott: Thanks 👍

specs/wp-signup-spec.js
alisterscott: Good idea
I didn't enter any of the "about" information in this scenario, I thought entering a random site title could cause the visual difference tests to complain about the dynamic information? What do you think? Just submit that form as we did before?

specs/wp-signup-spec.js
alisterscott: Yes, good catch 👍

alisterscott:
@brbrr I'm keen to get this merged today - to avoid conflicts - would you have time to add additional review on @rachelmcr's above? It's a big PR but it kinda needs to fit together.

specs/wp-signup-spec.js
```diff
@@ -271,96 +229,89 @@ testDescribe( `[${host}] Sign Up  (${screenSize}, ${locale})`, function() {
...
+						return pickAPlanPage.displayed().then( ( displayed ) => {
 							eyesHelper.eyesScreenshot( driver, eyes, 'Plans Page' );
 							return assert.equal( displayed, true, 'The pick a plan page is not displayed' );
 						} );
 					} );
 
 					test.it( 'Can select the premium plan', function() {
-						return this.pickAPlanPage.selectPremiumPlan();
+						const pickAPlanPage = new PickAPlanPage( driver );
+						return pickAPlanPage.selectPremiumPlan();
```
brbrr: Can we combine this step and step above? This should become a bit shorter

alisterscott:
@brbrr 

> test.describe( 'Step ${stepNum}: ....

Just a way to print step numbers like Step 1 - Domains, Step 2 Plans, Step 3 Checkout etc

We used to write them out like that but the steps constantly reorder and get out of numbered sync - so we put them dynamically

Any ideas to improve?

specs/wp-signup-spec.js
alisterscott: Good idea - thanks

brbrr:
When I run these tests locally - these `describe` blocks did not help me in debugging, they are just messing the output.

I would get rid of these blocks completely. We don't have them anywhere in other test files, and pattern that we use, 1 `describe` block per test, works well for me.

as an improvement, we could also try to combine some of the common actions into flow file.

alisterscott:
> When I run these tests locally - these describe blocks did not help me in debugging, they are just messing the output.
> I would get rid of these blocks completely. We don't have them anywhere in other test files, and pattern that we use, 1 describe block per test, works well for me.

Yes, legacy code before we could bail, so we had to nest

> as an improvement, we could also try to combine some of the common actions into flow file.

Yes, we have some of that already but would be worth trialling - separate PR though 😉

alisterscott:
Merging this while it's all good ✅

alisterscott:
Thanks for all your feedback @rachelmcr and @brbrr

specs/wp-signup-spec.js
rachelmcr: Yes, agreed, I wouldn't enter any "About" information here. If we wanted to enter information I'd do it after the screenshot is taken for the visual diff test, but also I think it's worth having a test that verifies you can continue without entering information here.

PR #1051, opened 2018-03-15
Try to make slack team ping works

brbrr:
I've disabled per test notifications and make some refactoring.

To test:
0. `npm install` (there is a change in `spec-xunit-reporter`) 
1. `export CIRCLE_BRANCH=master`
2. break some of the test specs
3. run that spec file with: `./node_modules/.bin/magellan --config=./magellan.json --test=./path/to/spec`


lib/reporter/magellan-reporter.js
```diff
@@ -251,27 +143,139 @@ Reporter.prototype.listenTo = function( testRun, test, source ) {
...
+// Only enable Slack messages on the master branch
+function getSlackClient() {
+	if ( process.env.CIRCLE_BRANCH === 'master' ) {
+		let slackHook = configGet( 'slackHook' );
+		return slack( slackHook );
+	}
+	return { send: function() {} };
+}
+
+// function sendNotifPerFailure( slackClient, reportDir, testRun ) {
```
alisterscott: Do we need these as it's commented out

PR #1052, opened 2018-03-15
Enable Pressable NUX flow only for PRESSABLE target

brbrr:
As discussed with Systems - current approach is not desirable, We agreed to limit this test to 8 times/day

Internal ref: p1521056822000619-slack-pressable-public


alisterscott:
Cool, so this just runs when Pressable target runs which is the scheduled runs only? (I guess pushing to a branch with jetpack will also run this but that's negligible?)

brbrr:
Yes, the intention is to make it run only with PRESSABLE target - scheduled runs only and I do think that runs on jetpack branches are negligible.

PR #1056, opened 2018-03-16
Attempting to use the bail_time argument for magellan

bsessions85:
Adding the bail_time value to all the magellan configuration files. It is currently set at 5 minutes.


alisterscott:
Changes look fine to me - I guess over time we'll see whether this improves stability on CircleCI

@rachelmcr do you think the visdiff timeout is long enough?

rachelmcr:
>do you think the visdiff timeout is long enough?

This looks ok to me. If we notice any timeout issues with the visdiff runs we can loop back to this on our next visdiff iteration.

PR #1057, opened 2018-03-19
Add connect flow for Jetpack.com Paid plan from Pricing page

brbrr:
Jetpack.com Pricing page, selecting a paid plan e.g. Premium (hits jetpack/connect/premium flow)

### DIFF
--- a/specs-jetpack-calypso/wp-jetpack-connect-spec.js
+++ b/specs-jetpack-calypso/wp-jetpack-connect-spec.js
@@ -421,21 +421,22 @@ test.describe( `Jetpack Connect: (${ screenSize })`, function() {
 
 		test.it( 'Can log into WP.com', function() {
 			const user = dataHelper.getAccountConfig( 'jetpackConnectUser' );
-			const loginPage = new LoginPage( driver );
-			return loginPage.login( user[0], user[1] );
+			return ( new LoginPage( driver ).login( user[0], user[1] ) );
+		} );
+
+		test.it( 'Can wait for Jetpack get connected', function() {
+			return ( new JetpackAuthorizePage( driver ).waitToDisappear() );
 		} );
 
 		test.it( 'Can see the secure payment page and enter/submit test payment details', function() {
 			const securePaymentComponent = new SecurePaymentComponent( driver );
-			securePaymentComponent.enterTestCreditCardDetails( testCreditCardDetails );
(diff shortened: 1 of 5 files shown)


specs-jetpack-calypso/wp-jetpack-connect-spec.js
```diff
@@ -365,4 +372,70 @@ test.describe( `Jetpack Connect: (${ screenSize })`, function() {
...
+			return wpAdminPluginsPage.connectJetpackAfterActivation();
+		} );
+
+		test.it( 'Can log into WP.com', function() {
+			const user = dataHelper.getAccountConfig( 'jetpackConnectUser' );
+			const loginPage = new LoginPage( driver );
+			return loginPage.login( user[0], user[1] );
+		} );
+
+		test.it( 'Can see the secure payment page and enter/submit test payment details', function() {
```
alisterscott: This failed a few times for me locally, because it was still on the previous "connecting" page which isn't mentioned in this flow, even though this connecting page disappears automatically it might be worth putting in a step that waits for it to disappear.

specs-jetpack-calypso/wp-jetpack-connect-spec.js
```diff
@@ -365,4 +372,70 @@ test.describe( `Jetpack Connect: (${ screenSize })`, function() {
...
+			const securePaymentComponent = new SecurePaymentComponent( driver );
+			securePaymentComponent.enterTestCreditCardDetails( testCreditCardDetails );
+			securePaymentComponent.submitPaymentDetails();
+			securePaymentComponent.waitForCreditCardPaymentProcessing();
+			return securePaymentComponent.waitForPageToDisappear();
+		} );
+
+		test.it( 'Can see Premium Thank You page', function() {
+			const pickAPlanPage = new CheckOutThankyouPage( driver );
+			assert( pickAPlanPage.isPremiumPlan() );
```
alisterscott: I think you could do one line if you wanted:

`assert( new CheckOutThankyouPage( driver ).isPremiumPlan() );`

alisterscott:
I saw this fail a few times on the "completing connection" step which isn't in the list of steps, I think this should be modelled as a page which we wait to disappear, much like the page that creates the customer's account when signing up.

screen shot 2018-03-20 at 2 50 22 pm

PR #1059, opened 2018-03-21
Fix/screensizes in runsh

alisterscott:
The screen sizes weren't surrounded in quotes
Hopefully this fixes IE11 duplicate runs


alisterscott:
@Automattic/flowpatrol I'm merging this as no one is around I want to see whether this fixes the issue #1060

PR #1062, opened 2018-03-21
Fix the selector prefix for the plan action buttons

taggon:
The e2e test will be broken when we roll out [the winner of the minimizedFreePlan test](https://github.com/Automattic/wp-calypso/pull/23489). This PR will fix the problem by changing the selector prefix into .plans-features-main, a higher level container.


alisterscott:
Thanks for the update @taggon 

I was thinking we could merge something that will work with the old _and_ new screens like:

```

	_selectPlan( level ) {
		let oldPrefix = 'table';
		const newPrefix = '.plans-features-main';

		if ( this.host !== 'WPCOM' && screenSize === 'mobile' ) {
			oldPrefix = '.plan-features__mobile-plan';
		}

		const selector = By.css( `${oldPrefix} button.is-${level}-plan:not([disabled]),${newPrefix} button.is-${level}-plan:not([disabled])` );
		return driverHelper.clickWhenClickable( this.driver, selector );
	}
```

so the e2e tests will work any either behaviour.

Once your PR has landed, we can merge a simple change here to remove the old selector completely.

What do you think?

alisterscott:
I've created a new PR to work with both old and new designs #1077 

We can use this one to remove the old one once https://github.com/Automattic/wp-calypso/pull/23489  is merged

taggon:
@alisterscott This is not a breaking change. The new selector will completely cover the old cases since the `table` and the `.plan-features__mobile-plan` are descendants of the `.plan-features-main` element.
image

alisterscott:
Oh, sorry about that. Then this is good to go, sorry you may need to merge it manually as I already merged that other one :(

taggon:
@alisterscott It's okay. I just rebased this branch. :)

taggon:
Merged. Thanks @alisterscott for handling this. :)

Issue #1064, opened 2018-03-21
Customizer tests are troublesome


alisterscott:
The customizer tests have ongoing issues with reliability, I'd like to remove them, since the themes tests just make sure the customizer can load correctly.

Anything beyond this isn't part of calypso, and I don't think we should focus on expansive testing within our e2e tests.

I'll create a PR to temporarily disable this spec, and then seeing how this goes, I will remove the spec and the page object code after that

PR #1065, opened 2018-03-21
Try no customizer specific tests

alisterscott:
See #1064


alisterscott:
@Automattic/flowpatrol heads-up I am trialling this for a while

PR #1078, opened 2018-03-23
Private invites test: Delete user even if test get failed

### LINKED ISSUE #1049: Work out way to delete private site viewers
These build up over time, and #1038 has issues when running tests in parallel

brbrr:
fixes #1049


alisterscott:
Thanks for this 👍

PR #1081, opened 2018-03-23
Commenting out these lines to get working for the weekend

bsessions85:
I'm not sure what this selector was for so I don't know what it is looking for. Possibly the undo link on the confirmation?


bsessions85:
I'm going to merge this to get the test passing over the weekend, but I would like someone to look and see if they know what that selector is supposed to be looking for. Possibly related to #233771234-gh-wp-calypso release

alisterscott:
This was unintended behaviour which was fixed in https://github.com/Automattic/wp-calypso/pull/23377

I'll update the test to be clearer now

PR #1083, opened 2018-03-26
Fix i18n tests to work locally

brbrr:
- Add `bailSuite` to stop execution if one of the tests get failed
- add `sleep` delay after switching to iframe to prevent `{"code":-
32000,"message":"Cannot find context with specified id"}` as suggested [here](https://stackoverflow.com/questions/41429723/unhandled-error-cannot-find-context-with-specified-id-using-robot-framework)
- remove debug mocha option `-d` from magellan config

To test:
`$ TARGET=I18N ./node_modules/.bin/magellan --config=./magellan-i18n.json`

### DIFF
--- a/specs-i18n/search-wordpress-on-google.js
+++ b/specs-i18n/search-wordpress-on-google.js
@@ -34,6 +34,7 @@ function doGoogleAdSearch( search_params ) {
 
 	test.describe( description + ' @i18n (' + locale + ')', function() {
 		this.timeout( mochaTimeOut );
+		this.bailSuite( true );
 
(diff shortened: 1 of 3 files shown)


specs-i18n/search-wordpress-on-google.js
```diff
@@ -34,6 +34,7 @@ function doGoogleAdSearch( search_params ) {
 
 	test.describe( description + ' @i18n (' + locale + ')', function() {
 		this.timeout( mochaTimeOut );
+		this.bailSuite( true );
```
akirk: Nice!

PR #1085, opened 2018-03-26
Disable saveAllScreenshots for i18n builds

brbrr:


akirk:
This is to fix the issue with many screenshots posted to Slack. I am not fully convinced that this is the right solution, though as it seems that in #1051 the underlying logic was changed. We had months of only failures posted to our Slack channel with this setting on.

brbrr:
Closing this as this approach isn't desired

PR #1087, opened 2018-03-26
Set TEMP_ASSET_PATH in the magellan reporter

hoverduck:
@brbrr I _think_ this does the job for setting the path variable so it's accessible to the media helper.  I still have no idea where it was coming from before, though.

It looked good for me locally, but I didn't have a chance to check it against the i18n repo.

### DIFF
--- a/lib/reporter/magellan-reporter.js
+++ b/lib/reporter/magellan-reporter.js
@@ -14,6 +14,9 @@ let Reporter = function() {
 util.inherits( Reporter, BaseReporter );
 
 Reporter.prototype.listenTo = function( testRun, test, source ) {
+	// Put the tmpAssetPath into an envvar so media-helper can find it for saving screenshots
+	process.env.TEMP_ASSET_PATH = testRun.tempAssetPath;
+
 	// Print STDOUT/ERR to the screen for extra debugging
 	if ( !! process.env.MAGELLANDEBUG ) {
 		source.stdout.pipe( process.stdout );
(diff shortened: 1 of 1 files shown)


brbrr:
Yeah, this does the job, kinda. After my investigations - I can say that `magellan-reporter` is called _after_ `spec-xunit-reporter`(which in charge of creating xunit reports).  `spec-xunit-reporter` need to know the assets path too, but at that point - `TEMP_ASSET_PATH` isn't set yet. 

To make `magellan-reporter` actually work as expected - you need to update `reportDir`/`screenshotDir` variables too:

```
// const reportDir = './reports';
// let screenshotDir = './screenshots';
const reportDir = testRun.tempAssetPath + '/reports';
let screenshotDir = testRun.tempAssetPath + '/screenshots';
```

After that try to run any test(which will fail):
`rm -rf temp/ screenshots/ reports/ && CIRLCE_BRANCH=master ./node_modules/.bin/magellan --config=./magellan.json --test=specs/wp-test-spec.js `

`wp-test-spec.js` =>
```
import assert from 'assert';
import test from 'selenium-webdriver/testing';
import * as driverManager from '../lib/driver-manager.js';

test.describe( 'TEST 1 @parallel', function() {
	test.it( 'fail', () => {
		driverManager.startBrowser();
		assert.equal( true, false, 'Oopsie doopsie' );
	} );
} );
```

you should see `ENOENT: no such file or directory, scandir '.../wp-e2e-tests/temp/build-magellan-XXXXXX_XXXXXXX__temp_assets/reports'`

also, check `./temp` folder - it will include 2 assets folders, and one of them would be empty. 

It is super easy to fix that with this change: https://github.com/Automattic/magellan-mocha-plugin/pull/2

brbrr:
I'm going to merge this and see how it will work out. Thanks, @hoverduck, I appreciate your help!

PR #1088, opened 2018-03-26
Jetpack sites not being disconnected

bsessions85:
The tests are leaving a lot of jetpacks sites and not clearing out the ones that are not reachable anymore. This runs a cleanup after the jetpack connect tests to clean up unreachable sites on the jetpackconnect user

### DIFF
--- a/specs-jetpack-calypso/wp-jetpack-connect-spec.js
+++ b/specs-jetpack-calypso/wp-jetpack-connect-spec.js
@@ -50,11 +50,6 @@ test.before( function() {
 test.describe( `Jetpack Connect: (${ screenSize })`, function() {
 	this.timeout( mochaTimeOut );
 
-	test.after( function() {
-		const jnFlow = new JetpackConnectFlow( driver, 'jetpackConnectUser' );
-		return jnFlow.removeSites();
-	} );
-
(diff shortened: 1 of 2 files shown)


specs-jetpack-calypso/wp-jetpack-connect-spec.js
```diff
@@ -50,6 +50,11 @@ test.before( function() {
 test.describe( `Jetpack Connect: (${ screenSize })`, function() {
 	this.timeout( mochaTimeOut );
 
+	test.after( function() {
+		const jnFlow = new JetpackConnectFlow( driver, 'jetpackConnectUser' );
```
brbrr: I think we just need to use `jetpackConnectUser` in `Disconnect expired sites` block. Looks like that username parameter was lost in one of refactors.

brbrr: Yeah, I was confused by that magic too :)

Thanks for the update!

brbrr:
:shipit:  LGTM!

bsessions85:
@brbrr  Can you please take a look at the last commit? It was failing for mobile because the ensureSidebarMenuVisible() function, which is used quite a bit, wasn't actually doing anything. I updated to actually do a check.

brbrr:
Looks good :)

PR #1099, opened 2018-03-29
Add a refresh if jurassic ninja doesn't show

alisterscott:
Sometimes this screen appears:

screen-shot-2018-03-12-at-20-20-48

All it needs is a refresh, which this PR now does

### DIFF
--- a/lib/pages/wporg-creator-page.js
+++ b/lib/pages/wporg-creator-page.js
@@ -46,6 +46,11 @@ export default class WporgCreatorPage extends BaseContainer {
 	}
 
 	waitForWpadmin() {
+		driverHelper.isEventuallyPresentAndDisplayed( this.driver, PASSWORD_ELEMENT ).then( ( wpAdminDisplayed ) => {
+			if ( !wpAdminDisplayed ) {
+				this.driver.navigate().refresh();
+			}
+		} );
 		return driverHelper.waitTillPresentAndDisplayed( this.driver, PASSWORD_ELEMENT );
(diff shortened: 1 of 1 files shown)


alisterscott:
Feel free to merge this if it helps

sirreal:
I believe this could happen on any Jurassic Ninja page, not just at this step, maybe @oskosk can confirm.

Perhaps there's a way to add some sort of helper that checks for the expected content of the error page and refreshes.

brbrr:
I've done some investigation and took a step forward with what @alisterscott suggested. I noticed that sometimes single restart just not enough. So I defined this function in `driverHelper`:

```
export function refreshIfJNError( driver, timeout = 1000 ) {
	const JNSiteError = by.xpath( "//pre[@class='error' and .='/srv/users/SYSUSER/log/APPNAME/APPNAME_apache.error.log']" );

	const refreshIfNeeded = ( t ) => isEventuallyPresentAndDisplayed( driver, JNSiteError, t )
	.then( ( jnErrorDisplayed ) => {
		if ( jnErrorDisplayed ) {
			return this.driver.navigate().refresh()
			.then( () => refreshIfNeeded( 5000 ) );
		}
		return true;
	} );

	return refreshIfNeeded( timeout );
}
```
which I call from `BaseContainer` constructor. But it won't call that function for the second time. I assume it is because of not enough time for page reload

brbrr:
I'm going to close this in favor of #1107

PR #1101, opened 2018-03-29
Add/prettier package & pre-commit hook

brbrr:
I've added prettier and pre-commit hook to apply changes on modified files (a similar approach used in calypso repo)


brbrr:
I'm going to merge this since this should not affect tests

PR #1105, opened 2018-03-29
Update the domain suggestions test name

hambai:
We're launching a new test today called `domainSuggestionKrakenV313`. I'm removing the old test name and override in favour of the new one.

Should be deployed after https://github.com/Automattic/wp-calypso/pull/23731 is merged

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -61,14 +61,13 @@
     "skipThemesSelectionModal",
     "checklistThankYouForFreeUser",
     "checklistThankYouForPaidUser",
-    "domainSuggestionTestV6",
     "siteGoalsShuffle",
     "upgradePricingDisplayV2",
     "redesignedSidebarBanner",
-    "mobilePlansTablesOnSignup"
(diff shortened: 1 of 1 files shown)


rachelmcr:
You can add the new test to the e2e test config at any time, if you'd like. You don't need to wait for the Calypso PR to be merged (and ideally the e2e tests will have the new test in their config before the new test is merged). It's ok for tests to be in the e2e test config for a little while before and after they are active.

One way to do that is to split this into two PRs, with the new test added right away and the old test removed any time after it's removed from Calypso. (We have a little more about that timing in the [Calypso docs](https://github.com/Automattic/wp-calypso/tree/master/client/lib/abtest#updating-our-end-to-end-tests-to-avoid-inconsistencies-with-ab-tests), for reference.)

hambai:
Ahh... I completely forgot about the e2e tests, and merged the domain suggestion tests yesterday without following up here. Next time I'll be prepared.

PR #1107, opened 2018-03-29
Try to handle Jurassic Ninja errors

brbrr:
This is a take 2 of https://github.com/Automattic/wp-e2e-tests/pull/1099

This PR make Jetpack `master` tests to pass: https://circleci.com/workflow-run/89d96b31-d4b4-4741-960b-925e6b66d90b


oskosk:
I can't find here where is this check is scoped to the `503` cgi kind of a error... Should we scope this fix to only those?

brbrr:
What is `cgi`? In this PR I looking for this element to check if this is a JN error:`//pre[@class='error' and .='/srv/users/SYSUSER/log/APPNAME/APPNAME_apache.error.log']` This isn't the best selector, but AFAIK this `Server Unavailable` error appears only on JN sites

https://github.com/Automattic/wp-e2e-tests/pull/1107/files#diff-0e3b6a0d13153e063b03d83d728e6521R406

oskosk:
There could be `500` errors coming from actual errors on our side (Jetpack's code). And we probably want to be aware of those, and they probably aren't refreshable...

The pattern this code is searching for would match any of the error templates that ServerPilot configures.

cgi is the common gateway interface (the way that binaries communicate with webservers when they're not modules of apache or so). and the error we've been seeing and the only one we can overcome with a refresh is reported with status `503` service unavailable. Meaning PHP couldn't access the cgi interface provided by php-fpm.

PR #1109, opened 2018-03-29
Add Jetpack onboarding tests

bsessions85:
Adding 3 tests to cover various flows through JPO


tyxla:
This is truly wonderful ❤️ Thanks so much for working on those! 💯 

The only thing that occurs to me in the first place is that we could also use a test that starts the flow as a logged in user for a site that is already connected to WP.com. Could be useful for cases where the user is re-running the flow, or just connects the site before going through the flow.

bsessions85:
Thanks for the feedback @tyxla! I have now added a test for that scenario.

specs-jetpack-calypso/wp-jetpack-onboarding-spec.js
```diff
@@ -0,0 +1,432 @@
...
+				.then( tagline => assert.equal( tagline, blogTagline, 'Site tagline not is not correct' ) );
+		} );
+	} );
+
+	test.describe( 'Onboard personal site that has already been connected: @parallel @jetpack', function() {
+		this.bailSuite( true );
+		const blogTitle = dataHelper.randomPhrase();
+		const blogTagline = dataHelper.randomPhrase();
+
+		test.it( 'Can create wporg site', function() {
```
brbrr: This connection flow could be replaced with `JetpackConnectFlow#connectFromWPAdmin`

brbrr:
Thanks for these changes! ❤️

PR #1115, opened 2018-04-04
Fix Pressable NUX flow

brbrr:
Pressable NUX flow was designed in a way that it might delete account's main site. This happened recently 😅 This PR makes sure that only `test` sites would be deleted. It also includes some refactorings, styling changes, and `User Account Requirements` doc section


brbrr:
Tests are passing so I'll merge it to prevent further issues: https://circleci.com/gh/Automattic/wp-e2e-tests/18614

PR #1117, opened 2018-04-05
Add test for Jetpack connection from WooCommerce wizard

bsessions85:
Added 1 test to verify connection from the WooCommerce wizard works.


brbrr:
The code looks good, but I noticed that CirceCI desktop test passed on 3d attempt so I tried to launch it locally with mocha, and the test is failed on connection step:

wordpress com 2018-04-06 08-46-57

That looks like a jetpack issue and there nothing you can do from e2e tests perspective but I'll expect that test will be flacky if the issue would not be resolved.

specs-jetpack-calypso/wp-jetpack-connect-spec.js
```diff
@@ -460,4 +466,61 @@ test.describe( `Jetpack Connect: (${ screenSize })`, function() {
 			} );
 		}
 	);
+
+	test.describe.only( 'Connect From WooCommerce plugin when Jetpack is not installed: @parallel @jetpack', function() {
```
brbrr: `.only` is not needed here :)

alisterscott:
Thanks for this - looks good

PR #1118, opened 2018-04-06
Dismiss NPS survey after logging in

rachelmcr:
This PR adds a step to the login flow to dismiss the NPS survey if it appears. This survey is being shown to all eligible users as of https://github.com/Automattic/wp-calypso/pull/23894.

Adding it to the main login flow feels a bit heavy handed, but given that it appears immediately after logging in (when eligible) this ensures that the tests can immediately move on to any possible next step.

I'm not sure why I ended up with all of these package-lock changes — they seem to be reverting the changes from https://github.com/Automattic/wp-e2e-tests/pull/1112.

### DIFF
--- a/lib/flows/login-flow.js
+++ b/lib/flows/login-flow.js
@@ -51,7 +51,8 @@ export default class LoginFlow {
 
 	login( { jetpackSSO = false, jetpackDIRECT = false, screenshot = false } = {}, eyes ) {
 		let loginURL = this.account.loginURL,
-			loginPage;
+			loginPage,
+			readerPage;
 
 		if (
 			( host === 'CI' || host === 'JN' ) &&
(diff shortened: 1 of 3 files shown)


alisterscott:
Looks like a good change to dismiss it. When someone is adding something like this what do you think about asking them to exclude showing it for e2e users? (by checking the user agent like we do for ab test overrides)

rachelmcr:
>When someone is adding something like this what do you think about asking them to exclude showing it for e2e users?

That's a good point. I thought about whether e2e users should be excluded from this, but also thought that interacting with it would ensure it could be dismissed/doesn't reappear unexpectedly (i.e. shortly after being dismissed). As I think about it now, though, I wonder if our e2e users are skewing the survey results by always dismissing it.

@taggon Do you think we should avoid showing the NPS survey to the test users we use for e2e testing? (Right now we're dismissing the survey if it appears.)

taggon:
> Do you think we should avoid showing the NPS survey to the test users we use for e2e testing?

@rachelmcr I think so. It would be better to not show the survey since the dismissals in the test process will be recorded as "a user submitted the survey without score" which affects the stats.

PR #1119, opened 2018-04-06
Fix signup magic link test flakiness

brbrr:
Make sure we wait until magic link email arrives. Also, some sign-up spec refactoring

### DIFF
--- a/lib/email-client.js
+++ b/lib/email-client.js
@@ -36,42 +36,57 @@ export default class EmailClient {
 		return d.promise;
 	}
 
-	getEmailsByRecipient( emailAddress ) {
-		var d = webdriver.promise.defer();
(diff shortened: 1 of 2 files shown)


alisterscott:
Thanks for this 👍

alisterscott:
@brbrr: I don't understand how we introduced the validator but didn't add it to all the email client calls - how does this work?

brbrr:
> didn't add it to all the email client calls 

That is a bit of magic stuff(not obvious at all). This chunk of code sets default validator if no custom validator was passed (was moved into default variable in further commits):
https://github.com/Automattic/wp-e2e-tests/blob/c79ba0fa4231ce2181a564a4e3679e1810e3e222/lib/email-client.js#L49-L51 

If custom validator is not defined - we use this one, which basically makes validation check to pass.

PR #1123, opened 2018-04-06
Try/deprecate promise

Stojdza:
Introducing async/await in wp-e2e-tests

PR #1124, opened 2018-04-09
Update: Babel7 and preset-env

sirreal:
Use `babel-preset-env`. Rather than manually curate the transforms and plugins we need, let babel determine based on some parameters, in this case `"node": "8.11.0"`. We don't expect to run tests on any other platforms so we can do the minimal set of transforms necessary.

Update to Babel 7. This was sort of a side-effect of using babel-preset-env and happened as I tried to power through to something working.

See #1123 and p6fDka-w1-p2 (#comment-1282) for motivation (cc: @Stojdza)

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,10 +1,9 @@
 {
 	"presets": [
-		"es2015",
-		"stage-1"
-	],
-	"plugins": [
-		"add-module-exports",
-		"syntax-export-extensions"
+		["@babel/preset-env", {
+			"targets": {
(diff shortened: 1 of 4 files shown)


sirreal:
This appears to work fine, but I lack the deep familiarity with this repo necessary to have a lot of confidence. I'd love to have more eyes on this.

rachelmcr:
As a heads up, Flow Patrol is on support rotation this week so there will be a delay in any non-emergency code reviews from the team.

PR #1129, opened 2018-04-11
Add support for domain registration redesign

jsnmoon:
The existing test fails because it expects the first domain suggestion to be the WordPress.com subdomain suggestion. With the domain registration redesign resulting from the Kracken project, this is no longer the case.

I've corrected/refined the WP.com subdomain suggestion in this PR.


jsnmoon:
Closing this in favor of #1130.

PR #1130, opened 2018-04-12
Add support for domain registration redesign

jsnmoon:
The existing test fails because it expects the first domain suggestion to be the WordPress.com subdomain suggestion. With the domain registration redesign resulting from the Kracken project, this is no longer the case.

I've corrected/refined the WP.com subdomain suggestion in this PR.

### DIFF
--- a/lib/components/find-a-domain-component.js
+++ b/lib/components/find-a-domain-component.js
@@ -59,7 +59,7 @@ export default class FindADomainComponent extends BaseContainer {
 	}
 
 	freeBlogAddress() {
-		const freeBlogAddressSelector = By.css( '.domain-suggestion__content h3' );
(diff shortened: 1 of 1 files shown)


alisterscott:
Thanks for the PR @jsnmoon 

Is this dependent on a change in Calypso not yet implemented? I tried this using the current version of Calypso and it matches all the domain result elements, so it just selects the first one, which is what we previously did. Should it only select the free result?

jsnmoon:
Hi @alisterscott, that's correct - this PR corresponds to [this PR from wp-calypso](https://github.com/Automattic/wp-calypso/pull/24137). Currently, we have this feature enabled on the horizon environment; I'm hoping to enable this feature in `stage` and `wpcalypso`.

alisterscott:
> @rachelmcr informed me that the canaries will use a corresponding feature branch in wp-calypso if I used the same branch name in wp-e2e-tests.

Yes, that's correct 👍

Issue #1126, opened 2018-04-10
Touch devices: incorrectly reporting failed test


keoshi:
The changes in https://github.com/Automattic/wp-calypso/pull/23979 introduced a `.touch` selector which only targets devices that support touch.

Tests are failing because they are executed on a device that doesn't support touch (judging by the scroll bars on the right side), so the testing conditions are not properly emulated.

![failed-en-mobile-can-then-see-the-plans-page-and-select-the-premium-plan--1523354369235](https://user-images.githubusercontent.com/390760/38552243-726943e8-3cb3-11e8-8cee-6e9ed957a2b9.png)

keoshi:
@alisterscott I see you've got more details on https://github.com/Automattic/wp-calypso/pull/23979#issuecomment-381511949

`.touch` should be detected with Modernizr, yeah: https://github.com/Automattic/wp-calypso/blob/c47764a96b9ec4d867fe0ba042e931f9b5d0fa19/client/lib/touch-detect/index.js

andfinally:
With the change in https://github.com/Automattic/wp-calypso/pull/24197 the plans scroll on non-touch devices again, so we can probably close this issue.

PR #1131, opened 2018-04-16
Remove sitegoalsshuffle and use vertical plans (for mobile)

alisterscott:
Includes

1. Removes sitesgoalshuffle override as this isn't in use any more
2. Sets the vertical plan behaviour due to issues with horizontal scroll raised in #1126 
3. Updates the plan selectors for using vertical plans
4. Temporarily disables the business sign up test #1132


alisterscott:
@Automattic/flowpatrol I will merge this to get tests to pass

PR #1146, opened 2018-04-18
Fix failing connect specs due to Remote installs feature

brbrr:


alisterscott:
What is the purpose of this change and what is it doing?

brbrr:
These connect specs were using JN site without Jetpack preinstalled, and since Remote Installs goes live - these tests get broken. This PR changing these tests to match current flow

alisterscott:
I don't really understand these Jetpack.com flows - it seems like they're a duplicate of using Calypso to start the connection and causing us to spin up too many jurassic.ninja instances

brbrr:
2 of Jetpack.com tests were converted to pre-connect (initial part before connecting);
Woo connect spec updated not ot override AB tests - this causes "Site is already connected" error.

PR #1150, opened 2018-04-20
Add a simple activity log verification to publishing a post

alisterscott:
This adds a check after publishing a public post that the post title is shown in the activity log

The Pressable Jetpack site doesn't work at present -> need to investigate why


alisterscott:
Hi @Stojdza 

Would you mind taking a look at this, as I've tried to include the async/await stuff in this new spec so that it will merge easier with the other specs you've already updated

Thanks

alisterscott:
It looks like this won't work for our Jetpack site as we're on the free Jetpack plan 🤔

PR #1153, opened 2018-04-22
Remove promise from shopping-cart-widget-component.js and add async/await

Stojdza:
@alisterscott This relates to the comments regarding shopping-cart-widget-component.js in PR #1123 

Without promise or async/await tests will [pass](https://circleci.com/gh/Automattic/wp-e2e-tests/19137?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), but items from shopping cart won't be removed (specs/wp-manage-domains-spec.js). 

After adding async/await to component, tests are still [passing](https://circleci.com/gh/Automattic/wp-e2e-tests/19138?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) and shopping cart is empty.

### DIFF
--- a/lib/components/shopping-cart-widget-component.js
+++ b/lib/components/shopping-cart-widget-component.js
@@ -7,24 +7,15 @@ import BaseContainer from '../base-container.js';
 
 export default class ShoppingCartWidgetComponent extends BaseContainer {
 	constructor( driver ) {
-		super( driver, by.className( 'cart-toggle-button' ) );
+		super( driver, by.css( '.cart-toggle-button' ) );
 	}
 
 	async open() {
-		const driver = this.driver;
-		const explicitWait = this.explicitWaitMS;
-
-		if ( ! await this.displayedOpen() ) {
-			return await driverHelper.clickWhenClickable(
-				driver,
-				by.css( '.cart-toggle-button' ),
-				explicitWait
-			);
-		}
-	}
-
-	async displayedOpen() {
-		return await driverHelper.isElementPresent( this.driver, by.css( 'cart-body' ) );
+		return await driverHelper.clickWhenClickable(
+			this.driver,
+			by.css( '.cart-toggle-button' ),
+			this.explicitWaitMS
+		);
 	}
 
 	async removeItem( self ) {
--- a/lib/components/shopping-cart-widget-component.js
+++ b/lib/components/shopping-cart-widget-component.js
@@ -10,12 +10,12 @@ export default class ShoppingCartWidgetComponent extends BaseContainer {
 		super( driver, by.className( 'cart-toggle-button' ) );
 	}
 
-	open() {
+	async open() {
 		const driver = this.driver;
 		const explicitWait = this.explicitWaitMS;
 
-		if ( ! this.displayedOpen() ) {
-			return driverHelper.clickWhenClickable(
+		if ( ! await this.displayedOpen() ) {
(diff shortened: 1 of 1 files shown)


lib/components/shopping-cart-widget-component.js
```diff
@@ -10,96 +10,41 @@ export default class ShoppingCartWidgetComponent extends BaseContainer {
 		super( driver, by.className( 'cart-toggle-button' ) );
```
alisterscott: Can you update this to a css selector whilst your here? (It's an old selector added before we try to use CSS selectors exclusively)

lib/components/shopping-cart-widget-component.js
```diff
@@ -10,96 +10,41 @@ export default class ShoppingCartWidgetComponent extends BaseContainer {
...
-		);
-
-		return d.promise;
+		const explicitWait = this.explicitWaitMS;
+
+		if ( ! await this.displayedOpen() ) {
+			return await driverHelper.clickWhenClickable(
+				driver,
+				by.css( '.cart-toggle-button' ),
+				explicitWait
```
alisterscott: this explicit wait isn't actually necessary as it's the same as the default value - would you remind removing it? (added before we defaulted the value I believe)

lib/components/shopping-cart-widget-component.js
```diff
@@ -10,96 +10,41 @@ export default class ShoppingCartWidgetComponent extends BaseContainer {
...
-				}
-			},
-			function() {
-				d.reject( 'Cart not open' );
-			}
-		);
-
-		return d.promise;
+	async displayedOpen() {
+		return await driverHelper.isElementPresent( this.driver, by.css( 'cart-body' ) );
```
alisterscott: It's great to have one line, but that selector doesn't look like it should have ever worked as it's not a CSS class selector and there's no cart-body element in the DOM. Would you mind updating this to a proper CSS class selector - you could even re-use the selector from the constructor if you wanted to

alisterscott:
This is looking promising - if we can get rid of promises like this in pages/components it should make it easier for the tests to convert over to call these components

lib/components/shopping-cart-widget-component.js
Stojdza: Done.

lib/components/shopping-cart-widget-component.js
Stojdza: Done. Did the same for driver.

lib/components/shopping-cart-widget-component.js
Stojdza: I removed displayedOpen(), and in open() method only click to open shopping cart (since in this flow cart is always closed and there is no need to check is it open, I removed check is it open).

alisterscott: Sorry, I'm not saying what you did is wrong, I just noticed that `by.css( 'cart-body' )` would never had worked I don't think 🤔

alisterscott: I think it should be `by.css( '.cart-body' )`

alisterscott: Oh, I misread it again, I think that makes sense as we shouldn't have to worry about whether it's open 😀

Stojdza: Yes, I understood. But it crossed my mind this would be more optimal solution? It will work both ways.

PR #1159, opened 2018-04-24
Fix Free Plan selection during Signup

ockham:
Companion PR to https://github.com/Automattic/wp-calypso/pull/24333, which broken tests that look for the 'Start with Free [Plan]' button based on the previous class name.


ockham:
Thanks for reviewing @rachelmcr!

PR #1154, opened 2018-04-22
Add async/await into /specs

Stojdza:
Specs affected with this PR:

- wp-log-in-out-spec.js
- wp-manage-domains-spec.js
- wp-notifications-spec.js
- wp-plan-purchase-spec.js
- wp-reader-spec.js
- wp-theme-multisite-spec.js
- wp-theme-switch-spec.js
- wp-view-site-spec.js
- wp-signup-spec.js

PR #1162, opened 2018-04-24
Add async/await into /specs/wp-invite-users-spec.js

Stojdza:
This is continuation of #1154 and adding async/await to e2e tests.


Stojdza:
Add async/await for newly added tests - #1134

PR #1163, opened 2018-04-25
Revoke invites test

ghost:
Adding a test that aims to test that pending invites can be revoked successfully


lib/pages/invite-error-page.js
```diff
@@ -0,0 +1,18 @@
+/** @format */
+
+import By from 'selenium-webdriver';
```
Stojdza: It looks like this import should be changed to:
`import { By } from 'selenium-webdriver';`

Stojdza (changes requested):
async/await calls look fine. 

Looking at this error message in CircleCI - `_seleniumWebdriver.default.css is not a function` - it looks like that `import` _in invite-error-page.js_ should be changed.

lib/pages/invite-error-page.js
sofiavistas: indeed it was this, thanks for the review!

specs/wp-invite-users-spec.js
```diff
@@ -286,6 +271,128 @@ testDescribe( `[${ host }] Invites:  (${ screenSize })`, function() {
 		} );
 	} );
 
+	test.describe.only( 'Inviting New User as an Editor And Revoke Invite: @parallel @jetpack', function() {
```
Stojdza: @sofiavistas Does `.only` should stay here? I do not know will CircleCI skip the other tests from this spec. Maybe somebody can confirm this?

bsessions85: Good catch @Stojdza, yes, CircleCI will skip the other tests, so this should be removed

Stojdza (approved):

alisterscott:
Thanks for the PR

LGTM so merging now before this gets out of date

PR #1164, opened 2018-04-25
Add async/await into /specs/wp-post-editor-spec.js

Stojdza:
Moving forward with async/await usage. 

To properly use _await_, functions should return promise object. Since some of them didn't return anything, I added missing returns - 091d7ae


alisterscott:
This looks good @Stojdza !

Out of interest, why do not `test.describe` blocks need to be updated to specify async? Is there a reason for this?

Stojdza:
@alisterscott Here is a one good explanation - https://github.com/mochajs/mocha/issues/2975#issuecomment-326162912
Anyway, I tried to run tests with `async`  in `describe` blocks, just to see what will happen,  and all tests have failed. Enough for me :)

PR #1166, opened 2018-04-26
Remove NPS survey from e2e tests

rachelmcr:
This removes the NPS survey steps added in https://github.com/Automattic/wp-e2e-tests/pull/1118. It can be merged after https://github.com/Automattic/wp-calypso/pull/24465, which will prevent the NPS survey from being shown during e2e tests.


alisterscott:
Thanks for this, it looks fine but I'm not sure why it's already passing when that other PR isn't merged yet 🤔

rachelmcr:
It's passing because previous test runs dismissed the NPS survey for our e2e test users, so even though Calypso still has the potential to show the NPS survey in our e2e tests (until the other PR is merged) the survey won't show up right away again for those users.

PR #1168, opened 2018-04-26
Add async/await into /wp-media-upload-spec.js and /wp-page-editor-spec.js

Stojdza:
Asyncify _wp-media-upload-spec.js_ and _wp-page-editor-spec.js_.


Stojdza:
I merged these two specs into one PR since both are using same background functions (updated with `return`).

PR #1169, opened 2018-04-30
Update a signup test to use a specific currency

### LINKED ISSUE #401: Create multi-currency signup tests
When a solution to https://github.com/Automattic/wp-calypso/issues/6550 has been implemented, we need to create signup tests in non-USD currencies.

alisterscott:
Fixes #401

It's now possible to set currency using a cookie
This does that for one of our existing sign up tests - and also cleans up the formatting and adds a style guide entry for this

### DIFF
--- a/docs/style-guide.md
+++ b/docs/style-guide.md
@@ -80,3 +80,33 @@ instead of:
 new StartPage( driver, true, 'en', '', true, '' ).displayed();
 ```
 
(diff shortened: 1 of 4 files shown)


alisterscott:
Found this bug during creation of this: https://github.com/Automattic/wp-calypso/issues/24559

PR #1176, opened 2018-05-02
Update theme sidebar selector

alisterscott:
This is for changes in https://github.com/Automattic/wp-calypso/pull/24587

This should work regardless of the order of the buttons now


alisterscott:
@Automattic/flowpatrol merging this one as all the theme tests are failing

PR #1177, opened 2018-05-02
Add product validation to sign up specs

alisterscott:
- Adds validation to cart contents on signup to make sure correct products are present
- Adds a variety of currencies (USD, GBP, EUR, CAD, AUD & JPY) to existing sign up specs
- Adds a new scenario for a personal plan signup which was missing from coverage

### DIFF
--- a/specs/wp-signup-spec.js
+++ b/specs/wp-signup-spec.js
@@ -407,10 +407,8 @@ test.describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function()
 			);
 
 			test.it(
-				'Can then see the secure payment page, and can enter and submit test payment details',
+				'Can then see the secure payment page with the expected currency in the cart',
 				async function() {
-					const testCreditCardDetails = dataHelper.getTestCreditCardDetails();
-
 					const securePaymentComponent = new SecurePaymentComponent( driver );
 					if ( driverManager.currentScreenSize() === 'desktop' ) {
 						const totalShown = await securePaymentComponent.cartTotalDisplayed();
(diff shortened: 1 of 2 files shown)


specs/wp-signup-spec.js
```diff
@@ -49,6 +49,11 @@ const screenSize = driverManager.currentScreenSize();
 const signupInboxId = config.get( 'signupInboxId' );
 const host = dataHelper.getJetpackHost();
 const locale = driverManager.currentLocale();
+const personalPlanSlug = 'personal-bundle';
+const premiumPlanSlug = 'value_bundle';
+const businessPlanSlug = 'business-bundle';
+const dotLiveDomainSlug = 'dotlive_domain';
+const privateWhoisSlug = 'private_whois';
```
brbrr: I think it would better feet into `SecurePaymentComponent` - the same approach as we do with Plans page

specs/wp-signup-spec.js
```diff
@@ -183,14 +188,16 @@ test.describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function()
...
 	test.describe(
-		'Sign up for a site on a premium paid plan through main flow @parallel @visdiff',
+		'Sign up for a site on a premium paid plan through main flow in USD currency @parallel @visdiff',
 		function() {
 			this.bailSuite( true );
 
 			const blogName = dataHelper.getNewBlogName();
 			const expectedBlogAddresses = dataHelper.getExpectedFreeAddresses( blogName );
 			const emailAddress = dataHelper.getEmailAddress( blogName, signupInboxId );
 			const password = config.get( 'passwordForNewTestSignUps' );
```
brbrr: I see this line appears 5 times in different specs. Maybe it worth to extract it to global scope? (defined constants right after imports)

specs/wp-signup-spec.js
```diff
@@ -409,42 +446,198 @@ test.describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function()
...
+			const password = config.get( 'passwordForNewTestSignUps' );
+			const currencyValue = 'GBP';
+			const expectedCurrencySymbol = '£';
+
+			test.before( async function() {
+				return await driverManager.ensureNotLoggedIn( driver );
+			} );
+
+			test.it( 'We can set the sandbox cookie for payments', async function() {
+				const sandboxCookieValue = config.get( 'storeSandboxCookieValue' );
```
brbrr: Same as the password variable. It could be extracted to the global scope.

specs/wp-signup-spec.js
alisterscott: done 🐌

specs/wp-signup-spec.js
alisterscott: done 🐌

specs/wp-signup-spec.js
alisterscott: done 🐌

alisterscott:
Thanks for your feedback @brbrr - fixed issues and merging

brbrr:
Thanks for your fixes!

Issue #1185, opened 2018-05-04
Workaround: clicking role twice on edit team member page


alisterscott:
We currently click role twice on editing a team member page
We should investigate why this is necessary

alisterscott:
This is being fixed in https://github.com/Automattic/wp-e2e-tests/pull/1263

PR #1186, opened 2018-05-04
Fix failing invite tests

alisterscott:
This should fix the failing invite tests

A few issues here:
1. Invites were being revoked by being first on list of pending which doesn't work as we use shared accounts and parallel runs - now revokes specifically by email address
2. A workaround for #1185 - I really don't know why that is necessary but don't have hours to investigate right now


alisterscott:
@Automattic/flowpatrol I'm going to merge this one so we're green over the weekend

PR #1187, opened 2018-05-04
Fix Jetpack Onboarding/Woo tests flakiness

brbrr:
Some of the Onboarding / Woo tests are failing from time to time (timing issues). This PR adds some delays for these failing tests.


alisterscott:
One thing I noticed is that every Jetpack On Boarding page is that the expected element is the same:

`super( driver, By.css( '.jetpack-onboarding' ) );`

so it'll never wait for a new page to load. 

The idea behind the expected element is to wait for a new page with the new element so we know we're exactly where we're meant to be - it may be possible to add some unique elements to each Jetpack OB page which may remove the need to have hard coded waits - just an idea

PR #1188, opened 2018-05-04
Add mothersDay2018Discount_20180411 test

adamziel:
Corresponding PR: https://github.com/Automattic/wp-calypso/pull/24688

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -66,7 +66,8 @@
     "springSale30PercentOff",
     "showMoneyBackGuarantee",
     "multiyearSubscriptions",
-    "signupDomainStrikethruPrice"
+    "signupDomainStrikethruPrice",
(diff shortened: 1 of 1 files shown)


config/default.json
```diff
@@ -66,7 +66,8 @@
     "springSale30PercentOff",
     "showMoneyBackGuarantee",
     "multiyearSubscriptions",
-    "signupDomainStrikethruPrice"
+    "signupDomainStrikethruPrice",
+    "mothersDay2018Discount_20180411"
```
brbrr: Are you sure that date is required here? It looks like date should be present in `overrideABTests` only

brbrr:
@adamziel should we merge this?

alisterscott:
I can't see this in active-tests so I will close this as it isn't necessary

PR #1189, opened 2018-05-07
Remove nesting in Invites tests

brbrr:
This PR removes nesting from invites tests


brbrr:
I have enabled the Contributor test, and removed that `testDescribe` thing.

PR #1194, opened 2018-05-08
Remove unused promises

Stojdza:
For easier maintenance remove unused functions which return promises.


lib/email-client.js
```diff
@@ -12,30 +12,6 @@ export default class EmailClient {
 		this.mailbox = new Mailosaur.Mailbox( mailboxId );
 	}
 
-	deleteAllEmailByID( emailID ) {
```
brbrr: This method is used here https://github.com/Automattic/wp-e2e-tests/blob/master/specs/wp-log-in-out-spec.js#L248

lib/pages/devdocs-page.js
```diff
@@ -109,79 +86,4 @@ export default class DevdocsPage extends BaseContainer {
 			'document.querySelector( ".environment-badge" ).style.visibility = "hidden";'
 		);
 	}
-
-	screenshotAllElements( eyes ) {
```
brbrr: This method is used here: https://github.com/Automattic/wp-e2e-tests/blob/master/specs-visdiff/wp-devdocs-visdiff.js.removed#L46

@alisterscott do know why do we keeping these `.removed` viss diff tests?

lib/push-client.js
```diff
@@ -1,38 +1,6 @@
...
 import { listen } from 'push-receiver';
-import request from 'request-promise';
-
-/**
- * Marks a push token as approved on WPCOM API
- * @param {String} pushToken push token that recieved via a push notification on an app
- * @param {String} bearerToken user bearer token to be used with the API
- * @returns {Promise} a request result promise
- */
-export const approvePushToken = ( pushToken, bearerToken ) =>
```
brbrr: This method is used here: https://github.com/Automattic/wp-e2e-tests/blob/master/specs/wp-log-in-out-spec.js#L167

lib/email-client.js
Stojdza: Also here:
https://github.com/Automattic/wp-e2e-tests/blob/master/specs/wp-log-in-out-spec.js#L323
https://github.com/Automattic/wp-e2e-tests/blob/master/specs/wp-log-in-out-spec.js#L390

lib/pages/devdocs-page.js
Stojdza: Also here: https://github.com/Automattic/wp-e2e-tests/blob/master/specs-visdiff/wp-devdocs-visdiff.js.removed#L63

lib/email-client.js
Stojdza: Also here:
https://github.com/Automattic/wp-e2e-tests/blob/master/specs/wp-log-in-out-spec.js#L323
https://github.com/Automattic/wp-e2e-tests/blob/master/specs/wp-log-in-out-spec.js#L390

Stojdza:
I put back functions which are used in temporary disabled tests and disable them. 

@alisterscott What about `.removed` files? Are those permanently deprecated? If yes, I think that functions used there can be safely removed.

alisterscott:
> What about .removed files? Are those permanently deprecated? If yes, I think that functions used there can be safely removed.

Yes, let's remove those files - we can add them back if we need to (and do promise stuff then)

PR #1197, opened 2018-05-09
Fix IE11 temporarily

alisterscott:
Emergency fix to get saucelabs to work


alisterscott:
@Automattic/flowpatrol merging this as saucelabs has DNS problems with circleCI https://status.saucelabs.com/incidents/tm272kz6yb5t

PR #1203, opened 2018-05-14
Remove Promise from email client

Stojdza:
Remove Promise and add async/await instead.


alisterscott:
Looks good, when the email is not found the error is:

```
 TypeError: emails is not iterable
    at Context.<anonymous> (specs/wp-signup-spec.js:167:24)
```

It might be better to just add something like:

`throw new Error( `Could not locate email for '${ emailAddress }' in ${ emailWaitMS }ms` );`

as the final line in `async pollEmailsByRecipient( emailAddress, validator ) {`

That we should see something like:

>  Error: Could not locate email for 'a11ne2eflowtesting1526358643480976.6efd1b57@mailosaur.io' in 60000ms

alisterscott:
> While we're at it, should we look into Mailosaur's latest API changes that would allow us to completely remove while loop and retries? @alisterscott

Yes, we should do this, but since this is working I say merge this first, then we can work switching to the new mailosaur API separately

lib/email-client.js
```diff
@@ -42,22 +42,29 @@ export default class EmailClient {
...
 			if ( validator( emails ) ) {
 				return emails;
 			}
+			await this.resolveAfterTimeout( intervalMS );
+			retries--;
 		}
+		throw new Error( `Could not locate email for '${ emailAddress }' in '${ emailWaitMS }'ms` );
+	}
+
+	resolveAfterTimeout( timeout ) {
```
Stojdza: I put this instead of `setInterval()` and it should work with disabled promise manager.

alisterscott:
Tested with a timeout and worked well - cheers 👍

PR #1204, opened 2018-05-18
Add payment button front end verification

### LINKED ISSUE #1030: Simple Payments: add a purchase test
Add a test to verify that payments may go through on a few core cases for existing products:
- USD with a decimal value eg `1.99`
- USD with whole units `20`
- Spot checks with other currencies like JPY

We may need to consult the Flow team on how best to structure this.

Related:
https://github.com/Automattic/wp-calypso/issues/22939
D10823-code
p58i-6nm-p2 for manual test instructions

cc @Automattic/lannister @retrofox

alisterscott:
Due to recent payment button issues, this enhances the payment button tests to be more comprehensive:

1. Adds a payment button front-end check that makes sure if shows and functions (opens paypal with the correct amount)
2. Uses two different currencies
3. Adds a payment button to a page as well as a post

Fixes #1030

### DIFF
--- a/lib/components/payment-button-front-end-component.js
+++ b/lib/components/payment-button-front-end-component.js
@@ -10,7 +10,7 @@ export default class PaymentButtonFrontEndComponent extends BaseContainer {
 		super( driver, by.css( '.jetpack-simple-payments-wrapper' ) );
 	}
 
-	clickPaymentButton() {
+	async clickPaymentButton() {
 		const payPalPayButtonSelector = by.css( '.paypal-button' );
 		this.driver.wait(
 			until.ableToSwitchToFrame( by.css( '.jetpack-simple-payments-button iframe' ) ),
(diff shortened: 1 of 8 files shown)


alisterscott:
@Stojdza would you mind checking this one out? I'm not sure whether the pages and components need async methods or whether it's just the actual specs that need to use await. I'm not adding specific promises.

lib/components/payment-button-front-end-component.js
```diff
@@ -0,0 +1,24 @@
...
+
+export default class PaymentButtonFrontEndComponent extends BaseContainer {
+	constructor( driver ) {
+		super( driver, by.css( '.jetpack-simple-payments-wrapper' ) );
+	}
+
+	async clickPaymentButton() {
+		const payPalPayButtonSelector = by.css( '.paypal-button' );
+		this.driver.wait(
+			until.ableToSwitchToFrame( by.css( '.jetpack-simple-payments-button iframe' ) ),
```
Stojdza: I propose to use `.xcomponent-component-frame` selector instead `.jetpack-simple-payments-button iframe` and it should be fine.

alisterscott:
Thanks for your review @Stojdza and the iFrame suggestion which worked a treat

Merging now as is - I'll leave the async stuff in there for when the /lib folder is converted also

PR #1209, opened 2018-05-21
Try waiting for username field

alisterscott:
I noticed in this build: https://circleci.com/gh/Automattic/wp-e2e-tests/19843 there are sometimes issues setting the username. This waits until it is present to set it

### DIFF
--- a/lib/pages/login-page.js
+++ b/lib/pages/login-page.js
@@ -22,15 +22,15 @@ export default class LoginPage extends BaseContainer {
 		driverHelper.waitTillNotPresent( driver, By.css( 'input[disabled]#usernameOrEmail' ) );
 	}
 
-	async login( username, password ) {
+	login( username, password ) {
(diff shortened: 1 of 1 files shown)


alisterscott:
Thanks @Stojdza 

> Have you tried to make this function async and add await before driverHelper.setWhenSettable( driver, userNameSelector, username );?

I just tried but had unexpected failures, so I'll do that in a separate PR

Stojdza:
@alisterscott I've done it [here](https://github.com/Automattic/wp-e2e-tests/blob/no-manager/lib/pages/login-page.js#L25) as a part of adding async functions to _/pages_. I was curious does we need _waitTillPresentAndDisplayed()_ if function is async and I will check it out after merging this change.

alisterscott:
> I've done it here as a part of adding async functions to /pages.

Thanks, happy to see how it goes

PR #1210, opened 2018-05-21
Try increasing workers to 6

alisterscott:
Move from 3 to 6 - should reduce execution time down substantially

### DIFF
--- a/magellan-canary.json
+++ b/magellan-canary.json
@@ -4,7 +4,7 @@
   ],
   "mocha_opts": "test/mocha.opts",
   "mocha_args": "-R spec-xunit-reporter",
-  "max_test_attempts": 2,
+  "max_test_attempts": 3,
   "max_workers": 3,
   "suiteTag": "canary",
   "sauce": false,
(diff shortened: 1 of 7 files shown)


alisterscott:
@hoverduck I can't remember why we choose 3 workers - was there a reason? It _seems_ to run fine with 6 on CircleCI, but I guess only time would tell

hoverduck:
I don't recall how many workers I tried, but I definitely saw stability issues with higher numbers and settled on 3 as a conservative point that was proven to work well.  And at the time it was still a huge performance gain for us, so it seemed a good choice.

We were also on the custom Docker image rather than the CircleCI-provided one, maybe theirs is more efficient.  And it's been a year since we turned it on, so who knows what other infrastructure improvements have been made on their end that could've improved things.

So no, there's no reason not to try bumping it up to 6+ and see how it goes :smile:

alisterscott:
@Automattic/flowpatrol a heads-up I'll merge this and see how things go - in case we start seeing failures because of too many workers we can just quickly revert this

PR #1216, opened 2018-05-23
Replace deprecated functions with #setTimeouts

### LINKED ISSUE #1207: Update how webdriverjs sets timeouts
Our current approach:

driver.manage().timeouts().implicitlyWait( webDriverImplicitTimeOutMS );
driver.manage().timeouts().pageLoadTimeout( webDriverPageLoadTimeOutMS );

is showing as deprecated - work out how to use capabilities to do this same thing

brbrr:
Fixes #1207

### DIFF
--- a/lib/driver-manager.js
+++ b/lib/driver-manager.js
@@ -186,12 +186,7 @@ export function startBrowser( { useCustomUA = true, resizeBrowserWindow = true }
 	}
 	driver
 		.manage()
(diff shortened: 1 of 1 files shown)


alisterscott:
This looks great - thanks @brbrr 

I noticed there's also a script timeout value - I wonder whether we should also set that? Maybe in a separate PR...

PR #1218, opened 2018-05-23
Refactor #removeSites to not to exceed test timeouts

brbrr:
Jetpack disconnection flow sometimes failing due to exceeded timeouts. This PR adds check to prevent these failures.

Now, `#removeSites` will stop deleting sites if time spent will exceed 80% of Mocha timeout.


alisterscott:
Neat solution - tested and works well ✅

PR #1221, opened 2018-05-23
Make payment button tests to work with Jetpack

brbrr:
Page/Post payment button tests aren't running against Jetpack targets. This PR fixes that.

**Config file is updated**


alisterscott:
Thanks - this looks great

Tested against both pressable and pressable bleeding edge - worked well 👍

Issue #1017, opened 2018-03-08
Update Mailosaur to 5.0.2


brbrr:
There is a new major version had released: https://github.com/mailosaur/mailosaur-node

The API has changed, so complete re-write of `EmailClient` would be required. 

Benefits are unknown 🙃

alisterscott:
@abidhahmed [points out](https://github.com/Automattic/wp-e2e-tests/pull/1203#pullrequestreview-119796574) we get:

> https://mailosaur.com/docs/api/#operation/Messages_WaitFor

for free in upgrading to this

PR #1223, opened 2018-05-24
Update email client with latest mailosaur api

abidhahmed:
Updated with latest Mailosaur API 5.0.3 (#1017) 

It therefore replaces `pollForEmailsByReceipient` with `waitForEmailByReceipient` for native wait ability and requires no need for timeout as Mailosaur returns a `204` if there's no email in <10s.

This potentially also fixes the race condition wherein emails were rejected by `validator` allowing them to exit polling on the first try.

Also added format check to reject potential emails that do not have any `html` formatting (_which is required for all emails from us anyway_).

### DIFF
--- a/lib/email-client.js
+++ b/lib/email-client.js
@@ -3,13 +3,13 @@
 // import webdriver from 'selenium-webdriver';
 import config from 'config';
 
-const emailWaitMS = config.get( 'emailWaitMS' );
+// const emailWaitMS = config.get( 'emailWaitMS' );
 
 export default class EmailClient {
 	constructor( mailboxId ) {
-		const apiKey = config.get( 'mailosaurAPIKey' );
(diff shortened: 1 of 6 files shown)


brbrr:
The code looks good, although I'm not sure that new implementation is equivalent to old one. For example, if there 2 emails would be present in a mailbox. Which email will be returned? 

The old approach returns all the emails, and afterward, we can search for what we need. New approach just returns first discovered email.

abidhahmed:
@brbrr - This waits for an email to arrive in the mailbox _realtime_. It does not allow us to do any sorting whatsoever as it only waits for a new email, it doesn't return the entire mailbox as a result.

I checked all the tests we have, and all check for the email _immediately_ after requesting for one, so it kinda works out OK. If we add any further tests that need to sort from multiple emails after sometime, we'll need to update this to add listing all emails as a separate function.

alisterscott:
Reverted in #1242 due to potential issues with Mailosaur

PR #1227, opened 2018-05-25
Try to fix login page flakiness

brbrr:
Sometimes `#login` randomly fails with `Timed out waiting for element with css selector of '#user_login, #usernameOrEmail' to NOT be present` error. Since `#login` is involved in almost every test - this error may affect entire test suite.

### DIFF
--- a/lib/pages/login-page.js
+++ b/lib/pages/login-page.js
@@ -32,17 +32,20 @@ export default class LoginPage extends BaseContainer {
 		driverHelper.waitTillPresentAndDisplayed( driver, userNameSelector );
 		driverHelper.setWhenSettable( driver, userNameSelector, username );
 
-		return driverHelper.isElementPresent( driver, wpcomLoginSelector ).then( isWpcom => {
-			if ( isWpcom ) {
-				driver.findElement( userNameSelector ).sendKeys( Key.ENTER );
-			}
+		return driverHelper
+			.isEventuallyPresentAndDisplayed( driver, wpcomLoginSelector, 1000 )
+			.then( isWpcom => {
+				if ( isWpcom ) {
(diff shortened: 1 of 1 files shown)


brbrr:
Actually this didn't help :/ https://circleci.com/gh/Automattic/wp-e2e-tests/19998

Stojdza:
@brbrr Maybe you can try with async/await, like [here](https://github.com/Automattic/wp-e2e-tests/blob/no-manager/lib/pages/login-page.js#L25)?

brbrr:
@alisterscott description updated

alisterscott:
@brbrr: I still don't think we fully know what is causing this to happen.

If we have a look at this test result: https://circleci.com/gh/Automattic/wp-e2e-tests-canary/8267

I can see the error you mention

But the screenshot shows that the username field has not even been filled in: https://8267-83742103-gh.circle-artifacts.com/0/home/circleci/wp-e2e-tests/wp-e2e-tests/screenshots/FAILED-EN-MOBILE-can-log-in-1527497452388.png

Which is unexpected and I would expect the test to fail at that point

Why do you think it's not filled in? All the lines of code before waiting for it not to be there seem to have passed so it may have been filled in at some point...

Should we simply the login method by having separate methods (or pages?) for wordpress.com and jetpack, and keep the login as simple as possible?

brbrr:
Yeah, that error is super strange. I have 2 versions:
- Selenium fills the fields correctly, but app (or selenium) revert these changes
- These buttons and input fields are present in DOM, and selenium do all the actions with not visible elements.

But we need some more data/logging to understand this problem. I think we need to add more logging and/or video recording to be able to track down such test flakiness

brbrr:
Closing this PR as it didn't fix anything

PR #1230, opened 2018-05-28
Add new Kracken domain suggestion engines test

hambai:
This is a new domain suggestion engines test we're running on NUX now.

Here's the Calypso PR: https://github.com/Automattic/wp-calypso/pull/25071


brbrr:
I think you missing date in `overrideABTests`. Don't you mind adding it?

PR #1232, opened 2018-05-28
Dissmis GDPR banner on mobile devices

brbrr:
Sometimes GDPR banner is covering some of the page elements which cause tests to fail.

To test:
run `BROWSERSIZE=mobile mocha specs -g 'Inviting new user as an Editor'`
Sometimes it fails because it miss-clicking on 'People' banner in sidebar

### DIFF
--- a/lib/components/sidebar-component.js
+++ b/lib/components/sidebar-component.js
@@ -1,10 +1,11 @@
 /** @format */
 
 import { By, until } from 'selenium-webdriver';
-import * as driverHelper from '../driver-helper.js';
 
 import BaseContainer from '../base-container.js';
 import DisconnectSurveyPage from '../pages/disconnect-survey-page.js';
+import * as driverHelper from '../driver-helper';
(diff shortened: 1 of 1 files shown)


alisterscott:
@brbrr is there any way for me to make it appear to test that this PR works?

brbrr:
@alisterscott I added test instructions to ticket description

alisterscott:
I can't get this to show, as it looks like I'm not in the right country - and it shoudn't show in CircleCI depending on which AWS region CircleCI uses.

However, I think instead of this approach which slows down every load of the sidebar by at least 1000ms, I think we should just set a cookie to not show the banner.

If you see the PR here that enables this message it includes what cookie it sets to which value, so I think we can just set that when we load the homepage or something.

What do you think @brbrr ?

brbrr:
I'm going to close this since the GDPR banner does not actually affect our tests.

PR #1240, opened 2018-06-01
Update timestamp on multiyearSubscriptions test to reflect calypso repo

adamziel:
Related to this PR: https://github.com/Automattic/wp-calypso/pull/25221


brbrr:
Thanks for the PR. Out of curiosity - why do you need to update the timestamps of the AB test?

PR #1241, opened 2018-06-02
Needed changes for disabled promise manager

Stojdza:
- Selenium promise manager is disabled
- All needed functions in `/lib` folder are asyncified so `/specs` can pass with disabled promise manager


alisterscott:
Well done on this - I think we should merge it before we start getting conflicts - then we can disable the promise manager separately to this and start fixing any individual issues we find.

A README or Style Guide for not using promises would help before we disable it :)

Stojdza:
Thanks, @alisterscott!

PR #1247, opened 2018-06-04
Try deleting new accounts as we create them

alisterscott:
This will stop emails being sent after the tests are finished


alisterscott:
🤔 that failed spectacularly on master - I will investigate

PR #1249, opened 2018-06-04
Set slackChannelForScreenshots for visdiff tests

brbrr:
⚠️ Config file is updated ⚠️ 

`slackChannelForScreenshots` wasn't set for `VISDIFF` target, so visdiff screenshots were reported into the full-suite channel.


brbrr:
merging this to check visdiff build

alisterscott:
Thanks for cleaning this up 👍

PR #1255, opened 2018-06-05
Fix failing jetpack tests

brbrr:
Bunch of Jetpack tests started to fail due to different reasons:
- `JetpackConnect` user got ~120 disconnected sites, which causes "Switch site" button did not show up right away. 
- Setting up AB tests on `JetpackAuthorizePage` & `JetpackConnectPage` causes an unexpected page reload, which breaks the tests.
- Some other tests started to fail due to async/await changes

I might fix some of the test flakiness along the way

### DIFF
--- a/lib/pages/editor-page.js
+++ b/lib/pages/editor-page.js
@@ -399,9 +399,9 @@ export default class EditorPage extends BaseContainer {
 	}
 
 	async titleShown() {
-		return await this.driver
-			.findElement( by.css( '.editor-title__input' ) )
-			.getAttribute( 'value' );
+		let titleSelector = by.css( '.editor-title__input' );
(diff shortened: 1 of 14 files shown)


alisterscott:
I had a quick look but these were still failing - I can take a look tomorrow - or else @abidhahmed may be able to review

lib/pages/editor-page.js
```diff
@@ -399,9 +399,9 @@ export default class EditorPage extends BaseContainer {
 	}
 
 	async titleShown() {
-		return await this.driver
-			.findElement( by.css( '.editor-title__input' ) )
-			.getAttribute( 'value' );
+		let titleSelector = by.css( '.editor-title__input' );
+		driverHelper.waitTillPresentAndDisplayed( this.driver, titleSelector );
```
alisterscott: does this need await?

PR #1257, opened 2018-06-06
Try/deleting new accounts (again)

alisterscott:
Second attempt
First attempt failed because I didn't wait properly for a page


alisterscott:
@bsessions85 just a heads-up I re-did this and have merged it

Issue #1243, opened 2018-06-03
Ability to disable domains tests in CircleCI


alisterscott:
Using a `@domains` tag or something in specs we should be able to "turn off" any tests that use domains when we have problems with domains providers/test domains accounts

alisterscott:
Being looked at in #1259

PR #1259, opened 2018-06-06
Skip tests if env variable is true

bsessions85:
This is one way to do this. Another option would be to use the SKIP_TEST_REGEX if we can get it to work with Magellan(#506).


alisterscott:
This fixes #1243

bsessions85:
I was sure I tested that, but sure enough, it wasn't right! Updated and verified it works with 
no variable, variable = true, and variable is != true. Thanks!

PR #1260, opened 2018-06-06
Fix IE11 canary issues

alisterscott:
These have been too short for IE11 builds which use Saucelabs

IE11 canaries are failing - this fixes them by removing the account cleanup for the single scenario


alisterscott:
Merging this as canaries are failing

PR #1258, opened 2018-06-06
Try async base container

alisterscott:
fixes #1253

Adds a `AsyncBaseContainer` which allows us to progressively change pages to support async initialisation using static async methods.

I have updated two pages to use this as an example to begin with.

PR #1261, opened 2018-06-07
Fix Jetpack login issues

alisterscott:
These were introduced in the async container PR #1258


alisterscott:
Merging these small fixes

PR #1262, opened 2018-06-08
Update sign up tests to use async pages

alisterscott:
Updates the sign up specs to use async page classes


alisterscott:
I'd like to merge this before it gets too big and then do a subsequent PR for other specs

lib/pages/themes-page.js
```diff
@@ -52,28 +46,12 @@ export default class ThemesPage extends BaseContainer {
...
-				return await driverHelper
-					.isElementPresent( self.driver, selector )
-					.then( function( present ) {
-						return ! present;
-					} );
-			},
-			this.explicitWaitMS,
-			'The themes are still loading after waiting for them'
-		);
+		return driverHelper.waitTillNotPresent( this.driver, by.css( '.themes-list .is-placeholder' ) );
```
Stojdza: We should use `await` for driverHelper methods i.e. for every method which returns promise. This is not mandatory for now since Promise manager is enabled and once we disable it we need to check all our methods once again.

bsessions85:
This looks good to me. Tests ran great locally!

lib/pages/themes-page.js
alisterscott: Thanks, missed that one

PR #1263, opened 2018-06-11
Update invite spec

### LINKED ISSUE #1185: Workaround: clicking role twice on edit team member page

alisterscott:
This updates the invite users spec to use async page objects:

- cleans up a lot of code
- removes the follower test as it doesn't work on our test site
- changes the description of the private site test to indicate this is WordPress.com only (there are no private Jetpack sites)

Also fixes #1185


Stojdza:
I'll merge this now, to avoid potential conflicts.

PR #1265, opened 2018-06-13
Async more pages

alisterscott:
This moves a bunch of new pages to use the new base class

Looks like we have done 47 pages now
With 86 pages to go


bsessions85:
I'm going to merge this so I can get the latest and update some more

PR #1266, opened 2018-06-13
Update/async pages

bsessions85:
Updated a few more pages to use the AsyncBaseContainer.

On theme-dialog-component.js I initially tried to use a _postInit function but was having issues. It could be that I don't understand how that part should work, so it might be good to have someone else take a look and see if was even necessary for the wait for the element to be located in the constructor

### DIFF
--- a/lib/components/widget-contact-info-component.js
+++ b/lib/components/widget-contact-info-component.js
@@ -9,12 +9,14 @@ export default class WidgetContactInfoComponent extends AsyncBaseContainer {
 		super( driver, By.css( '.widget_contact_info' ) );
 	}
 
-	getName() {
-		return this.driver.findElement( By.css( '.widget_contact_info h2.widget-title' ) ).getText();
+	async getName() {
+		return await this.driver
+			.findElement( By.css( '.widget_contact_info h2.widget-title' ) )
+			.getText();
 	}
 
-	getAddress() {
-		return this.driver
+	async getAddress() {
+		return await this.driver
 			.findElement( By.css( '.widget_contact_info div.confit-address a' ) )
 			.getText();
 	}
--- a/lib/components/widget-contact-info-component.js
+++ b/lib/components/widget-contact-info-component.js
@@ -2,9 +2,9 @@
 
 import { By } from 'selenium-webdriver';
 
-import BaseContainer from '../base-container.js';
+import AsyncBaseContainer from '../async-base-container.js';
 
-export default class WidgetContactInfoComponent extends BaseContainer {
+export default class WidgetContactInfoComponent extends AsyncBaseContainer {
 	constructor( driver ) {
 		super( driver, By.css( '.widget_contact_info' ) );
 	}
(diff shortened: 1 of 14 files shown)


alisterscott:
Hi @bsessions85 

Thanks for this PR

I probably haven't explained this very well, but each page we update we also need to update the specs to use either `Expect` or `Visit` from the `AsyncBaseContainer`:
So,

```
test.it( 'Can see the theme thanks dialog', async function() {
  let themeDialogComponent = new ThemeDialogComponent( driver );
  await themeDialogComponent.customizeSite();
} );
```

needs to be updated to:

```
test.it( 'Can see the theme thanks dialog', async function() {
  const themeDialogComponent = await ThemeDialogComponent.Expect( driver );
  await themeDialogComponent.customizeSite();
} );
```

Constructors can't be async, so what we do is have two static async (`Expect` and `Visit`) functions in the `AsyncBaseContainer` which construct the new page - so we need to use these instead of `new`.

The _postInit() function will be called after the object is constructed, if this exists (most pages won't have it).

The theme dialog component has some code which is run after it's constructing using the base constructor, which will no longer work.

So it currently looks like:

```
export default class ThemeDialogComponent extends BaseContainer {
	constructor( driver ) {
		super( driver, By.css( '.themes__thanks-modal' ) );
		this.driver.wait(
			until.elementLocated( By.css( '.themes__thanks-modal h1' ) ),
			this.explicitWaitMS * 2,
			'The message for activating a theme on the "thank you" dialog is not showing after waiting for it to show'
		);
	}
}
```

changing this to extend `AsyncBaseContainer` we need to remove the code after the `super` call.

Two options, one is to use the `_postInit()` function which runs after it's constructed:

```
export default class ThemeDialogComponent extends AsyncBaseContainer {
	constructor( driver ) {
		super( driver, By.css( '.themes__thanks-modal' ) );
	}

	async _postInit() {
		return await driverHelper.waitTillPresentAndDisplayed( this.driver, By.css( '.themes__thanks-modal h1' ), this.explicitWaitMS * 2 );
	}
}
```

The other option is to simply update our expected element to that other element, and increase the default wait:

```
export default class ThemeDialogComponent extends AsyncBaseContainer {
	constructor( driver ) {
		super( driver, By.css( '.themes__thanks-modal h1' ), url, config.get( 'explicitWaitMS' ) * 2 );
	}
}
```

I'll go with the `_postInit()` to keep it the same as it was for now.

I hope this makes sense.

I'll update and merge this PR as you're AFK and so this doesn't get out of date.

lib/components/widget-contact-info-component.js
```diff
@@ -2,9 +2,9 @@
 
 import { By } from 'selenium-webdriver';
 
-import BaseContainer from '../base-container.js';
+import AsyncBaseContainer from '../async-base-container.js';
```
alisterscott: All the methods for a container using AsyncBaseContainer should be `async` and use `await`

alisterscott:
@bsessions85 when converting a page or component to extend `AsyncBaseContainer` we need to check each function is `async` and uses `await` for all calls also.

alisterscott:
After this PR we have 53 async pages and 80 non 👍

PR #1267, opened 2018-06-14
Update WooCommerce spec to be async

### LINKED ISSUE #639: Workaround: WooCommerce Mobile Screens
On the WooCommerce add product screen on mobile sized executions the publish button half dissapears into the header, which means WebDriver can't click it (as I think it is trying to click it in the middle or 0,0)

Workaround in place to execute a javascript click(); but seems hacky and probably best sorted out what is actually happening:

screen shot 2017-07-20 at 6 36 55 pm

Interestingly when resizing wider than narrower again problem goes away?!?

alisterscott:
Fixes some current issues
Also fixes  #639


alisterscott:
@bsessions85 I'm going to merge this as these tests are currently failing

PR #1273, opened 2018-06-15
Update plan selector

alisterscott:
This is so our tests for ready for when https://github.com/Automattic/wp-calypso/pull/25499

This can be merged before that PR as this now forces the 'vertical' variant which is what will be the standard view going forward


alisterscott:
@Automattic/flowpatrol  heads-up I am merging this

PR #1280, opened 2018-06-19
Increase timeout for Jetpack disconnect

brbrr:
Adding long timeout for `Disconnect expired sites` spec, to make sure it deletes most of the sites


brbrr:
I'm going to merge this to prevent flooding of expired sites in `jetpackConnectUser` account

Issue #1282, opened 2018-06-19
Update dependencies


bsessions85:
In #1269 There is a list of dependencies that should be upgraded if possible. Update those that we can.

As a follow-up we can upgrade npm audit fixes:

```
"testarmada-magellan"
"mailosaur"
"grunt"
"grunt-cli"
"growl"
"debug"
"minimatch"
"parsejson"
"lodash"
```

<details><summary>Full <code>npm audit</code> output</summary>

```
                       === npm audit security report ===

# Run  npm install testarmada-magellan@10.1.1  to resolve 15 vulnerabilities
SEMVER WARNING: Recommended action is a potentially breaking change
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Moderate      │ Remote Memory Exposure                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ request                                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ testarmada-magellan                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ testarmada-magellan > node-slackr > request                  │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/309                       │
└───────────────┴──────────────────────────────────────────────────────────────┘


┌───────────────┬──────────────────────────────────────────────────────────────┐
│ High          │ Denial-of-Service Memory Exhaustion                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ qs                                                           │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ testarmada-magellan                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ testarmada-magellan > node-slackr > request > qs             │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/29                        │
└───────────────┴──────────────────────────────────────────────────────────────┘


┌───────────────┬──────────────────────────────────────────────────────────────┐
│ High          │ Denial-of-Service Extended Event Loop Blocking               │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ qs                                                           │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ testarmada-magellan                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
[...]

PR #1283, opened 2018-06-19
Update Dependencies

bsessions85:
For #1282, I've updated the dependencies that we could. Magellan > 10.0.0 does not support Mocha and we were asked by Mailosaur to downgrade from 5.0.3 to 4.0.0 because we were overwhelming their systems somehow.


alisterscott:
Thanks @bsessions85 

> we were asked by Mailosaur to downgrade from 5.0.3 to 4.0.0 because we were overwhelming their systems somehow

I think this was because we were using a new polling method - we could try upgrading the library but keeping our approach the same to see if this goes ok? What do you think?

alisterscott:
Raised the Magellan issue here: https://github.com/Automattic/wp-e2e-tests/issues/1284

bsessions85:
@alisterscott  I just looked at updating Mailosaur the polling feature we were using, waitFor, is the only real option for finding an email in the form we want it. The other option, search, returns a list of emails, but doesn't return any of the emails themselves. A second call would need to be made to get the full email. Do you think that would be worth the effort?

alisterscott:
@bsessions85 can we leave our mailosaur code exactly the same and just update the client, or is the approach we're taking not compatible with the new client? If it's not, I'd say we should just stick with the old client for now - at least until Mailosaur tells us what we should do or releases a fix for `waitFor`

bsessions85:
@alisterscott Yeah, they completely changed the methods so we'll just wait for a fix.

PR #1286, opened 2018-06-20
Update/jetpack onboarding specs

alisterscott:
Updates all pages referenced in `wp-jetpack-onboarding-spec.js` to be async

Stats with this PR

Async pages: 110
Non-Async: 20


alisterscott:
@brbrr thanks for taking this on 👍

PR #1287, opened 2018-06-21
Update logged out home page selector

alisterscott:
We have a new logged out wordpress.com home page
This updates the selector we have to work with it


alisterscott:
@Automattic/flowpatrol merging this as it's causing noise in the e2e test results (deleting accounts)

PR #1290, opened 2018-06-21
Domains: Added new abtest for Kracken and removed the old one

hambai:
This depends on https://github.com/Automattic/wp-calypso/pull/25665


alisterscott:
Reverted in #1314

PR #1296, opened 2018-06-22
Update assert.equal

alisterscott:
`assert.equa`l is deprecated
use `strictEqual` instead

See: https://nodejs.org/docs/latest/api/assert.html#assert_assert_strictequal_actual_expected_message


alisterscott:
On hold until the promise manager is fully disabled

PR #1298, opened 2018-06-22
CircleCI: Fix key -> keys list

sirreal:
Multiple `key` is not the same as `keys`. Allow this to fallback correctly.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,8 +15,9 @@ jobs:
       # Cache first on exact package-lock.json
       # Fallback cache on Node version
       - restore_cache:
-          key: npm-v1-{{ checksum ".nvmrc" }}-{{ checksum "package-lock.json" }}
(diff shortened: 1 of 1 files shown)


sirreal:
Seems to work:

```
npm-v1-et7p7EbN0qTnQ3orb0azlyQ8lf0W7xlMJbDJ2ENSGyM=-K5ncuwC0g0rFIBMHBmqr5VYY5RyyGKPdy5skO_lx8Xk=
```

Before it was (only the second checksum)

```
npm-v1-et7p7EbN0qTnQ3orb0azlyQ8lf0W7xlMJbDJ2ENSGyM=
```

blowery:
Tests failed with a simple timeout, not because of this change. Re-running.

blowery:
And nice catch @sirreal

alisterscott:
I updated from master and all passing now - thanks for your commitment to improving the e2e tests!

PR #1301, opened 2018-06-25
Workaround for Open Live Demo

alisterscott:
Workaround for https://github.com/Automattic/wp-calypso/issues/25720


alisterscott:
Actually, looking at the results the theme switch test still failed on mobile (passed on retry) so maybe this workaround doesn't work

PR #1314, opened 2018-06-28
Revert "Domains: Added new abtest for Kracken and removed the old one"

alisterscott:
Reverts Automattic/wp-e2e-tests#1290


alisterscott:
The wp-calypso PR was reverted

PR #1315, opened 2018-06-28
Disable promise manager in Mocha

### LINKED ISSUE #1313: Disable promise manager when running raw `mocha`
When running raw `mocha` - promise manager is not disabled, which causes differences in test behavior between local & CI test runs.

alisterscott:
Fixes #1313


alisterscott:
@brbrr would you mind checking this does what you expect? In my testing this worked locally

PR #1319, opened 2018-06-29
Updates screenshots and quit browser to be async

alisterscott:
This updates some of the mocha hooks to be async

We missed these in switching to async/await and these can be causing problems

### DIFF
--- a/lib/after.js
+++ b/lib/after.js
@@ -56,7 +56,7 @@ test.afterEach( async function() {
 
 		try {
 			const url = await driver.getCurrentUrl();
-			slackNotifier.warn( `FAILED: Taking screenshot of: '${ url }'` );
+			console.log( `FAILED: Taking screenshot of: '${ url }'` );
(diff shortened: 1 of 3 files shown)


lib/after.js
```diff
@@ -130,6 +126,6 @@ test.after( function() {
 		config.get( 'closeBrowserOnComplete' ) === true ||
 		global.isHeadless === true
 	) {
-		return driverManager.quitBrowser( driver );
+		return await driverManager.quitBrowser( driver );
```
blowery: since this is an async function, you can just return the promise from `driverManager.quitBrowser`. No need for the `await` here.

alisterscott: Thanks, I wasn't sure about that but it makes sense

alisterscott:
I reverted this as once I merged it screenshots were no longer being captured on CircleCI, I will need to work out why

PR #1325, opened 2018-07-03
Add calypsoify AB test

retrofox:
This PR adds the `calypsoifyPlugins` AB test.


alisterscott:
Thanks for the PR - can we keep the `original` behaviour whilst the test runs?

alisterscott:
It looks like this is being implemented without an AB test so closing this PR - reopen if necessary please

PR #1327, opened 2018-07-04
Try using mocha steps

alisterscott:
This converts just one spec to use mocha-steps

The good thing about mocha-steps is it allows us to convert specs one at a time as it's backwards compatible and shouldn't cause any dramas

The idea is to get rid of bail.suite from our codebase that way we can remove our fork of mocha and keep things up to date

Blog post explaining this here: https://watirmelon.blog/2018/07/04/bailing-with-mocha-e2e-tests/

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -73,5 +73,8 @@
 		// Common top-of-file requires, expressions between external, interal
 		"vars-on-top": 1,
 		"yoda": 0
-	}
+	},
+  	"globals": {
+	  "step": false
+  	}
 }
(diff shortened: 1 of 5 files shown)


alisterscott:
> I don't love the way it makes the test results look in Webstorm, but it does do the bail as expected. With the old way it shows everything after the failed step as skipped, not passed.

Are you sure that's not showing the old results? Those steps don't show as passed on command line, they don't appear at all as they haven't been run or considered.

Issue #960, opened 2018-02-21
Create a scenario for a .blog subdomain site


alisterscott:
We currently have none of these

alisterscott:
Closed by #1334 

@bsessions85

PR #1334, opened 2018-07-05
Add/subdomain test

bsessions85:
Addresses #960

### DIFF
--- a/specs/wp-signup-spec.js
+++ b/specs/wp-signup-spec.js
@@ -1401,7 +1401,7 @@ test.describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function()
 		}
 	);
 
-	test.describe.only( 'Sign up for free subdomain site @parallel', function() {
+	test.describe( 'Sign up for free subdomain site @parallel', function() {
 		this.bailSuite( true );
 
 		const blogName = dataHelper.getNewBlogName();
--- a/specs/wp-signup-spec.js
+++ b/specs/wp-signup-spec.js
@@ -1401,12 +1401,11 @@ test.describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function()
 		}
 	);
 
-	test.describe.only( 'Basic sign up for a free site @parallel', function() {
+	test.describe( 'Sign up for free subdomain site @parallel', function() {
 		this.bailSuite( true );
 
 		const blogName = dataHelper.getNewBlogName();
 		const expectedDomainName = `${ blogName }.art.blog`;
-		let newBlogAddress = '';
 
 		test.it( 'Ensure we are not logged in as anyone', async function() {
 			return await driverManager.ensureNotLoggedIn( driver );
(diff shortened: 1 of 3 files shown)


specs/wp-signup-spec.js
```diff
@@ -1399,4 +1400,87 @@ test.describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function()
...
 		}
 	);
+
+	test.describe( 'Sign up for free subdomain site @parallel', function() {
+		this.bailSuite( true );
+
+		const blogName = dataHelper.getNewBlogName();
+		const expectedDomainName = `${ blogName }.art.blog`;
+
+		test.it( 'Ensure we are not logged in as anyone', async function() {
```
alisterscott: Better to use `test.before` as per #1333

specs/wp-signup-spec.js
```diff
@@ -1399,4 +1400,87 @@ test.describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function()
...
+			return await driverManager.ensureNotLoggedIn( driver );
+		} );
+
+		test.it( 'Can enter the subdomains flow and select design type', async function() {
+			await driver.get( config.get( 'calypsoBaseURL' ) + '/start/subdomain/?vertical=a8c.1' );
```
alisterscott: This won't work for live branches. We have the StartPage which offers a function `getStartURL` to do this - see the free sign up scenario for usage.

specs/wp-signup-spec.js
```diff
@@ -1399,4 +1400,87 @@ test.describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function()
...
+			const designTypePage = await DesignTypePage.Expect( driver );
+			await designTypePage.setABTestControlGroupsInLocalStorage( { flow: 'subdomain' } );
```
alisterscott: You can call this off StartPage like the free scenario since this refreshes the page and can cause problems with the page changing based upon AB test groups

specs/wp-signup-spec.js
```diff
@@ -1399,4 +1400,87 @@ test.describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function()
...
+				emailAddress,
+				blogName,
+				passwordForTestAccounts
+			);
+		} );
+
+		test.it(
+			"Can then see the sign up processing page -  will finish and show a 'Continue' button which is clicked",
+			async function() {
+				await SignupProcessingPage.hideFloatiesinIE11( driver );
```
alisterscott: I don't think this necessary as we only run canary tests in IE

specs/wp-signup-spec.js
```diff
@@ -1399,4 +1400,87 @@ test.describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function()
...
+			}
+		);
+
+		test.it(
+			'We are on the view blog page, can see trampoline, our URL and title',
+			async function() {
+				const viewBlogPage = await ViewBlogPage.Expect( driver );
+				await viewBlogPage.waitForTrampolineWelcomeMessage();
+				let displayed = await viewBlogPage.isTrampolineWelcomeDisplayed();
+				return assert.strictEqual(
```
alisterscott: Could you assert we're seeing the correct .blog URL as per the title of this block?

bsessions85:
Made those update. Thanks!

specs/wp-signup-spec.js
```diff
@@ -1399,4 +1400,100 @@ test.describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function()
...
+			const signupFlow = 'subdomain';
+			const startPage = await StartPage.Visit(
+				driver,
+				StartPage.getStartURL( {
+					culture: locale,
+					flow: signupFlow,
+					query: 'vertical=a8c.1'
+				} )
+			);
+			await startPage.setABTestControlGroupsInLocalStorage( { flow: signupFlow} );
```
alisterscott: I don't think is necessary since #1338

alisterscott:
I'll merge this so I can add another scenario to this spec 👍

alisterscott:
> Addresses #960

I looked at why this didn't close the issue - I think you need to use closes, fixes or resolves 👍 

https://help.github.com/articles/closing-issues-using-keywords/

PR #1335, opened 2018-07-05
Try adding simple wait for dserve to not be present

bsessions85:
Trying this to see if we're still on a dserve page. We will wait a little bit for it to be gone.

Thoughts?

### DIFF
--- a/lib/async-base-container.js
+++ b/lib/async-base-container.js
@@ -3,10 +3,12 @@
 import config from 'config';
 import assert from 'assert';
 import { isEqual } from 'lodash';
+import { By } from 'selenium-webdriver';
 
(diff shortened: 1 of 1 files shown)


alisterscott:
I'm not sure about this, I guess it can't harm, but I think we need to work out why our waiting for dserve script doesn't work and live branch tests are running when dserve isn't running

I've added some more logging to the waiting script so we can see what's going on:

https://github.com/Automattic/wp-e2e-tests-for-branches/pull/34

alisterscott:
@bsessions85 I saw this: https://github.com/Automattic/wp-e2e-tests-for-branches/pull/35 which looks good - do you think we still need this PR if that other approach works?

PR #1341, opened 2018-07-10
Add account only sign up

### LINKED ISSUE #1340: Account only sign up flow
Add account only sign up flow

alisterscott:
Fixes #1340 

This is in the new /signup sub directory - if this looks okay I will split the existing sign up specs from the sign up spec into this sub directory.


alisterscott:
@bsessions85 I actually realized this new spec doesn't get executed as it's in a subdirectory 🤔

PR #1342, opened 2018-07-10
Updates the e2e test as the checklist a/btest ends

taggon:
With this update the e2e test can cover https://github.com/Automattic/wp-calypso/pull/25929

### DIFF
--- a/specs/wp-signup-spec.js
+++ b/specs/wp-signup-spec.js
@@ -34,6 +34,7 @@ import ThemeDetailPage from '../lib/pages/theme-detail-page';
 import AccountSettingsPage from '../lib/pages/account/account-settings-page';
 import CloseAccountPage from '../lib/pages/account/close-account-page';
 import DesignTypePage from '../lib/pages/signup/design-type-page';
+import ChecklistPage from '../lib/pages/checklist-page';
 
 import FindADomainComponent from '../lib/components/find-a-domain-component.js';
 import SecurePaymentComponent from '../lib/components/secure-payment-component.js';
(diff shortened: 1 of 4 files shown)


alisterscott:
Thanks for your contribution @taggon 

I was thinking you could just override the AB test behaviour to not show the checklist and therefore no functional changes are needed to the e2e tests?

What do you think?

taggon:
Although the CI failed, this PR works well for the corresponding Calypso PR.

specs/wp-signup-spec.js
```diff
@@ -1188,18 +1543,89 @@ describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function() {
...
-		if ( locale === 'en' ) {
-			step( 'Can see the correct blog title displayed', async function() {
-				const viewBlogPage = await ViewBlogPage.Expect( driver );
-				const title = await viewBlogPage.title();
-				return assert.strictEqual(
-					title,
-					'Site Title',
-					'The expected blog title is not displaying correctly'
-				);
+	describe( 'Basic sign up for a free blog @parallel @email @canary @ie11canary', function() {
```
alisterscott: Can you please remove `@canary` and `@ie11canary` from this as we only want a single canary and the non-checklist one will be good for now to ensure backwards compatibility with existing PRs

specs/wp-signup-spec.js
```diff
@@ -202,7 +203,140 @@ describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function() {
...
+				return signupProcessingPage.continueAlong();
+			}
+		);
+
+		step( 'Can then see the onboarding checklist', async function() {
+			const checklistPage = await ChecklistPage.Expect( driver );
+			const header = await checklistPage.header();
+			const subheader = await checklistPage.subheader();
+
+			assert.strictEqual(
```
alisterscott: We avoid asserting on strings as a) they're translated so these can change with locales, and b) they can change even in `en` so we just avoid them. Could you just assert a `.formatted-header__subtitle` and `.formatted-header__title` is present and displayed? That should be enough I think.

taggon:
@alisterscott I've updated this PR as per your request but the test failed again. Trying to look into it.

PR #1344, opened 2018-07-11
Using sha for dserve

bsessions85:
This should help with getting the correct version of calypso.live before we start the test.

Should be merged with https://github.com/Automattic/wp-e2e-tests-gh-bridge/pull/45

### DIFF
--- a/docs/running-tests.md
+++ b/docs/running-tests.md
@@ -52,6 +52,7 @@ The `run.sh` script takes the following parameters, which can be combined to exe
     -a [workers]  - Number of parallel workers in Magellan (defaults to 3)
     -R      - Use custom Slack/Spec/XUnit reporter, otherwise just use Spec reporter
     -p      - Execute the tests in parallel via CircleCI envvars (implies -g -s mobile,desktop)
+    -S [commitHash]  - Run tests against given commit via https://calypso.live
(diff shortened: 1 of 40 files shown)


bsessions85:
Tested in https://circleci.com/gh/Automattic/wp-e2e-tests-for-branches/11526#build-parameters/containers/0

alisterscott:
Hi @bsessions85 

Thanks for taking this on, this looks good and worked well in my local testing:

`./run.sh -C -b fb14a04f61db4c18571a418fd7e1760886a040c0 `

I noticed that the `docs/running-tests.md` needs an update

One thought though, could we make this change backwards compatible so we could either use branch or hash for now? That way when we're merging this we won't get an issues with live branches being executed using the wrong parameters and trying to run branches against hashes etc.

Just an idea, perhaps another run.sh argument that kicks off a build with the hash instead of the branch, and then we can get this merged, then merge the other PRs that send the updated argument?

What do you think?

bsessions85:
Yeah, that makes sense and is a good idea. I updated it and tested that it works with either one.

alisterscott:
Tests well:

`./run.sh -C -b fb14a04f61db4c18571a418fd7e1760886a040c0`

PR #1345, opened 2018-07-12
Update domain mapping spec to use new flow.

delputnam:
This should fix the tests that were broken by Automattic/wp-calypso#25573 and disabled in #1336.

It updates the mapping section of the domain management spec to use the new transfer/mapping selection page.


lib/pages/domain-my-own-page.js
```diff
@@ -2,10 +2,19 @@
...
 	constructor( driver ) {
-		super( driver, By.css( '.transfer-domain-step__domain-description' ) );
+		super( driver, By.css( '.use-your-domain-step__content' ) );
+	}
+
+	async selectBuyDomainMapping() {
+		return await driverHelper.clickWhenClickable(
+			this.driver,
+			By.css( '.use-your-domain-step__option-button:not(.is-primary)' ),
+			this.explicitWaitMS
```
alisterscott: Minor -` this.explicitWaitMS` isn't needed as it's the default

specs/wp-manage-domains-spec.js
```diff
@@ -15,7 +15,6 @@ import ShoppingCartWidgetComponent from '../lib/components/shopping-cart-widget-
 import SidebarComponent from '../lib/components/sidebar-component.js';
 import NavBarComponent from '../lib/components/nav-bar-component.js';
 import MyOwnDomainPage from '../lib/pages/domain-my-own-page';
-import MapADomainComponent from '../lib/components/map-a-domain-component';
```
alisterscott: We can remove this whole file I believe as I can't find any other references in the project to it

alisterscott:
Thanks for your PR @delputnam!

Worked well 

```
$ env BROWSERSIZE=desktop ./node_modules/.bin/mocha specs


  [WPCOM] Managing Domains: (desktop) @parallel
    Map a domain to an existing site @parallel
      ✓ Log In and Select Domains (16770ms)
      ✓ Can see the Domains page and choose add a domain (1763ms)
      ✓ Can see the domain search component (2535ms)
      ✓ Can select to use an existing domain (123ms)
      ✓ Can see use my own domain page
      ✓ Can select to buy domain mapping (91ms)
      ✓ Can see enter a domain component
      ✓ Can enter the domain name (167ms)
      ✓ Can add domain to the cart (70ms)
      ✓ Can see checkout page (1415ms)


  10 passing (30s)
```

lib/pages/domain-my-own-page.js
delputnam: Ah. Ok. Good to know. Thanks!

specs/wp-manage-domains-spec.js
delputnam: Argh. Yep. I meant to do that. Thanks for reminding me!

delputnam:
Thanks for taking a look @alisterscott. This was fun doing my first update to the e2e tests! I've made the updates you mentioned.

aidvu:
@delputnam / @alisterscott would be nice to add Transfer e2e tests too? What do you think? :)

aidvu:
Probably not the full flow, but to make sure we end up on the Transfer screen and a bad auth code returns an error?

alisterscott:
> Probably not the full flow, but to make sure we end up on the Transfer screen and a bad auth code returns an error?

Raised and tracking this in #1353

Issue #1347, opened 2018-07-13
Upgrade node to 10.6.0 to match wp-calypso


alisterscott:
branch: upgrade/node-10.6.0
PR title: Upgrade Node to 10.6.0

- [x] this main repo
- [x] https://github.com/Automattic/wp-e2e-tests-for-branches
- [x] https://github.com/Automattic/wp-e2e-tests-canary
- [x] https://github.com/Automattic/wp-e2e-tests-woocommerce
- [x] https://github.com/Automattic/wp-e2e-tests-i18n
- [x] https://github.com/Automattic/wp-e2e-tests-ie11
- [x] https://github.com/Automattic/wp-e2e-tests-jetpack
- [x] https://github.com/Automattic/wp-e2e-tests-jetpack-be

alisterscott:
Finished 👍

PR #1348, opened 2018-07-13
Upgrade Node to 10.6.0

alisterscott:
https://github.com/Automattic/wp-e2e-tests/issues/1347


alisterscott:
Merging since this is a task we repeat and passing tests is enough 👍

PR #1350, opened 2018-07-13
Add plansBannerFromDomainNudge AB test

travisw:
This adds the Calypso plansBannerFromDomainNudge AB test.
Calypso PR: Automattic/wp-calypso#26034


travisw:
hmm, I'm not sure about the failed test. It doesn't look related.

bsessions85:
@travisw That is a strange message. I suspect it is the extra commas at the end of the test arrays. Try taking those off

PR #1352, opened 2018-07-16
Add Jetpack connection smoke test

brbrr:
Adds simple connect/disconnect spec for Jetpack. Was inspired recent Jetpack's point release: https://github.com/Automattic/jetpack/pull/9859

The idea is to run this test using every Jetpack test host, to cover most of the PHP/WP versions. Right now this test is not included in CI builds.

To test:

* Decrypt config. New user is added.
* run `JETPACKHOST=PRESSABLE TARGET=JETPACK mocha specs-jetpack/connect-disconnect-spec.js`


alisterscott:
This ran well for me:

```
$ JETPACKHOST=PRESSABLE TARGET=JETPACK BROWSERSIZE=mobile ./node_modules/.bin/mocha specs-jetpack/connect-disconnect-spec.js


  Disconnect wporg site
    ✓ Can disconnect wporg site (16861ms)

  Jetpack Connect and Disconnect: (mobile)
    Connect Jetpack and see if post page is loading correctly
      ✓ Can login into WordPress.com (13270ms)
      ✓ Login into wporg site (7586ms)
      ✓ Can navigate to the Jetpack dashboard (182ms)
      ✓ Can click the Connect Jetpack button (6368ms)
Found an AB Testing key in local storage that isn't known: 'nudgeAPalooza'. This may cause inconsistent A/B test behaviour, please check this is okay and add it to 'knownABTestKeys' in default.config
      ✓ Can approve connection on the authorization page (8173ms)
      ✓ Can click the free plan button (6570ms)
      ✓ Can navigate to the Posts page (263ms)
      ✓ Can load a post and make sure it is loaded correctly (4265ms)
    Disconnect from Jetpack and load a post page
      ✓ Can disconnect Jetpack connection in wp-admin (14333ms)
      ✓ Can navigate to the Posts page (188ms)
      ✓ Can load a post and make sure it is loaded correctly (1850ms)


  12 passing (1m)

```

My main concern is repeatabilty and parallelism. How could this spec run across two screensizes in CI say when it uses a single site and disconnects/reconnects the site over and over again? If the desktop test was in the process of connecting the site and the mobile test disconnected it then it would fail and we may not know why. This is why jurassic ninja spinning up brand new sites are good as they are standalone and won't tread on each other - what's the intention with this test? Why does it need to be a single site?

brbrr:
> My main concern is repeatability and parallelism

Yeah, you absolutely right, it will not work like this. I plan not to run it continuously, but rather on every beta release, and using a single screen size.

> what's the intention with this test?

The idea is to check wp-admin/frontend and make sure there no fatals there in different PHP/WP configurations. Since Jetpack supports a variety of PHP versions - there might be some compatibility issues etc on different hosts.

> Why does it need to be a single site?

Not sure that I understand your question. Right now it uses a single site since it is an MVP, all other sites will be added in further PRs

PR #1354, opened 2018-07-17
adding nudgeapalooza test

bsessions85:
This test wasn't in test list and was making noise in slack


bsessions85:
merging to fix the test broken and noise in logs

PR #1355, opened 2018-07-18
Update theme selector to be backwards compatible

alisterscott:
Whenever a PR isn't rebased against master this test will fail as the selector has been changed


bsessions85:
I totally forgot about being able to include multiple selectors. Thanks for updating!

PR #1358, opened 2018-07-18
Adding Invite Followers test back in

bsessions85:
I researched why adding followers test wasn't working. It had been removed so I am adding it back as a disabled test pending 26178-wp-calypso. 

I'm not sure if we should add this back in now, or wait until the bug is fixed, but here it is.


alisterscott:
> I'm not sure if we should add this back in now, or wait until the bug is fixed, but here it is.

I'm in two minds on disabled tests - at least if they're in there then it's easy to find usages etc. but if they're not running anyway they'll get out of date as they don't work.

So to answer your question: sorry I don't have an answer 😝

bsessions85:
I left a note on the issue about enabling this test when the bug is fixed. Merging it in now.

PR #1359, opened 2018-07-19
Refactor getting URLs

### LINKED ISSUE #1346: Move getting calypso live URL into one place
we have something like this in a number of places:

```
if ( dataHelper.isRunningOnLiveBranch() ) {
	url = url + '?' + dataHelper.getCalypsoLiveQuery();
 }
```

we could move this into one place to make updating anything to do with calypso.live easier

See: #1344

alisterscott:
Fixes #1346 

This cleans up the usage of getting the base URL for calypso and checking whether we're on a live branch

**Review / Testing Instructions**

1. Make sure all tests are passing on CI
2. Make sure the code is clean and easy to understand

### DIFF
--- a/lib/data-helper.js
+++ b/lib/data-helper.js
@@ -263,12 +263,12 @@ export function getCalypsoURL( route = '', queryStrings = [] ) {
 	}
 
 	if ( this.isRunningOnLiveBranch() ) {
-		queryString = this._appendQueryString( queryString, this.getCalypsoLiveQuery() );
 		if ( url.indexOf( 'calypso.live' ) < 0 ) {
 			throw new Error(
 				`It looks like live branches are being used on a non calypso.live URL: '${ url }'`
 			);
 		}
+		queryString = this._appendQueryString( queryString, this.getCalypsoLiveQuery() );
 	}
 	return url + queryString;
 }
--- a/lib/data-helper.js
+++ b/lib/data-helper.js
@@ -254,16 +254,28 @@ export function getCalypsoLiveQuery() {
 	return 'branch=' + config.get( 'branchName' );
 }
 
-export function getCalypsoURL( route = '' ) {
-	let url;
-	url = this.configGet( 'calypsoBaseURL' ) + '/' + route;
+export function getCalypsoURL( route = '', queryStrings = [] ) {
+	let queryString = '';
+	let url = this.configGet( 'calypsoBaseURL' ) + '/' + route;
+
+	for ( let qs of queryStrings ) {
+		queryString = this._appendQueryString( queryString, qs );
+	}
+
 	if ( this.isRunningOnLiveBranch() ) {
+		queryString = this._appendQueryString( queryString, this.getCalypsoLiveQuery() );
 		if ( url.indexOf( 'calypso.live' ) < 0 ) {
 			throw new Error(
 				`It looks like live branches are being used on a non calypso.live URL: '${ url }'`
 			);
 		}
-		url = url + '?' + this.getCalypsoLiveQuery();
 	}
-	return url;
+	return url + queryString;
+}
+
+export function _appendQueryString( existingQueryString, queryStringPair ) {
+	if ( existingQueryString === '' ) {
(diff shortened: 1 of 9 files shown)


lib/data-helper.js
```diff
@@ -253,3 +253,29 @@ export function getCalypsoLiveQuery() {
 	}
 	return 'branch=' + config.get( 'branchName' );
 }
+
+export function getCalypsoURL( route = '', queryStrings = [] ) {
+	let queryString = '';
```
alisterscott: is it obvious how `queryString` is different from `queryStrings`?

lib/data-helper.js
```diff
@@ -253,3 +253,29 @@ export function getCalypsoLiveQuery() {
...
+	let url = this.configGet( 'calypsoBaseURL' ) + '/' + route;
+
+	for ( let qs of queryStrings ) {
+		queryString = this._appendQueryString( queryString, qs );
+	}
+
+	if ( this.isRunningOnLiveBranch() ) {
+		if ( url.indexOf( 'calypso.live' ) < 0 ) {
```
alisterscott: This isn't really necessary but I thought it would be nice to stop any accidental errors on trying live branch testing using a different environment like wordpress.com (sometimes I forget to update the config URL when debugging)

lib/pages/login-page.js
```diff
@@ -96,19 +95,6 @@ export default class LoginPage extends AsyncBaseContainer {
 	}
 
 	static getLoginURL() {
-		const baseURL = dataHelper.configGet( 'calypsoBaseURL' );
-		let loginURL = `${ baseURL }/log-in`;
-
-		// If Calypso is running locally in Docker, force login to actual WordPress.com first
```
alisterscott: we don't support local calypso docker any more - use calypso.live instead (which uses docker anyway)

lib/pages/reader-page.js
```diff
@@ -35,10 +35,6 @@ export default class ReaderPage extends AsyncBaseContainer {
 	}
 
 	static getReaderURL() {
-		let url = dataHelper.configGet( 'calypsoBaseURL' ) + '/read';
-		if ( dataHelper.isRunningOnLiveBranch() ) {
-			url = url + '?' + dataHelper.getCalypsoLiveQuery();
-		}
-		return url;
+		return dataHelper.getCalypsoURL( 'read' );
```
alisterscott: I actually don't think wordpress.com/read is different to wordpress.com/ 🤷‍♂️

lib/pages/signup/start-page.js
```diff
@@ -14,52 +14,22 @@ export default class StartPage extends AsyncBaseContainer {
...
-		url = dataHelper.configGet( 'calypsoBaseURL' ) + '/start';
+	static getStartURL( { culture = 'en', flow = '', query = '' } = {} ) {
+		let route = 'start';
+		let queryStrings = [];
 
 		if ( flow !== '' ) {
-			url += '/' + flow;
-		}
-
-		if ( domainFirst === true ) {
```
alisterscott: I'm passing domain first as the flow now and using `query` which is now supported 🎉

lib/pages/signup/start-page.js
```diff
@@ -14,52 +14,22 @@ export default class StartPage extends AsyncBaseContainer {
...
-		url += queryStrings;
-		return url;
-	}
-
-	static _appendQueryString( existingQueryString, queryStringPair ) {
-		if ( existingQueryString === '' ) {
-			return `?${ queryStringPair }`;
-		}
-		return `${ existingQueryString }&${ queryStringPair }`;
+		return dataHelper.getCalypsoURL( route, queryStrings );
```
alisterscott: this does all the live branch stuff 👍

lib/data-helper.js
bsessions85: Yes, I think it is clear enough

lib/data-helper.js
bsessions85: Makes sense to me!

PR #1360, opened 2018-07-20
Update/domains spec skip

alisterscott:
The map a domain spec wasn't being skipped when SKIP_DOMAIN_TESTS=true

This skips these scenarios when that is set


alisterscott:
Merging as this is causing problems on `master`

Issue #1353, opened 2018-07-17
Add partial domain transfer e2e test


alisterscott:
As per https://github.com/Automattic/wp-e2e-tests/pull/1345#issuecomment-405336361

aidvu:
@alisterscott want me to try and implement it when back from AFK?

alisterscott:
@aidvu sounds good to me - happy to provide any help necessary 👍

PR #1368, opened 2018-07-25
Domains: Add partial Transfer Domain test

aidvu:
- Should be able to select transfer screen
- Should be able to get to precheck screen
- Make sure there's an auth code error

Fixes: https://github.com/Automattic/wp-e2e-tests/issues/1353

### DIFF
--- a/specs/wp-manage-domains-spec.js
+++ b/specs/wp-manage-domains-spec.js
@@ -16,6 +16,8 @@ import SidebarComponent from '../lib/components/sidebar-component.js';
 import NavBarComponent from '../lib/components/nav-bar-component.js';
 import MyOwnDomainPage from '../lib/pages/domain-my-own-page';
 import MapADomainPage from '../lib/pages/domain-map-page';
+import TransferDomainPage from '../lib/pages/transfer-domain-page';
+import TransferDomainPrecheckPage from '../lib/pages/transfer-domain-precheck-page';
 import EnterADomainComponent from '../lib/components/enter-a-domain-component';
 import MapADomainCheckoutPage from '../lib/pages/domain-map-checkout-page';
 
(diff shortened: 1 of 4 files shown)


specs/wp-manage-domains-spec.js
```diff
@@ -185,4 +187,84 @@ describe( `[${ host }] Managing Domains: (${ screenSize }) @parallel`, function(
 			return await shoppingCartWidgetComponent.empty();
 		} );
 	} );
+
+	describe( 'Transfer a domain to an existing site (partial) @parallel', function() {
+		const domain = 'andrija.me';
```
bsessions85: I think this should probably be put in to the encrypted config file

specs/wp-manage-domains-spec.js
```diff
@@ -185,4 +187,84 @@ describe( `[${ host }] Managing Domains: (${ screenSize }) @parallel`, function(
...
+		} );
+
+		step( 'Click check auth code', async function() {
+			const transferDomainPrecheckPage = await TransferDomainPrecheckPage.Expect( driver );
+			return await transferDomainPrecheckPage.clickCheckAuthCode();
+		} );
+
+		step( 'Can see auth code error', async function() {
+			const transferDomainPrecheckPage = await TransferDomainPrecheckPage.Expect( driver );
+			return await transferDomainPrecheckPage.clickCheckAuthCode();
```
bsessions85: Should this 'authCodeErrorDisplayed?

specs/wp-manage-domains-spec.js
aidvu: This is just a random taken domain. Nothing important in any way. :)

specs/wp-manage-domains-spec.js
aidvu: 🤦‍♂️

PR #1371, opened 2018-07-26
Update date on nudgeAPalooza test

adamziel:
We want to enable nudgeAPalooza test for everyone. We had a happy accident where some users got already assigned to this test with a skewed distribution. In a related PR (https://github.com/Automattic/wp-calypso/pull/26330) the datestamp is updated to start fresh allocation, and this PR reflects that.


alisterscott:
Merging since the wp-calypso change is live

PR #1372, opened 2018-07-26
Domains: Update transfer check up to precheck screen

aidvu:
No point checking stuff on the precheck screen as these
might change, and we're testing with real domains. Make sure
we can get to the page.


alisterscott:
Thanks @aidvu!!!

PR #1373, opened 2018-07-26
Skip domains tests if domain registration is unavailable

### LINKED ISSUE #1308: Cater for Domains Unavailability
Sometimes the ability to search/buy domains are globally disabled

In this case, we should be able to detect this screen and skip the domains spec with a warning to Slack until domains are enabled again - instead of failures

failed-en-desktop-can-select-to-use-an-existing-domain-1530058446321

Stojdza:
Check if domain registration is available, and skip tests if it's not. 

**To test `wp-manage-domains-spec`:**
- change [this line](https://github.com/Automattic/wp-calypso/blob/7b91038e6b42d9fc63722cdc051d76e95e5ba698/client/my-sites/domains/domain-search/domain-search.jsx#L126) in Calypso to `if ( this.state.domainRegistrationAvailable ) {` 
- run tests against that Calypso
- domain registration should be unavailable

**To test `wp-signup-spec`:**
_**NOTE:** This includes only `Sign up for a site on a business paid plan w/ domain name coming in via /create as business flow in CAD currency @parallel` test._
- Change blocks of [this](https://github.com/Automattic/wp-calypso/blob/3048bc27bd9d5844659cb6ee5291580ced75c74d/client/components/domains/register-domain-step/index.jsx#L683) and [this](https://github.com/Automattic/wp-calypso/blob/3048bc27bd9d5844659cb6ee5291580ced75c74d/client/components/domains/register-domain-step/index.jsx#L617) functions to:
```
this.showValidationErrorMessage( domain, 'domain_registration_unavailable', {
     	maintenanceEndTime: null,
} );
```
- run tests against that Calypso
- domain registration should be unavailable

**Result:** Tests should be skipped and not failed. 

**Additional changes:** `Sign up for a domain only purchase coming in from wordpress.com/domains in EUR currency @parallel` - Now it opens _/domains_ page and searches for a domain instead opening direct link for a wanted _.live_ domain. This was needed to be done because we will add check for domain registration unavailability in the future. 

Fixes #1308

### DIFF
--- a/specs/wp-signup-spec.js
+++ b/specs/wp-signup-spec.js
@@ -45,7 +45,6 @@ import NoSitesComponent from '../lib/components/no-sites-component';
 import * as SlackNotifier from '../lib/slack-notifier';
 
 import EmailClient from '../lib/email-client.js';
-import RegistrationUnavailableComponent from '../lib/components/domain-registration-unavailable-component';
 
 const mochaTimeOut = config.get( 'mochaTimeoutMS' );
 const startBrowserTimeoutMS = config.get( 'startBrowserTimeoutMS' );
@@ -94,17 +93,7 @@ describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function() {
 		step(
 			'Can then see the domains page, and Can search for a blog name, can see and select a free .wordpress address in the results',
 			async function() {
-				let findADomainComponent;
-				try {
-					findADomainComponent = await FindADomainComponent.Expect( driver );
-				} catch ( err ) {
-					if ( await RegistrationUnavailableComponent.Expect( driver ) ) {
-						await SlackNotifier.warn( 'SKIPPING: Domain registration is currently unavailable. ', {
-							suppressDuplicateMessages: true,
-						} );
-						return this.skip();
-					}
-				}
+				const findADomainComponent = await FindADomainComponent.Expect( driver );
 				await findADomainComponent.searchForBlogNameAndWaitForResults( blogName );
 				await findADomainComponent.checkAndRetryForFreeBlogAddresses(
 					expectedBlogAddresses,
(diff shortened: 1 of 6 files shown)


alisterscott:
Thanks for the update @Stojdza!

The manage domains spec works well 👍 

```
env BROWSERSIZE=desktop ./node_modules/.bin/mocha specs/wp-manage-domains-spec.js 


  [WPCOM] Managing Domains: (desktop) @parallel
    Adding a domain to an existing site 
      ✓ Log In and Select Domains (9482ms)
      ✓ Can see the Domains page and choose add a domain (1984ms)
SKIPPING: Domain registration is currently unavailable. 
      - Can see the domain search component
      - Can search for a blog name
      - Can select the .com search result and decline Google Apps for email
      - Can see checkout page, choose privacy and enter registrar details
      - Can then see secure payment component
      - Empty the cart
    Map a domain to an existing site @parallel
      ✓ Log In and Select Domains (8555ms)
      ✓ Can see the Domains page and choose add a domain (2250ms)
      - Can see the domain search component
      - Can select to use an existing domain
      - Can see use my own domain page
      - Can select to buy domain mapping
      - Can see enter a domain component
      - Can enter the domain name
      - Can add domain to the cart
      - Can see checkout page
      - Empty the cart
    Transfer a domain to an existing site (partial) @parallel
      - Log In and Select Domains
      - Can see the Domains page and choose add a domain
      - Can see the domain search component
      - Can select to use an existing domain
      - Can see use my own domain page
      - Can select to transfer a domain
      - Can see the transfer my domain page
      - Can enter the domain name
      - Click transfer domain button
      - Can see the transfer precheck page
      - Can enter auth code on the transfer precheck page
      - Click check auth code
      - Can see auth code error


  4 passing (1m)
  28 pending
```

With the sign up spec; I couldn't get this to fail because the step still shows just doesn't show paid results:

screen shot 2018-07-30 at 2 07 41 pm

So I was thinking for the sign up tests that register a paid domain only, we could maybe just check the existence of the info message on that screen and abort those tests. The sign up tests that use free wordpress.com addresses don't need to be updated AFAIK

specs/wp-signup-spec.js
```diff
@@ -93,7 +94,17 @@ describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function() {
 		step(
 			'Can then see the domains page, and Can search for a blog name, can see and select a free .wordpress address in the results',
 			async function() {
-				const findADomainComponent = await FindADomainComponent.Expect( driver );
+				let findADomainComponent;
```
alisterscott: I don't think we need to check for unavailability as this uses a free wordpress.com address which doesn't depend on the domains service

Stojdza: Agree. I'll add checking only for paid domains and remove these.

alisterscott:
@Stojdza you'll need to rebase or merge master into this since https://github.com/Automattic/wp-e2e-tests/pull/1342 has landed which made some changes to the sign up spec

Stojdza:
I've added changes for signup spec and update PR description.

alisterscott:
Sorry I didn't get to review this today - @bsessions85: would you mind re-reviewing the latest updates?

bsessions85:
This is looking good to me. I'm curious as to why we're not adding the check in `Sign up for a domain only purchase coming in from wordpress.com/domains in EUR currency @parallel` now. If domains goes down, this test will still fail.

Stojdza:
@bsessions85 For now there is no way to test it locally, and because of that I left [_TODO_](https://github.com/Automattic/wp-e2e-tests/blob/add/check-unavailable-registration/specs/wp-signup-spec.js#L1107)

bsessions85:
🚢

PR #1375, opened 2018-07-28
Update/jetpack async driver helper

### LINKED ISSUE #1295: Revisit DriverHelper
Since we're using async and await now, we should revisit the DriverHelper class and do some cleanup of these functions

These are broadly used so we need to make sure we don't break anything in wp.com and jetpack e2e tests by changing these

We also rely on these for catching things like stale element exceptions so moving to await we should still be able to catch errors and retry within these functions

mel-OR:
Fixes #1295

Changed a few DriverHelper function calls that used promise chain to use async/await instead.
Also added a brief update on how to name branches so jetpack tests run in CI.

### DIFF
--- a/docs/miscellaneous.md
+++ b/docs/miscellaneous.md
@@ -7,6 +7,7 @@
 - [Launch Logged-In Window](#launch-logged-in-window)
 - [User account requirements](#user-account-requirements)
 - [List of wrapper repos & friends](#list-of-wrapper-repos--friends)
+- [What to name your branch](#naming-branches)
 
 ## NodeJS Version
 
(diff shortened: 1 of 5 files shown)


docs/miscellaneous.md
```diff
@@ -7,6 +7,7 @@
 - [Launch Logged-In Window](#launch-logged-in-window)
 - [User account requirements](#user-account-requirements)
 - [List of wrapper repos & friends](#list-of-wrapper-repos--friends)
+- [What to name your branch](#naming-branches)
```
alisterscott: I coudn't get the anchor link to work - I think the anchor needs to be something like #what-to-name-your-branch to work

mel-OR: I will take another look.

PR #1379, opened 2018-07-31
Upping timeout on transfer domains precheck

bsessions85:
Tests are timing out. This fixes it locally

### DIFF
--- a/lib/pages/transfer-domain-precheck-page.js
+++ b/lib/pages/transfer-domain-precheck-page.js
@@ -6,6 +6,6 @@ import AsyncBaseContainer from '../async-base-container';
 
 export default class TransferDomainPrecheckPage extends AsyncBaseContainer {
 	constructor( driver ) {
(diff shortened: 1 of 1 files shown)


bsessions85:
Merging to get tests passing again

lib/pages/transfer-domain-precheck-page.js
```diff
@@ -6,6 +6,6 @@ import AsyncBaseContainer from '../async-base-container';
 
 export default class TransferDomainPrecheckPage extends AsyncBaseContainer {
 	constructor( driver ) {
-		super( driver, By.css( '.transfer-domain-step__precheck' ) );
+		super( driver, By.css( '.transfer-domain-step__precheck' ), null, 40000 );
```
alisterscott: we try to avoid hardcoded values like this, maybe use `this.explicitWait * 2` which will give you the same result

bsessions85: Great, thanks for letting me know, and for updating!

PR #1384, opened 2018-08-01
Disable test

bsessions85:
Test is failing on wp-e2e-tests-for-branches because /domains isn't in calypso

@Stojdza


bsessions85:
Merging to get calypso.live tests running

alisterscott:
The reason this doesn't work is /domains doesn't exist on calypso and therefore won't exist on calypso.live - we should what we did before and start at the entry point in calypso

PR #1387, opened 2018-08-02
Add videos to test runs in CircleCI

bsessions85:
This PR adds recording of test runs when run in CircleCI. All runs are recorded, but passing runs are deleted so that there aren't a bunch of useless videos. The videos are named after the step in which the test failed, similar to screenshots.

Here is a build where I added some intentional test failures- https://circleci.com/gh/Automattic/wp-e2e-tests/21836

Here is an example video - https://21836-57936731-gh.circle-artifacts.com/1/home/circleci/wp-e2e-tests/screenshots/videos/can-see-the-correct-blog-url-displayed-2018-08-03T18-54-20.mpg

Testing this one requires getting tests to fail in CircleCI. I have done that in the links above, but it would also be good to check that tests pass and fail as expected when run locally, even though videos will not be created.

Please let me know if you see any room for improvements or have any questions!


bsessions85:
@brbrr  I think the delay is more an issue with tests that are re-tried. Of course I can't get a smooth run with no retries when I want to, but when desktop tests ran without issues, that finished in 9:22, which is right where I would expect it with 15-20 seconds of overhead for installing ffmpeg.

bsessions85:
Updated to use guard clauses as suggested by @brbrr

PR #1389, opened 2018-08-03
Remove scheduled post when test is finished

### LINKED ISSUE #1229: Create a cron job to clean-up scheduled posts
Our `defaultUser` has a warning in its Report Card: `This blog has had post scheduling limited due to having 401 scheduled posts.`

We need to review tests accounts used in `Schedule Basic Public Post` test and create/update a CRON job to clean-up these scheduled posts/

Stojdza:
After successful test `Schedule Basic Public Post @parallel @jetpack`, scheduled post is trashed. 

Fixes #1229

### DIFF
--- a/specs/wp-post-editor-spec.js
+++ b/specs/wp-post-editor-spec.js
@@ -531,6 +531,11 @@ describe( `[${ host }] Editor: Posts (${ screenSize })`, function() {
 					'The newly scheduled post is not showing in the editor as scheduled'
 				);
 			} );
+
+			after( async function() {
+				let postEditorSidebarComponent = await PostEditorSidebarComponent.Expect( driver );
+				await postEditorSidebarComponent.trashPost();
+			} );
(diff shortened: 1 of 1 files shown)


specs/wp-post-editor-spec.js
```diff
@@ -531,6 +531,11 @@ describe( `[${ host }] Editor: Posts (${ screenSize })`, function() {
 					'The newly scheduled post is not showing in the editor as scheduled'
 				);
 			} );
+
+			after( async function() {
```
alisterscott: You probably need the _after_ here (just use a _step_) since this will run after any step that fails, so if it fails logging on then it will still try to delete the post (and also fail)

Stojdza:
Thanks for the feedback @bsessions85 @alisterscott 

Now it's step instead after, and it will remove post only if everything was successful before that step.

Issue #1386, opened 2018-08-02
GDPR cookie banner causes test failures in Europe


rachelmcr:
When running the tests locally in Europe, some steps fail because the GDPR cookie banner overlaps the click target.

An example where the step "Can open themes menu" failed because it's hidden behind the banner:

<img src="https://user-images.githubusercontent.com/8658164/43599692-589ec786-9680-11e8-97de-064fb1e15789.png" alt="failed-en-mobile-can-open-themes-menu-1533224790967" width="50%">

Should we add a step to dismiss this banner if it shows up? I'm open to other workarounds if it's only a concern for folks running the tests locally, and not something we want to change in the tests themselves.

rachelmcr:
I confirmed that running the tests with my VPN set to a non-EU location resolves it for me.

I'll work on a fix to dismiss the banner when it appears.

rachelmcr:
I ran across an earlier discussion of the GDPR banner in our tests and am going to try this approach (set a cookie to not show the banner at all, instead of waiting/dismissing it): https://github.com/Automattic/wp-e2e-tests/pull/1232#issuecomment-392725610

PR #1390, opened 2018-08-03
Set cookie to hide GDPR banner regardless of location

rachelmcr:
Fixes https://github.com/Automattic/wp-e2e-tests/issues/1386

This PR sets a cookie on the home page (for signup tests) and login page (for other specs) that prevents the GDPR cookie banner from showing.

This resolves an issue where running the tests from within the EU results in failures due to the cookie banner overlapping page elements used in the tests.

To test:
1. Set your location as in the EU (e.g. with a VPN).
2. Run the full suite of tests with `./run.sh -g`.
3. Confirm that the GDPR cookie banner does not appear and there are no related test failures.

### DIFF
--- a/lib/pages/login-page.js
+++ b/lib/pages/login-page.js
@@ -98,11 +98,4 @@ export default class LoginPage extends AsyncBaseContainer {
 	static getLoginURL() {
 		return dataHelper.getCalypsoURL( 'log-in' );
 	}
-
-	async setGdprCookie() {
-		const setCookieCode = function() {
-			window.document.cookie = 'sensitive_pixel_option=no;domain=wordpress.com';
-		};
-		return await this.driver.executeScript( setCookieCode );
-	}
 }
--- a/lib/pages/login-page.js
+++ b/lib/pages/login-page.js
@@ -98,4 +98,11 @@ export default class LoginPage extends AsyncBaseContainer {
 	static getLoginURL() {
 		return dataHelper.getCalypsoURL( 'log-in' );
 	}
+
+	async setGdprCookie() {
+		const setCookieCode = function() {
+			window.document.cookie = 'sensitive_pixel_option=no;domain=wordpress.com';
+		};
+		return await this.driver.executeScript( setCookieCode );
+	}
 }
--- a/lib/pages/wp-home-page.js
+++ b/lib/pages/wp-home-page.js
@@ -12,6 +12,10 @@ export default class WPHomePage extends AsyncBaseContainer {
 		super( driver, by.css( 'body' ), url );
 	}
 
+	async _postInit() {
+		return await this._setGdprCookie();
+	}
+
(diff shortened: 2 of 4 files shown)


alisterscott:
Thanks for your PR @rachelmcr 

I had a look at this and how it can show - it's complicated by the four domains we run our tests against:

1. calypso.localhost:3000
2. calypso.live
3. wpcalypso.wordpress.com / horizon.wordpress.com
4. wordpress.com

It seems the approach in this PR will only work for #4 (wordpress.com) as the cookie is set for that domain. For example, running the reader test against calypso.localhost still shows the banner even though the cookie was set on the calypso.localhost:3000/log-in page

I think we need to do something similar to setting the sandbox cookie where we only do that against wordpress.com itself even when running against another environment, although this may make every test a bit slower as you'd probably want to do it for every test (we only set the sandbox cookie for payment tests)

Maybe you could hook it into `ensureNotLoggedIn( driver ) {` where after it clears the cookies it visits wordpress.com and sets this override cookie? 🤔

lib/pages/wp-home-page.js
```diff
@@ -36,4 +40,11 @@ export default class WPHomePage extends AsyncBaseContainer {
 			`The current url: '${ currentUrl }' does not include ${ target }`
 		);
 	}
+
+	async _setGdprCookie() {
```
alisterscott: Since this is repeated could we move it into a helper file?

lib/pages/wp-home-page.js
```diff
@@ -12,6 +12,10 @@ export default class WPHomePage extends AsyncBaseContainer {
 		super( driver, by.css( 'body' ), url );
 	}
 
+	async _postInit() {
```
alisterscott: This is good since it runs against wordpress.com, but not _every_ sign up test uses this page (only paid sign ups) therefore not every sign up test will have the banner dismissed

lib/pages/login-page.js
```diff
@@ -98,4 +98,11 @@ export default class LoginPage extends AsyncBaseContainer {
 	static getLoginURL() {
 		return dataHelper.getCalypsoURL( 'log-in' );
 	}
+
+	async setGdprCookie() {
+		const setCookieCode = function() {
```
alisterscott: This doesn't seem to work when running against any other environment except wordpress.com (eg. calypso.localhost)

rachelmcr:
>Maybe you could hook it into ensureNotLoggedIn( driver ) { where after it clears the cookies it visits wordpress.com and sets this override cookie? 🤔

Good call. Thanks for reminding me of the other environments and signup tests where that approach wasn't working. I'd considered using `ensureNotLoggedIn( driver )` before but thought it wouldn't be necessary; looking again with these scenarios in mind I think it's the right approach and seems to be working for me with not _too_ much added time.

PR #1396, opened 2018-08-06
Try/jetpack onboarding custom data type

### LINKED ISSUE #1196: Add unique locators for Onboarding pages
Right now Onboarding pages are using `driver.sleep` to wait for page loads. This approach isn't ideal, and we should add some unique locators into pages (in Calypso) so we can get rid of these `sleep` calls.

`driver.sleep` added here: https://github.com/Automattic/wp-e2e-tests/pull/1187

JavonDavis:
Used custom data type as unique locator to reduce the use of sleeps in steps during onboarding tests. Depends on https://github.com/Automattic/wp-calypso/pull/26501.

Closes #1196

### DIFF
--- a/lib/components/wizard-navigation-component.js
+++ b/lib/components/wizard-navigation-component.js
@@ -20,7 +20,7 @@ export default class WizardNavigationComponent extends AsyncBaseContainer {
 	}
 
 	async skipStep( stepNumber ) {
-		await driverHelper.waitTillPresent(
+		await driverHelper.waitTillPresentAndDisplayed(
 			this.driver,
 			By.css( `.wizard__progress-indicator[data-e2e-type="step-indicator-${ stepNumber }"]` )
 		);
--- a/lib/components/wizard-navigation-component.js
+++ b/lib/components/wizard-navigation-component.js
@@ -19,7 +19,11 @@ export default class WizardNavigationComponent extends AsyncBaseContainer {
 		);
 	}
 
-	async skipStep() {
+	async skipStep( stepNumber ) {
+		await driverHelper.waitTillPresent(
+			this.driver,
+			By.css( `.wizard__progress-indicator[data-e2e-type="step-indicator-${ stepNumber }"]` )
+		);
 		return await driverHelper.clickWhenClickable(
 			this.driver,
 			By.css( 'a.wizard__navigation-link .gridicons-arrow-right' )
--- a/lib/driver-helper.js
+++ b/lib/driver-helper.js
@@ -253,23 +253,6 @@ export function unsetCheckbox( driver, selector ) {
 	} );
 }
 
-export function waitTillPresent( driver, selector, waitOverride ) {
-	const timeoutWait = waitOverride ? waitOverride : explicitWaitMS;
-	let self = this;
-
-	return driver.wait(
-		function() {
-			return self.isElementPresent( driver, selector ).then( function( isPresent ) {
-				return isPresent;
-			} );
-		},
-		timeoutWait,
-		`Timed out waiting for element with ${ selector.using } of '${
-			selector.value
-		}' to NOT be present`
-	);
-}
-
 export function waitTillNotPresent( driver, selector, waitOverride ) {
 	const timeoutWait = waitOverride ? waitOverride : explicitWaitMS;
 	let self = this;
--- a/lib/driver-helper.js
+++ b/lib/driver-helper.js
@@ -253,6 +253,23 @@ export function unsetCheckbox( driver, selector ) {
 	} );
 }
 
+export function waitTillPresent( driver, selector, waitOverride ) {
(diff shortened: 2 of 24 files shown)


lib/driver-helper.js
```diff
@@ -253,6 +253,23 @@ export function unsetCheckbox( driver, selector ) {
 	} );
 }
 
+export function waitTillPresent( driver, selector, waitOverride ) {
```
bsessions85: We already have the function `waitTillPresentAndDisplayed` which appears to do the same thing. I swapped it in to test and it works too. Is there a reason to have a different one or was it just overlooked?

JavonDavis: You're right! This was overlooked from my first approach using a hidden input element when I needed to know if it was only present and not displayed, which would have returned false. Since this isn't needed anymore I'll push an update removing this function.

bsessions85:
Good work. When the Calypso PR is in, let's re-run and make sure it all passes and that there aren't any surprises.

lib/components/wizard-navigation-component.js
```diff
@@ -19,7 +19,11 @@ export default class WizardNavigationComponent extends AsyncBaseContainer {
 		);
 	}
 
-	async skipStep() {
+	async skipStep( stepNumber ) {
+		await driverHelper.waitTillPresentAndDisplayed(
+			this.driver,
+			By.css( `.wizard__progress-indicator[data-e2e-type="step-indicator-${ stepNumber }"]` )
```
tyxla: Question: why is this specificity necessary? I'd expect that skipping a step should be the same, regardless of which step we're trying to skip?

JavonDavis: I introduced the step number as a unique identifier for the specific step page we're on... The problem I was having here in the onboarding tests was this error happening occasionally, 

`TimeoutError: Timed out waiting for element with css selector of 'a.wizard__navigation-link .gridicons-arrow-right' to be clickable` 

I think what was causing this is the test not waiting for the step page finish loading before it gave up looking for the right arrow button. Introducing a `driver.sleep` as was done in https://github.com/Automattic/wp-e2e-tests/issues/1196  resulted in not seeing the error at all as expected, but sleeps aren't the best approach so I instead approached it by waiting on  `.wizard__progress-indicator` css selector but then this doesn't tell us with certainty what onboarding step we're on, which led me to implementing it in a way using the step number as a unique id.

tyxla: I understand, thanks for elaborating. 

FWIW, step number sometimes won't be reliable, because we have cases when some steps are omitted from the flow, and in these situations the same step might correspond to different step numbers. So I think it might be a better idea to rely on the `data-e2e-type` attribute of the step instead.

JavonDavis: Thanks for pointing that out I was thinking of ways to work this out without the step number, it'd involve writing more specific `skipStep` functions like `skipContactForm` etc.. but I'm not sure in this case it'd be necessary... The step number here is mapped to a specific step It's more linked to how far we've gotten in the flow(through looking at the Progress indicator), it's designed as a function on the navigation wizard and not a specific page object so I think we can get away with the step number here( get away when skipping, the page objects themselves wait on their respective `data-e2e-type`). 

Keeping in mind though this can become troublesome in cases like [this](https://github.com/Automattic/wp-e2e-tests/blob/try/jetpack-onboarding-custom-data-type/specs-jetpack-calypso/wp-jetpack-onboarding-spec.js#L180) where I've assumed skipping the contact form means skipping step 4 but if for some reason this changes then the number has to be updated in the test but I don't think that's something that changes often since that specific flow is always expected to have the contact form as step 4(?)... Another flow might have it as step 5 and then in that case we can go ahead and pass 5 to the skipStep function.

tyxla: Yeah, your considerations make sense - thanks for clarifying. What I see problematic here is that such approach usually doesn't help make the e2e tests sustainable: every time we change the logic to display or hide a step from the flow, or introduce a new step, or delete an existing step, we might end up breaking e2e tests that don't have anything to do with the functionality that we've introduced - from the developer PoV, I'd probably find that experience confusing.

JavonDavis: Yep, totally get where you're coming from here and why it might seem a little weird that kind of simple change in the flow breaks the tests... The key thing here from the developer PoV I think is like you'd expect, we don't want simple changes to code to go breaking e2e tests and try to focus on testing expected behavior vs specific implementations(this made me a little apprehensive to using step number in the test, but it's a better alternative than sleeping) but I think removing/adding a step to a flow is enough to expect the tests to break(?) and need to be updated since we'd expect users to see that step(or not) and be able to skip it(or not) along with being able to use the actual functionality introduced in that step, if any.

PR #1404, opened 2018-08-09
A/B Test: Update .blog subdomains test

aidvu:
Update to new test name/date.


bsessions85:
The test also needs to be added to the `knownABTestKeys` section

PR #1405, opened 2018-08-09
Remove old A/B test

aidvu:


aidvu:
Test not yet removed. That's why the tests are failing.

bsessions85:
The test should also be removed from `knownABTestKeys`

Issue #956, opened 2018-02-20
Test more login flows


Tug:
Copying the list of cases from our internal trello board to a github issue for more visibility:

We have a lot of possible cases for login: 
- [x] Login with username and password
- [ ] Login with email and password
- [x] Login with email and password and 2fa sms
- [x] Login with email and password and 2fa otp
- [ ] Login with email and password and 2fa backup code
- [x] Login with email and password and 2fa push
- [x] Login with a magic link email for a normal account (/log-in/link)
- [x] Login with a magic link email for a passwordless account
- [x] Login with a magic link email for a passwordless account with 2fa sms
- [x] Login with a magic link email for a passwordless account with 2fa otp
- [ ] Login with google using the popup flow
- [ ] Login with google using the popup flow for an account with 2fa code
- [ ] Login with google using the redirect flow
- [ ] Login with google using the redirect flow for an account with 2fa code
- [ ] Login from woocommerce.com with email and password
- [ ] Login from woocommerce.com with email and password and 2fa code
- [ ] Login from woocommerce.com with a magic link email for a passwordless account
- [ ] Login from woocommerce.com with a magic link email for a passwordless account with 2fa code
- [ ] Login from woocommerce.com with google using the popup flow
- [ ] Login from woocommerce.com with google using the popup flow for an account with 2fa code
- [ ] Login from woocommerce.com with google using the redirect flow
- [ ] Login from woocommerce.com with google using the redirect flow for an account with 2fa code
- [x] Login from Jetpack connect (SSO) with email and password
- [ ] Login from Jetpack connect (SSO) with email and password and 2fa code
- [ ] Login from Jetpack connect (SSO) with a magic link email for a passwordless account
- [ ] Login from Jetpack connect (SSO) with a magic link email for a passwordless account with 2fa code
- [ ] Login page text is correctly translated in another locale than english (/log-in/fr)
- [ ] Login page text is correctly translated in the user's locale (/log-in) when user is logged in
- [ ] Login page text is correctly translated in another locale than english (/log-in/fr) when user is logged in
- [ ] Login page redirect_to external links
- [ ] Login page redirect_to internal (calypso) links
- [ ] Login page redirect_to after google connect
- [ ] Login page redirect_to after magic link
- [ ] Login page redirect_to after 2fa sms

alisterscott:
Hi @Tug 

[...]

PR #1407, opened 2018-08-10
Update to the two factor authentication tests

JavonDavis:
The 2fa authentication tests weren't updated when the project moved to the async/await style, uncommenting the tests and running them led to errors because of this.. Updated the tests to now run when uncommented using async/await.... The existing problem mentioned in #956 still exists though so these tests will be commented out again before merging while I continue my research spike around this.

**NB:** I was unable to locally test '+2fa-push -passwordless' due to no account with those features...


lib/flows/login-flow.js
```diff
@@ -49,7 +49,10 @@ export default class LoginFlow {
 		}
 	}
 
-	async login( { jetpackSSO = false, jetpackDIRECT = false, screenshot = false } = {}, eyes ) {
+	async login(
+		{ emailSSO = false, jetpackSSO = false, jetpackDIRECT = false, screenshot = false } = {},
+		eyes
```
Stojdza: Maybe you will have some merge conflicts since eyes helper will be removed. I'm not sure is it on master yet, but try to merge master first and see what will happen.

specs/wp-log-in-out-spec.js
```diff
@@ -110,302 +109,336 @@ describe( `[${ host }] Authentication: (${ screenSize }) @parallel @jetpack @vis
...
-	// 			} );
-	// 		} );
-	// 	} );
-	// }
+	if (
+		dataHelper.hasAccountWithFeatures( '+2fa-sms -passwordless' ) &&
+		! dataHelper.isRunningOnLiveBranch()
+	) {
+		describe( 'Can Log in on a 2fa account', function() {
+			let loginFlow, twoFALoginPage, twoFACode;
```
Stojdza: Tests are failing for me locally on this step - text set in input field is `1234567undefined` and because of that test cannot proceed.

alisterscott:
I've lost track of where this one is at - are we keeping it?

JavonDavis:
> I've lost track of where this one is at - are we keeping it?

With this change, it wouldn't run unless the PR has the specific secure Auth label so we shouldn't have the same problem as before, seems I'll need to resolve some conflicts from master once I do that it should be good to go.

JavonDavis:
I've done the merge if someone from @Automattic/flowpatrol has time to check this out locally again I can merge this in

JavonDavis:
@Stojdza No there isn't but it seems other login tests might have been running at the same time you were testing(probably me testing at the same time) so the session seemed to expire. That's why the goal of this was to not run it so often, do you mind trying again in about 30 minutes?

Stojdza:
> do you mind trying again in about 30 minutes?

13hrs passed, so I gave it a shot and it works :)

PR #1410, opened 2018-08-12
Included tests for coupons feature 

### LINKED ISSUE #1303: Need e2e test around coupon
Create a coupon for like 1c off a plan and then we can use it for e2e tests to make sure coupons don't break

JavonDavis:
* Included coupon test for Plan purchase on existing account
* Included coupon test for sign up with premium plan

To test:
* Include "testCouponCode" in local config with the coupon code shared in Slack.

Fixes #1303 

Depends on https://github.com/Automattic/wp-calypso/pull/26652


alisterscott:
I've requested some changes on the related wp-calypso change

lib/components/secure-payment-component.js
```diff
@@ -117,6 +117,67 @@ export default class SecurePaymentComponent extends AsyncBaseContainer {
...
 	}
 
+	async clickCouponButton() {
+		await driverHelper.waitTillPresentAndDisplayed(
+			this.driver,
+			By.css( 'button.cart__toggle-link' )
+		);
+
+		// 2 elements have this locator on the page including the mobile element
+		let elements = await this.driver.findElements( By.css( 'button.cart__toggle-link' ) );
```
alisterscott: I think it would be best here to have two selectors, one for mobile and another for desktop which both return a single element (something like `.payment-box__content button.cart__toggle-link`) - that way there's no need to loop - it's easy to know whether the test is mobile or desktop (`driverManager.currentScreenSize();`)

lib/components/secure-payment-component.js
```diff
@@ -117,6 +117,67 @@ export default class SecurePaymentComponent extends AsyncBaseContainer {
...
+
+	async cartTotalAmount() {
+		await driverHelper.waitTillPresentAndDisplayed( this.driver, By.css( '.cart-total-amount' ) );
+		let cartText = await this.driver.findElement( By.css( '.cart-total-amount' ) ).getText();
+		let amountMatches = cartText.match( /\d+\.?\d*/g );
+		return await parseFloat( amountMatches[ 0 ] );
+	}
+
+	async applyCoupon() {
+		return await driverHelper.clickWhenClickable( this.driver, By.css( '#apply-coupon' ) );
```
alisterscott: needs updating with new wp-calypso data attribute

lib/components/secure-payment-component.js
```diff
@@ -117,6 +117,67 @@ export default class SecurePaymentComponent extends AsyncBaseContainer {
...
+	}
+
+	async enterCouponCode( couponCode ) {
+		await this.clickCouponButton();
+		await driverHelper.setWhenSettable( this.driver, By.css( '#coupon-code' ), couponCode );
```
alisterscott: needs updating

lib/components/secure-payment-component.js
```diff
@@ -117,6 +117,67 @@ export default class SecurePaymentComponent extends AsyncBaseContainer {
...
+		return await this.applyCoupon();
+	}
+
+	async removeCoupon() {
+		await driverHelper.waitTillPresentAndDisplayed( this.driver, By.css( '.cart__remove-link' ) );
+		let elements = await this.driver.findElements( By.css( '.cart__remove-link' ) );
+		// 2 of these elements on page
+		for ( let i = 0; i < elements.length; i++ ) {
```
alisterscott: same as above, use two selectors

lib/components/secure-payment-component.js
```diff
@@ -117,6 +117,67 @@ export default class SecurePaymentComponent extends AsyncBaseContainer {
...
+			let element = elements[ i ];
+			element.click().then(
+				function() {
+					return true;
```
alisterscott: not necessary if you have one selector per screen size

specs/wp-signup-spec.js
```diff
@@ -305,6 +305,216 @@ describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function() {
...
+			let testEnvironment = 'WordPress.com';
+			let testName = `Signup [${ global.browserName }] [${ screenSize }]`;
+			eyesHelper.eyesOpen( driver, eyes, testEnvironment, testName );
+		} );
+
+		before( async function() {
+			return await driverManager.ensureNotLoggedIn( driver );
+		} );
+
+		step( 'We can set the sandbox cookie for payments', async function() {
```
alisterscott: coupons won't work sandboxed - so we shouldn't set the sandbox, but we won't also complete the purchase

specs/wp-signup-spec.js
```diff
@@ -305,6 +305,216 @@ describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function() {
...
+			originalCartAmount = await securePaymentComponent.cartTotalAmount();
+
+			await securePaymentComponent.enterCouponCode( dataHelper.getTestCouponCode() );
+
+			let newCartAmount = await securePaymentComponent.cartTotalAmount();
+			let expectedCartAmount = parseFloat( ( originalCartAmount * 0.99 ).toFixed( 2 ) );
+			assert.strictEqual( newCartAmount, expectedCartAmount, 'Coupon not applied properly' );
+		} );
+
+		step( 'Can enter and submit test payment details', async function() {
```
alisterscott: Since we're not using a sandboxed mode, we can remove the rest of this test and just let the test finish at the coupon in the cart step

lib/components/secure-payment-component.js
```diff
@@ -150,16 +140,24 @@ export default class SecurePaymentComponent extends AsyncBaseContainer {
...
 				By.css( '.checkout__summary-toggle' )
 			);
 		}
 	}
 
 	async cartTotalAmount() {
+		let cartElement = await this.driver.findElement( By.css( '.cart-total-amount' ) );
+
+		if ( currentScreenSize() === 'mobile' ) {
+			await this.driver.executeScript( 'arguments[0].scrollIntoView()', cartElement );
```
JavonDavis: Had some trouble with the cart not coming into view on mobile causing other steps to fail. Explicitly bringing into view resolved that for me

JavonDavis:
Automattic/wp-calypso#26652 Already merged so this PR is no longer blocked.

specs/wp-plan-purchase-spec.js
```diff
@@ -66,4 +67,68 @@ describe( `[${ host }] Plans: (${ screenSize }) @parallel @jetpack`, function()
...
+
+		step( 'Can Remove Coupon', async function() {
+			const planCheckoutPage = await PlanCheckoutPage.Expect( driver );
+
+			await planCheckoutPage.removeCoupon();
+
+			// Needs a little time for the coupon changes to take effect
+			// Had a driver.sleep( 1000 ) here but it might it seems the time it takes to attempt to toggle twice covers it
+			await planCheckoutPage.toggleCartSummary();
+			await planCheckoutPage.toggleCartSummary();
```
Stojdza: I think that because of this notice, you need to `toggleCartSummary()` twice. If you could manage to dismiss notice maybe you would need to do it only once? 
screen shot 2018-10-28 at 23 07 12

Stojdza:
If test `Viewing a specific plan with coupon:` breaks after adding the coupon and before removing it, every next time when I run the test it will fail - because the coupon is already applied from the first run. Was this expected?

lib/components/secure-payment-component.js
Stojdza: Does [`scrollIntoView()`](https://github.com/Automattic/wp-e2e-tests/blob/master/lib/driver-helper.js#L414) from _driver-helper.js_ can do the same thing as `executeScript()`?

JavonDavis: Yep! I'll push a change with the driverHelper function instead.

specs/wp-plan-purchase-spec.js
JavonDavis: There's no notice that appears after removing a coupon, so that's why I decided to toggle here, to give a little time for the change to happen. But I removed them and it seems to work fine now. 
Thanks for pointing this notice out though I included dismissing it into the step after applying the coupon.

JavonDavis:
> If test `Viewing a specific plan with coupon:` breaks after adding the coupon and before removing it, every next time when I run the test it will fail - because the coupon is already applied from the first run. Was this expected?

Totally correct! Didn't account for this, I'm pushing a change that will check for and remove any already applied coupons at plan checkout

PR #1414, opened 2018-08-15
Update node version based in wrapper repos

bsessions85:
This will update the node version variable in the wrapper repos so that we only have to change it once.

The variable will only be updated if the .nvmrc file has changed


scripts/circleci-branch-update.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+update-wrapper-node-version () {
```
alisterscott: Should only call this for master branch

scripts/circleci-branch-update.sh
```diff
@@ -0,0 +1,30 @@
...
+    version=$(<.nvmrc)
+    version=${version:1}
+    config=( $(jq -r '.circleCIToken' ./config/local-${NODE_ENV}.json) )
+    token=${config[0]}
+
+    curl --write-out %{http_code} --silent --output /dev/null -X POST --header "Content-Type: application/json" -d "{\"name\":\"NODE_VERSION\", \"value\":\"$version\"}" https://circleci.com/api/v1.1/project/github/automattic/wp-e2e-tests-for-branches/envvar?circle-token=${token}
```
alisterscott: Maybe add a note to the README under wrapper repo's stating to add them here if you add a new one

bsessions85:
Updated to only use master branch and added note to documentation about updating the script as well

PR #1416, opened 2018-08-16
Change the signup spec as we're going to rollout the checklist to other goal types

taggon:
We no longer need to separate tests into the ones for blog and non-blog sites.

Related: https://github.com/Automattic/wp-calypso/pull/26721


taggon:
Rebased and passed all tests.

PR #1418, opened 2018-08-16
Adding config values for API application

tammullen:
The important one is the access code that will allow us to make API calls on behalf of our current default user. It is a global code that will allow access all blogs that the user has access.


rachelmcr:
👍 If we end up adding these API calls to our tests, let's remember to add details about this (the config setup and how to get the access code) to our [config values doc](https://github.com/Automattic/wp-e2e-tests/blob/master/docs/config.md#config-values).

PR #1419, opened 2018-08-21
Fix activity log test

### LINKED ISSUE #1413: Re-enable Activity Log Test
When A/L is available again re-enable this test

alisterscott:
Re-enables the activity log test

Updates the selector that finds the activity log entry for publishing a post

Closes #1413

### DIFF
--- a/lib/pages/stats/activity-page.js
+++ b/lib/pages/stats/activity-page.js
@@ -12,18 +12,14 @@ export default class ActivityPage extends AsyncBaseContainer {
 
 	async postTitleDisplayed( postTitle ) {
 		const driver = this.driver;
-		return driver.wait( () => {
-			driver.navigate().refresh();
-			return driverHelper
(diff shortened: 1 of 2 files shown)


lib/pages/stats/activity-page.js
```diff
@@ -18,7 +18,7 @@ export default class ActivityPage extends AsyncBaseContainer {
 				.isElementPresent(
 					driver,
 					By.xpath(
-						`//div[@class='activity-log-item__description-content']//a//em[text()='${ postTitle }']`
+						`//div[@class='activity-log-item__description-content']//a[text()='${ postTitle }']`
 					)
 				)
 				.then( present => {
```
Stojdza: Should we consider and put async instead `.then` in this method?

alisterscott: Good idea - fixed 👍

Issue #1424, opened 2018-08-22
Remove Visual-Diff Tests from Codebase


rachelmcr:
To do:
- [x] Remove screenshot steps from test specs
- [x] Remove eyes helper
- [x] Remove from run.sh and magellan
- [x] Remove from config
- [x] Remove related documentation
- [x] Update [Calypso A/B test docs](https://github.com/Automattic/wp-calypso/blob/master/client/lib/abtest/README.md#updating-our-end-to-end-tests-to-avoid-inconsistencies-with-ab-tests)
- [x] Remove [visdiff wrapper repo](https://github.com/Automattic/wp-e2e-tests-visdiff)

alisterscott:
Thanks for completing this 🙏

PR #1425, opened 2018-08-22
Remove visdiffs from test specs

rachelmcr:
The first step to removing the visual diff tests from our codebase (https://github.com/Automattic/wp-e2e-tests/issues/1424). This removes the visual diff steps from the individual test specs.

To test:
1. Run `./run.sh -v` to confirm that no tests are run (no tests tagged `@visdiff`).
2. Confirm that other tests run as expected (e.g. with `./run.sh -g`).


rachelmcr:
Thanks for the review!

> Is it ok that I'm getting message Error: no tests found when there are no tests?

Yep, that's expected since there are no longer any tests tagged `@visdiff`. I'll remove the other pieces in a followup PR but wanted to keep this one focused on the test specs.

Issue #1393, opened 2018-08-06
Theme test sometimes fails


alisterscott:
The theme test sometimes fails, 

https://21850-57936731-gh.circle-artifacts.com/0/home/circleci/wp-e2e-tests/screenshots/FAILED-EN-MOBILE-can-see-theme-details-page-and-open-the-live-demo-1533529831243.png

https://circleci.com/gh/Automattic/wp-e2e-tests/21850#artifacts/containers/0

```
  1) [WPCOM] Switching Themes: (mobile) Switching Themes @parallel @jetpack @visdiff Can switch free themes Can see theme details page and open the live demo:
04:30:31      TimeoutError: Timed out waiting for element with css selector of 'a.theme__sheet-preview-link,li.is-theme-preview a' to be clickable
04:30:31 Wait timed out after 20025ms
04:30:31       at /home/circleci/wp-e2e-tests/node_modules/selenium-webdriver/lib/promise.js:2201:17
04:30:31       at process._tickCallback (internal/process/next_tick.js:68:7)
```

I believe this is due to https://github.com/Automattic/wp-calypso/issues/25720

rachelmcr:
In Automattic/wp-calypso#25720 it looks like the button not being present is because the theme is active on the site already. Note that in your second screenshot in that issue there's a "Customize site" button (which shows when you view the details for your site's active theme) at the top of the page instead of an "Activate this design" button.

Could this be happening because our theme switch tests are running in parallel? (That is, while one test is opening the theme details another test is activating that theme on the site.)

I'll watch for this failure to pop up again — now that we're capturing videos for failed tests, we could verify if that's happening when this test fails.

PR #1428, opened 2018-08-23
Shorten theme test to avoid theme switch collisions

rachelmcr:
Fixes: https://github.com/Automattic/wp-e2e-tests/issues/1393

The theme tests have been sporadically failing because they run in parallel and the theme switch in one test affects the other test. This PR fixes that issue with the following changes to the spec:

- Changes the `Switching Themes` test to `Previewing Themes`.
- Previews a "Twenty S..." theme (instead of a "Twenty F..." theme, one of which will be active on the site).
- Removes the activation step.
- Stops once we confirm the live demo appears as expected with an option to activate the theme from there.

This does also remove the check for being able to go back to the theme detail page after activation, but that seems like a minor step in this flow and overall resolves the problem we've been seeing with failures in this spec.

Feedback welcome on this approach versus other possibilities to resolve the issue with these two theme tests running in parallel.

### DIFF
--- a/lib/pages/theme-detail-page.js
+++ b/lib/pages/theme-detail-page.js
@@ -12,7 +12,7 @@ export default class ThemeDetailPage extends AsyncBaseContainer {
 	async openLiveDemo() {
 		return await driverHelper.clickWhenClickable(
 			this.driver,
-			By.css( '.theme__sheet-screenshot' )
+			By.css( 'a.theme__sheet-preview-link,li.is-theme-preview a' )
 		);
 	}
 
--- a/lib/pages/theme-preview-page.js
+++ b/lib/pages/theme-preview-page.js
@@ -17,6 +17,10 @@ export default class ThemePreviewPage extends AsyncBaseContainer {
 		);
 	}
 
+	async activateButtonVisible() {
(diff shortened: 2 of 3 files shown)


alisterscott:
Sorry I didn't get to review this one today - I'll jump on it Monday morning if someone else hasn't looked at it

alisterscott:
This looks good - we still have a test that switches themes in the same spec (and in the multisite spec) - is it ok to keep this since these run in parallel (desktop and mobile) and across different builds - do you think this is still an issue?

rachelmcr:
>we still have a test that switches themes in the same spec (and in the multisite spec) - is it ok to keep this since these run in parallel (desktop and mobile) and across different builds - do you think this is still an issue?

The theme switch test in this spec won't cause problems for the multisite spec, since those specs use different test accounts/sites. It's possible that the runs across different screen sizes/builds could cause problems, but since it's such a precise timing issue (a theme switch from one test happening after a theme is selected but before it's activated in another test) I think it's less likely to occur with only one switch per site.

I'll merge this and watch the results this week — if it's still happening we could remove the theme switch altogether.

PR #1430, opened 2018-08-27
Replace nudgeAPalooza 5-way test with 4 smaller 2-way tests

adamziel:
This PR adds a set of tests that nudgeAPalooza was split into. After initial round of testing, we need to adjust course a bit and perform a few detailed 50/50 tests instead of one large test. More details in comments under this post: p8Eqe3-sy-p2.

Related PRs:

https://github.com/Automattic/wp-calypso/pull/26868
https://github.com/Automattic/wp-calypso/pull/26869
https://github.com/Automattic/wp-calypso/pull/26873
https://github.com/Automattic/wp-calypso/pull/26877

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -78,6 +78,9 @@
 		[ "aboutSuggestionMatches_20180704", "control" ],
 		[ "includeDotBlogSubdomainV2_20180813", "no" ],
 		[ "gSuiteDiscountV2_20180822", "control" ],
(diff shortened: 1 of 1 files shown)


rachelmcr:
@adamziel Can you also add these tests to `knownABTestKeys` in the same `config/default.json` file?

adamziel:
@rachelmcr on that!

PR #1433, opened 2018-08-28
Re-enable sign up activation emails in sign up tests

### LINKED ISSUE #819: Update signup spec for email verification changes
The timing for new account verification emails is currently being tested (ref: p91bmw-da-p2), which broke the signup tests that relied on it (see #817).

Once the timing for those emails is finalized, we should update the signup spec accordingly:

- Reenable the check in `Sign up for a free site and log in via a magic link` for the number of expected emails to a new user.
- Possibly add a new test for new users that verifies the account email and publishes a post, taking into account the new timing for the verification email.

alisterscott:
We no longer delay any sign up emails

Fixes #819

**Testing instructions**

Make sure the new scenario works desktop and mobile

### DIFF
--- a/lib/pages/editor-page.js
+++ b/lib/pages/editor-page.js
@@ -329,9 +329,9 @@ export default class EditorPage extends AsyncBaseContainer {
 	}
 
 	async publishEnabled() {
-		return await driverHelper.isEventuallyPresentAndDisplayed(
+		return await driverHelper.isElementPresent(
 			this.driver,
-			'.editor-publish-button:not([disabled])'
+			by.css( '.editor-publish-button:not([disabled])' )
 		);
 	}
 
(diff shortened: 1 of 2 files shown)


lib/pages/editor-page.js
```diff
@@ -325,9 +329,9 @@ export default class EditorPage extends AsyncBaseContainer {
 	}
 
 	async publishEnabled() {
-		return await driverHelper.isEventuallyPresentAndDisplayed(
+		return await driverHelper.isElementPresent(
 			this.driver,
-			'.editor-publish-button:not([disabled])'
+			by.css( '.editor-publish-button:not([disabled])' )
```
alisterscott: The previous selector wasn't valid so would have never worked 🤔

PR #1434, opened 2018-08-28
Fix the signup tests to include a canary

alisterscott:
This was accidently removed recently - adding it back


alisterscott:
@Automattic/flowpatrol merging this one to get the canary signup running again (PRs and on merge)

PR #1436, opened 2018-08-28
Update/restore nudge a palooza

adamziel:
Restores previously removed nudgeAPalooza A/B test until we remove it from the product repo


bsessions85:
Thanks @adamziel! Merging this in.

PR #1437, opened 2018-08-28
Update/add nudge tests to known tests

adamziel:
It's a follow up on the following PR: https://github.com/Automattic/wp-e2e-tests/pull/1430

PR #1438, opened 2018-08-28
Fix  known test keys

bsessions85:
The datestamp was included in known test keys in #1437. Removing it.


bsessions85:
Merging to clean up test noise

PR #1440, opened 2018-08-28
Fixing broken account cleanup steps

### LINKED ISSUE #1435: Investigate warnings in Slack
```
There was an error in the hooks that clean up the test account but since it is cleaning up we really don't care: 'TimeoutError: Timed out waiting for element with css selector of 'header.current-section a' to be clickable
Wait timed out after 20028ms' - encountered on 'chrome' at screen size 'mobile' in CircleCI build #22339 on branch 'fix/missing-canary'
```

We need to check this is still working

bsessions85:
Closes #1435 

Fixes 2 issues with cleaning up accounts. 

To Test:
Ensure that no warnings related to the hooks that clean up the test accounts are output to slack when run in CircleCI.
Ensure that cleanup tasks run all steps when run locally.


alisterscott:
Merging to reduce noise

PR #1449, opened 2018-08-30
Add readerSearchPlaceholder to known tests

bluefuton:
This is a new A/B test to be added in https://github.com/Automattic/wp-calypso/pull/26969 (copy test only).


alisterscott:
Thanks for the update - this can be merged any time - before the other PR is fine 👍

PR #1451, opened 2018-08-30
Renable domains transfer test

alisterscott:
Re-enables the domains transfer test which was disabled recently due to timeouts


alisterscott:
Passed straight away on Circle so merging this to re-enable it

PR #1456, opened 2018-09-04
Add reader AB test override

alisterscott:
Looks like we need to override it after all


alisterscott:
@Automattic/flowpatrol merging this one to reduce the messages posted in Slack

PR #1457, opened 2018-09-04
Register .live domain

### LINKED ISSUE #830: Update add domain to existing site to register (and cancel) .live domain
We can use the sandbox method for adding a fake .live domain to an existing site and purchase it then refund and cancel it.

Stojdza:
Adding test for #830. Existing user should purchase _.live_ domain, and after that cancel and refund it. 

Resolves #830

### DIFF
--- a/specs/wp-manage-domains-spec.js
+++ b/specs/wp-manage-domains-spec.js
@@ -1,9 +1,11 @@
 /** @format */
 
 import config from 'config';
+import assert from 'assert';
 import * as driverManager from '../lib/driver-manager.js';
 import * as dataHelper from '../lib/data-helper.js';
 
+import WPHomePage from '../lib/pages/wp-home-page.js';
 import DomainsPage from '../lib/pages/domains-page.js';
 import CheckOutPage from '../lib/pages/signup/checkout-page.js';
 import ReaderPage from '../lib/pages/reader-page.js';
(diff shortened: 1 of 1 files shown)


specs/wp-manage-domains-spec.js
```diff
@@ -283,4 +291,189 @@ describe( `[${ host }] Managing Domains: (${ screenSize }) @parallel`, function(
 			return await TransferDomainPrecheckPage.Expect( driver );
 		} );
 	} );
+
+	describe.only( 'Register (and cancel) .live domain to an existing site ', function() {
```
Stojdza: This `.only` is intentional, so we can run only this test with default test user.

specs/wp-manage-domains-spec.js
```diff
@@ -283,4 +291,189 @@ describe( `[${ host }] Managing Domains: (${ screenSize }) @parallel`, function(
...
+				);
+				return this.skip();
+			}
+		} );
+
+		before( async function() {
+			return await driverManager.ensureNotLoggedIn( driver );
+		} );
+
+		step( 'We can visit set the sandbox cookie for payments', async function() {
```
Stojdza: Sandbox cookie is used like on several different places, but I'm not sure is it set properly to purchase `.live` domain. 
As you can see in the test run purchasing `.live` domain is passing. But when user want to cancel and refund it there is no option for that. 

screen shot 2018-09-04 at 16 20 33

alisterscott:
@Stojdza it looks like it's being set correctly as the sandbox payment is passing. I'm not sure why the cancel option doesn't appear. I can see the domain in S.A as long as I am in sandboxed mode. Maybe as in #domains-dev ? Maybe we don't need to worry about cancelling these?

Stojdza:
@alisterscott I've tried with #domains-dev but they didn't know why is this happening. We wanted here to close the complete loop - adding, canceling and refund. Since these steps are covered separately in other tests and too much time is lost here, maybe it's not urgent and we can leave it for HACK week or meetup?

alisterscott:
> Since these steps are covered separately in other tests and too much time is lost here, maybe it's not urgent and we can leave it for HACK week or meetup?

sounds good, feel free to close it

PR #1459, opened 2018-09-05
Remove page refresh when a/b test values are set

bsessions85:
The page refresh was causing tests to fail on [this PR](https://github.com/Automattic/wp-calypso/pull/26942). Removing the refresh causes the tests to pass and all other tests pass as well. 

In the issue above, the query string was removed during a new redirect. When the page is reloaded without the query string it doesn't know the context anymore and since the url we saved doesn't have it, it causes a different redirect. 

TO TEST:
Run test suite and ensure that all pass and that there aren't issues with expected A/B tests not matching actual

### DIFF
--- a/lib/async-base-container.js
+++ b/lib/async-base-container.js
@@ -100,8 +100,6 @@ export default class AsyncBaseContainer {
 	async setABTestControlGroupsInLocalStorage() {
 		const overrideABTests = config.get( 'overrideABTests' );
 
-		const startUrl = await this.driver.getCurrentUrl();
-
 		const expectedABTestValue = overrideABTests
 			.map( entry => {
 				return '"' + entry[ 0 ] + '":"' + entry[ 1 ] + '"';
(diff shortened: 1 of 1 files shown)


alisterscott:
@Automattic/flowpatrol: I'm trying to think why we did this.

I believe it's because say you start on wordpress.com/start you then set the AB tests. If you don't then refresh that page it will mean that if there's an active AB test on that page you may experience inconsistency - like sometimes the start page displays different options depending on AB tests.

I'm happy to leave this change in place but just remember that if you see something strange happening on the start page (or any page where we call this method) then this is probably the reason.

alisterscott:
This is causing issues for this PR: https://github.com/Automattic/wp-calypso/pull/27254 as visiting wordpress.com/start may show the user sign up page which isn't expected

I've taken a slightly different approach in: https://github.com/Automattic/wp-e2e-tests/pull/1488

We visit wordpress.com/themes, set the AB tests, then visit wordpress.com/start to see the expected start page

PR #1467, opened 2018-09-10
Update password in config file

brbrr:
We need to update config encryption password. Right after this was merged we need to update all the env configs for wrapper repos.
env config example: https://circleci.com/gh/Automattic/wp-e2e-tests/edit#env-vars
list of wrapper repos: https://github.com/Automattic/wp-e2e-tests/blob/master/docs/miscellaneous.md#list-of-wrapper-repos--friends

To test: 
- Get a CONFIG_KEY from SS via searching for `E2E Config decode key`
- decode a config with `npm run decryptconfig`


alisterscott:
Thanks for wrangling this

PR #1468, opened 2018-09-12
Skip domain transfer - very flaky

alisterscott:
Times out often causing inconsistent test results


alisterscott:
Merging to stop the failing tests on master

PR #1474, opened 2018-09-14
Themes: Update Theme Thumbnail Selector

ockham:
Companion PR to https://github.com/Automattic/wp-calypso/pull/27169

PR #1475, opened 2018-09-14
Give an option for thumbnail

bsessions85:
Thumbnail tests have been failing because the selector was changed. Setting it to find one or the other

Related to https://github.com/Automattic/wp-e2e-tests/pull/1474


alisterscott:
> Our best guess is that the element that was supposed to be clicked contains only absolute positioned contents and is 0 x 0. It's likely that this doesn't qualify as an element that can be clicked on.

Yes, since it's sized to 0,0 it can't be clicked as it doesn't really appear visible

ockham:
Oops, looks like I caused this with https://github.com/Automattic/wp-calypso/pull/27169. Sorry about that, and thanks for fixing!

PR #1481, opened 2018-09-17
Fix test run videos

bsessions85:
The tests are showing just a black screen. It seems to be related to a change that added spaces in the brackets for a variable the screen recording was using. I changed it back and added a note that it should not be adjusted. 

I added a broken test in https://github.com/Automattic/wp-e2e-tests/commit/364a42a6359ef47ec4ecea833d04a14a7ea02019 to verify the fix.

### DIFF
--- a/lib/driver-manager.js
+++ b/lib/driver-manager.js
@@ -145,7 +145,8 @@ export function startBrowser( { useCustomUA = true, resizeBrowserWindow = true }
 				}
 
 				if ( global.displayNum ) {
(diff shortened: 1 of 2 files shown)


alisterscott:
Sorry about that, it was my change that did that. It was actualy prettier auto formatting that file and any change to that file will make it happen again.

I added a prettier ignore comment to the line so prettier won't break it again.

See: https://github.com/Automattic/wp-e2e-tests/pull/1485

PR #1482, opened 2018-09-17
Simulate browser Back button during sign up

### LINKED ISSUE #832: Increase coverage of back button use during sign up
Change or add e2e test coverage so it would find an issue like this: https://href.li/?https://github.com/Automattic/wp-calypso/pull/20127

Using the back button during sign up should end up with a good result

Stojdza:
Regarding request #832 we needed to simulate browser 'Back' button during sign up flow. Two options were here:
1. Add completely new _sign up_ test with 'Back' button option and see is it working
2. Simulate 'Back' in some existing _sign up_ test

I choose second one since tests execution is longer for only 1-2 seconds.

Steps added:
- select store during sign up
- submit form
- press browser's 'Back' button
- unselect store

Test should continue execution as usual.

Fixes #832

### DIFF
--- a/lib/pages/signup/about-page.js
+++ b/lib/pages/signup/about-page.js
@@ -43,4 +43,30 @@ export default class AboutPage extends AsyncBaseContainer {
 			await driverHelper.setCheckbox( this.driver, By.css( '#promote' ) );
 		}
 	}
+
+	async unsetCheckBox( {
+		showcase = false,
+		share = false,
+		sell = false,
+		educate = false,
+		promote = false,
+	} ) {
+		// Wait before click and unset checkbox
+		await driverHelper.waitForFieldClearable( this.driver, By.css( '#siteTitle' ) );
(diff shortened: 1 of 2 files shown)


lib/pages/signup/about-page.js
```diff
@@ -43,4 +43,30 @@ export default class AboutPage extends AsyncBaseContainer {
...
+		share = false,
+		sell = false,
+		educate = false,
+		promote = false,
+	} ) {
+		// Wait before click and unset checkbox
+		await driverHelper.waitForFieldClearable( this.driver, By.css( '#siteTitle' ) );
+		if ( showcase === true ) {
+			await driverHelper.unsetCheckbox( this.driver, By.css( '#showcase' ) );
+		}
```
bsessions85: Could you just pass a selector or a field id and then unset whatever value is set?  This function seems a bit more elaborate than it needs to be.

PR #1483, opened 2018-09-17
Use subdomain format for livebranches

bsessions85:
Uses the new subdomain format of calypso.live URLs


bsessions85:
Merging in to get branch tests working

PR #1488, opened 2018-09-19
Add support for avoiding user first sign up flow

alisterscott:
This makes sure we can override the user first sign up flow in https://github.com/Automattic/wp-calypso/pull/27254

Since we don't refresh the start page after setting the AB test groups this means the user first test may show and then the tests will fail

What this does instead:

1. Visits wordpress.com/themes (or whichever env it is using) and sets AB test groups
2. Then visits wordpress.com/start - this will redirect to appropriate start page - this may be user first sign up if the flow is set that way

**To Test**

Run against https://github.com/Automattic/wp-calypso/pull/27254 and make sure sign ups work and we don't encounter the user first sign up flow

### DIFF
--- a/lib/pages/signup/start-page.js
+++ b/lib/pages/signup/start-page.js
@@ -8,8 +8,8 @@ import AsyncBaseContainer from '../../async-base-container';
 
 export default class StartPage extends AsyncBaseContainer {
 	constructor( driver, url ) {
-		const loggedOutRootURL = dataHelper.getCalypsoURL();
-		super( driver, By.css( '#wpcom' ), loggedOutRootURL );
+		const loggedOutThemesURL = dataHelper.getCalypsoURL( 'themes' );
+		super( driver, By.css( '#wpcom' ), loggedOutThemesURL );
 		this.startURL = url;
 	}
 
--- a/lib/pages/signup/start-page.js
+++ b/lib/pages/signup/start-page.js
@@ -3,15 +3,20 @@
 import { By } from 'selenium-webdriver';
 
 import * as dataHelper from '../../data-helper';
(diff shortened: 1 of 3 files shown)


lib/pages/signup/start-page.js
```diff
@@ -3,15 +3,20 @@
 import { By } from 'selenium-webdriver';
 
 import * as dataHelper from '../../data-helper';
+import * as driverHelper from '../../driver-helper';
 import AsyncBaseContainer from '../../async-base-container';
 
 export default class StartPage extends AsyncBaseContainer {
 	constructor( driver, url ) {
-		super( driver, By.css( '.step-wrapper' ), url );
+		const loggedOutThemesURL = dataHelper.getCalypsoURL( 'themes' );
```
alisterscott: this is a SSR page that works logged out - I guess we could also use the log-in page

lib/pages/signup/start-page.js
```diff
@@ -3,15 +3,20 @@
...
+		super( driver, By.css( '#wpcom' ), loggedOutThemesURL );
+		this.startURL = url;
 	}
 
 	async _postInit() {
-		return await this.setABTestControlGroupsInLocalStorage();
+		await this.setABTestControlGroupsInLocalStorage();
```
alisterscott: We set the AB test control groups then load the start page

Stojdza:
@alisterscott What do you think about a solution to put back refresh in `setABTestControlGroupsInLocalStorage()`, but set it to false by default?  And where it's needed to call refresh with true. Or otherwise - to be true by default, and where it's not needed or causing problems, set it to false. 

I'm not sure is it applicable to all tests, but I think that way we would skip calling `/themes` page for every test.

PR #1489, opened 2018-09-19
Scroll Viewers

### LINKED ISSUE #1479: Improve removing invited user
At some point, test `Inviting New User as a Viewer of a WordPress.com Private Site: @parallel` started to fail because there are too many _Viewers_ for a private site ([example](https://circleci.com/gh/Automattic/wp-e2e-tests/22750)). I tried to reproduce it locally and found that if there are 21 or more viewers test will fail (both, desktop and mobile). Also, it started to fail on [CircleCI](https://circleci.com/gh/Automattic/wp-e2e-tests/22760). 

It looks like that invited user is not on the list, but it is and it's at the bottom of the list. I assume that we have some problem with scrolling and getting that element. 

In test, we have `removeUserByName()` in two places - in [step](https://github.com/Automattic/wp-e2e-tests/blob/master/specs/wp-invite-users-spec.js#L284) and in [after](https://github.com/Automattic/wp-e2e-tests/blob/master/specs/wp-invite-users-spec.js#L309). @alisterscott I'm guessing that second one is here just to be sure that user will be removed?

Stojdza:
We have an [edge case](https://circleci.com/gh/Automattic/wp-e2e-tests/22836) where removing Viewers from the site didn't work if there were more than 20 of them (even we are removing Viewer in two places - in step and in after block, it's happening that we somehow gather more than 20 Viewers). It's an infinite list of Viewers, and they didn't load all at once. This will also cause a problem if, for example, we want to check that the user is removed and there are 50+ Viewers - we would get the false positive result (if user is the last item in the list). 

With this fix, Viewers list will be scrolled depending on their number, so we'll be sure that user is removed or present (depends on the test).

**To test:** TBD

Fixes #1479

### DIFF
--- a/lib/driver-helper.js
+++ b/lib/driver-helper.js
@@ -408,11 +408,11 @@ export async function refreshIfJNError( driver, timeout = 2000 ) {
 	return await refreshIfNeeded();
 }
 
-export async function scrollIntoView( driver, selector ) {
+export async function scrollIntoView( driver, selector, block = 'center', inline = 'center' ) {
 	let selectorElement = await driver.findElement( selector );
 
 	return await driver.executeScript(
-		'arguments[0].scrollIntoView( { block: "center", inline: "center" } )',
+		'arguments[0].scrollIntoView( { block: "' + block + '", inline: "' + inline + '" } )',
 		selectorElement
 	);
 }
--- a/lib/driver-helper.js
+++ b/lib/driver-helper.js
@@ -408,11 +408,20 @@ export async function refreshIfJNError( driver, timeout = 2000 ) {
 	return await refreshIfNeeded();
 }
 
-export async function scrollIntoView( driver, selector, block = 'center', inline = 'center' ) {
+export async function scrollIntoView( driver, selector ) {
 	let selectorElement = await driver.findElement( selector );
 
 	return await driver.executeScript(
-		'arguments[0].scrollIntoView( { block: "' + block + '", inline: "' + inline + '" } )',
+		'arguments[0].scrollIntoView( { block: "center", inline: "center" } )',
+		selectorElement
+	);
+}
+
+export async function scrollToBottom( driver, selector ) {
+	let selectorElement = await driver.findElement( selector );
+
+	return await driver.executeScript(
+		'arguments[0].scrollIntoView( { block: "end", inline: "center" } )',
 		selectorElement
 	);
 }
--- a/lib/pages/people-page.js
+++ b/lib/pages/people-page.js
@@ -57,10 +57,32 @@ export default class PeoplePage extends AsyncBaseContainer {
 	}
 
 	async viewerDisplayed( username ) {
+		await DriverHelper.waitTillPresentAndDisplayed(
+			this.driver,
+			By.css( '.people-selector__infinite-list' )
+		);
+		await DriverHelper.scrollIntoView(
+			this.driver,
+			By.css( '.people-selector__infinite-list' ),
+			'end'
+		);
 		return DriverHelper.isEventuallyPresentAndDisplayed(
 			this.driver,
 			By.css( `.people-profile__login[data-e2e-login="${ username }"]` )
 		);
+		// for ( let i = 0; i < 10; i++ ) {
+		// 	let displayed = DriverHelper.isEventuallyPresentAndDisplayed(
+		// 		this.driver,
+		// 		By.css( `.people-profile__login[data-e2e-login="${ username }"]` )
+		// 	);
+		// 	if ( displayed ) {
+		// 		return displayed;
+		// 	}
+		// 	else {
+		// 		await DriverHelper.waitTillPresentAndDisplayed( this.driver, By.css( '.people-selector__infinite-list' ) );
+		// 		await DriverHelper.scrollIntoView( this.driver, By.css( '.people-selector__infinite-list' ), 'end' );
+		// 	}
+		// }
 	}
 
 	async removeUserByName( username ) {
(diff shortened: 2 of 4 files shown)


lib/driver-helper.js
```diff
@@ -416,3 +416,12 @@ export async function scrollIntoView( driver, selector ) {
 		selectorElement
 	);
 }
+
+export async function scrollToBottom( driver, selector ) {
```
Stojdza: `scrollIntoView()` didn't work always because Viewers list is infinite and users didn't load all at once. That's why I needed a way to scroll list(or, generally, any other element) step by step.

lib/pages/people-page.js
```diff
@@ -57,18 +57,41 @@ export default class PeoplePage extends AsyncBaseContainer {
...
-			By.css( `.people-profile__login[data-e2e-login="${ username }"]` )
+		await DriverHelper.isEventuallyPresentAndDisplayed( this.driver, By.css( '.count' ) );
+		let noOfViewers = parseInt(
+			await this.driver.findElement( By.css( '.section-header__label .count' ) ).getText()
 		);
+
+		let retries = Math.round( noOfViewers / 10 );
+		let displayed;
+
+		for ( let i = 0; i < retries; i++ ) {
```
Stojdza: List will be scrolled enough times to check every user. `retries` is rounded number of all users divided by 10.

specs/wp-invite-users-spec.js
```diff
@@ -206,6 +206,7 @@ describe( `[${ host }] Invites:  (${ screenSize })`, function() {
 		const newInviteEmailAddress = dataHelper.getEmailAddress( newUserName, inviteInboxId );
 		const siteName = config.get( 'privateSiteForInvites' );
 		const siteUrl = `https://${ siteName }/`;
+		let removedViewerFlag = false;
```
Stojdza: This flag is used in `after` block - if the user is successfully deleted in `Can see new user added and can be removed` step, there is no need to try deleting user again.

alisterscott:
Thanks for working on this @Stojdza 

This is such a tricky one - the site has 30 viewers currently and it's still not fully working all the time with this PR - I see messages like:

```
  2) [WPCOM] Invites:  (desktop)
       Inviting New User as a Viewer of a WordPress.com Private Site: @parallel
         "after all" hook:
     NoSuchElementError: no such element: Unable to locate element: {"method":"css selector","selector":".people-selector__infinite-list"}
  (Session info: chrome=69.0.3497.100)
  (Driver info: chromedriver=2.41.578706 (5f725d1b4f0a4acbf5259df887244095596231db),platform=Mac OS X 10.13.6 x86_64)
```

If we got rid of the assertions about the user being displayed and not, and just invited them, make sure they can view logged in, then remove them and make sure they can't view I think this would be enough.

We wouldn't need to assert the content of the list but we would need to make sure we can click the remove button - will this work when there are 30+ and 50+ viewers displayed? The button click not the assert?

Otherwise I can have a look at seeing what it would take to get calypso to show a search field to filter the results...

alisterscott:
Added a request to calypso for this: https://github.com/Automattic/wp-calypso/issues/27334

I can't work out how to add it easily - I'd need some help

Stojdza:
@alisterscott I've tried with more than 50 users and it's passing locally. I'm investigating why it doesn't work on CircleCI and what could be different. 

I'll try to remove assertions and go only with clicking Remove button.

Stojdza:
> If we got rid of the assertions about the user being displayed and not, and just invited them, make sure they can view logged in, **then remove them** and make sure they can't view I think this would be enough.

This won't work, because if there are 50 or 100 users, we won't be able to remove last user from the list.

Stojdza:
> Otherwise I can have a look at seeing what it would take to get calypso to show a search field to filter the results...

:bulb: Or, to show last invited user at the top of the list?

Stojdza:
@alisterscott I think we have it now. No need for workarounds :)

lib/pages/people-page.js
```diff
@@ -57,18 +57,38 @@ export default class PeoplePage extends AsyncBaseContainer {
...
-			By.css( `.people-profile__login[data-e2e-login="${ username }"]` )
+		await DriverHelper.isEventuallyPresentAndDisplayed( this.driver, By.css( '.count' ) );
+		let noOfViewers = parseInt(
+			await this.driver.findElement( By.css( '.section-header__label .count' ) ).getText()
 		);
+
+		let retries = Math.round( noOfViewers / 10 );
+		let displayed;
+
+		for ( let i = 0; i <= retries; i++ ) {
```
Stojdza: List will be scrolled enough times to check every user. `retries` is rounded number of all users divided by 10.

specs/wp-invite-users-spec.js
```diff
@@ -206,6 +206,7 @@ describe( `[${ host }] Invites:  (${ screenSize })`, function() {
 		const newInviteEmailAddress = dataHelper.getEmailAddress( newUserName, inviteInboxId );
 		const siteName = config.get( 'privateSiteForInvites' );
 		const siteUrl = `https://${ siteName }/`;
+		let removedViewerFlag = true;
```
Stojdza: This flag is used in `after` block - if the user is successfully deleted in `Can see new user added and can be removed` step, there is no need to try deleting user again.

lib/pages/people-page.js
```diff
@@ -57,18 +57,38 @@ export default class PeoplePage extends AsyncBaseContainer {
...
+			displayed = await DriverHelper.isEventuallyPresentAndDisplayed(
+				this.driver,
+				By.css( `.people-profile__login[data-e2e-login="${ username }"]` ),
+				500
+			);
+			if ( displayed ) {
+				break;
+			} else {
+				await DriverHelper.scrollToBottom( this.driver, By.css( '.layout__primary' ) );
```
Stojdza: This `.layout__primary` selector solved scrolling issue. At first, I used `.people-selector__infinite-list` and it worked only locally.

PR #1497, opened 2018-09-26
Update Woo Tests to use Steps instead of Before to Log In

alisterscott:
For some reason this is causing problems when running on https://github.com/Automattic/wp-e2e-tests-woocommerce


alisterscott:
Merging now to fix problems on Woo builds

PR #1503, opened 2018-10-01
Add new version of domainManagementSuggestion and remove old one.

delputnam:
Adding new version of the domainManagementSuggestion and removing the old version.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -70,10 +70,10 @@
 		"readerSearchPlaceholder",
 		"showPlanCreditsApplied",
 		"cartNudgeUpdateToPremium",
-		"domainManagementSuggestion",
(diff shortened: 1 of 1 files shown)


config/default.json
```diff
@@ -70,10 +70,10 @@
 		"readerSearchPlaceholder",
 		"showPlanCreditsApplied",
 		"cartNudgeUpdateToPremium",
-		"domainManagementSuggestion",
 		"userFirstSignup",
 		"signupAtomicStoreVsPressable",
-		"dotBlogSuggestions"
+		"dotBlogSuggestions",
+		"domainManagementSuggestionV2",
```
alisterscott: Actually, I think there's an additional comma now

Issue #30, opened 2016-05-18
Remove all xpath selectors


alisterscott:
1. Find each xpath selector
2. Add data-attribute to wp-calypso react code
3. Use css selector of data attribute added

hoverduck:
Xpath selectors are inefficient, fragile, and difficult to maintain.  We should scrub the list of the ones we're using and link them to open issues in wp-calypso to have data attributes added that we can use instead.

zhongruige:
I've started looking into this (removing xpaths is always a happy project), and I created a branch to work on, which you can see [here](https://github.com/Automattic/wp-e2e-tests/tree/remove-all-xpath-selectors).

Please let me know if you have any comments/suggestions. Thanks!

alisterscott:
@zhongruige do you mind creating a pull request for your changes on your branch? It makes it easier to review/comment

zhongruige:
@alisterscott Sure thing! I'll go ahead and do that now.

zhongruige:
Hi there! I just wanted to give a quick update on where I am at with this.

There's still two pages I am not sure on where they point to,  [wp-admin-snippets-page](https://github.com/Automattic/wp-e2e-tests/blob/master/lib/pages/wp-admin/wp-admin-snippets-page.js), and the [admin-jetpack-settings-page](https://github.com/Automattic/wp-e2e-tests/blob/master/lib/pages/wp-admin/wp-admin-jetpack-settings-page.js). The latter looks like it should be the settings page in wp-admin, but it doesn't seem to match up to the current UI I have in my account.

Overall: I am currently stuck on xpath selectors in the following files:

- customer-page.js: [one here](https://github.com/Automattic/wp-e2e-tests/blob/72688e9791687a5d0987f25f3600fc761d1672b5/lib/pages/customizer-page.js#L306) and [one here](https://github.com/Automattic/wp-e2e-tests/blob/72688e9791687a5d0987f25f3600fc761d1672b5/lib/pages/customizer-page.js#L390)
- themes-page.js: [one here](https://github.com/Automattic/wp-e2e-tests/blob/72688e9791687a5d0987f25f3600fc761d1672b5/lib/pages/themes-page.js#L80) and [one here](https://github.com/Automattic/wp-e2e-tests/blob/72688e9791687a5d0987f25f3600fc761d1672b5/lib/pages/themes-page.js#L94)
[...]

PR #1505, opened 2018-10-04
Remove xpath selectors

Stojdza:
This PR is related to https://github.com/Automattic/wp-e2e-tests/issues/30

Additional data-e2e are added on wp-calypso side - https://github.com/Automattic/wp-calypso/pull/27545 and https://github.com/Automattic/wp-calypso/pull/27624 (waiting for these two to be merged, after that tests should pass)

As a result, some of xpath selectors are removed, and we can use class selector with specific data-e2e attribute. Some of these attributes will be specific (for example, same as comment which is posted by the test) and easy to locate. And ofc, tests won't fail if something is changed on the page.


Stojdza:
Thanks @bsessions85

PR #1506, opened 2018-10-06
Fix domain mapping spec

alisterscott:
This was failing as a message was displayed on the mapping page that the domain was available instead of showing the cart:

screen shot 2018-10-06 at 2 34 40 pm

Updated spec to use a domain already registered (go.com)


alisterscott:
@Automattic/flowpatrol merging to get builds passing

PR #1519, opened 2018-10-12
Run Gutenpack tests against JN site which includes PR changes

brbrr:
To run Gutenberg tests against calypso PR - we need to pass calypso branch argument into JN site creation URL. This PR makes it possible. 

One thing that I not certain about - usage of `branchName` outside of `-b` context. So, a bit of review there would be very helpful.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -86,7 +86,7 @@ while getopts ":a:RpS:b:B:s:gjWCJH:wzyl:cm:fiIUvxu:h" opt; do
       ;;
     S)
       export LIVEBRANCHES="true"
-      NODE_CONFIG_ARGS+=("\"liveBranch\":\"true\",\"calypsoBaseURL\":\"https://hash-$OPTARG.calypso.live\"")
+      NODE_CONFIG_ARGS+=("\"liveBranch\":\"true\",\"calypsoBaseURL\":\"https://hash-$OPTARG.calypso.live\",\"branchName\":\"$BRANCHNAME\"")
       continue
       ;;
     b)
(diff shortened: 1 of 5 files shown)


run.sh
```diff
@@ -86,7 +86,7 @@ while getopts ":a:RpS:b:B:s:gjWCJH:wzyl:cm:fiIUvxu:h" opt; do
       ;;
     S)
       export LIVEBRANCHES="true"
-      NODE_CONFIG_ARGS+=("\"liveBranch\":\"true\",\"calypsoBaseURL\":\"https://hash-$OPTARG.calypso.live\"")
+      NODE_CONFIG_ARGS+=("\"liveBranch\":\"true\",\"calypsoBaseURL\":\"https://hash-$OPTARG.calypso.live\",\"branchName\":\"$BRANCHNAME\"")
```
bsessions85: Adding the branchName here causes `getCalypsoLiveQuery` to add a branch querystring parameter. That would normally be fine, but with dserve, if the url is one of the new hash subdomain urls, if there are `hash` or `branch` params, then the other parameters are stripped. There are some tests that this causes problems(domains tests I think).

brbrr: Right, but since we moved to branch subdomains - `getCalypsoLiveQuery` becomes not needed? I tried to find where it still used, but it looks like `-b` not used anymore and might be removed.

bsessions85: Yeah, we left that in there mostly as a fallback. I had originally ripped it out and after talking with @alisterscott  we decided to leave it. It also adds the hash for when a developer uses a hash with a base url of calypso.live.

brbrr: So it's better to leave it as is? (not remove it for now)

bsessions85: Lets just remove it since it adds some unnecessary complexity to the configuration.

PR #1521, opened 2018-10-12
Use an env var for the circleci image.

bsessions85:
This sets the CircleCI config to use the image based on environment variable. When .nvmrc file is updated, the script will update the environment variable for use on future builds.

Testing will occur with next node version change


alisterscott:
Thanks for this @bsessions85 

One question this raises is how we test a new node version?

I created this PR which updates to Node 10.12.0 but it will still run on the older version since the env variable won't update until I merge the PR? I think this is a 🐔 and 🥚situation as I don't know it'll be fine to merge until I've tested it

What do you recommend?

bsessions85:
That is a good question. I guess our option is to revert this one and always update in both places, or do a combination approach and update the node version in the image before we run the tests.

PR #1522, opened 2018-10-15
Update node to 10.12.0

alisterscott:
I don't think CircleCI is actually testing the node version change

> Starting container circleci/node:10.11.0-browsers


bsessions85:
Merging this one in

PR #1533, opened 2018-10-16
Update node version when updating nvmrc file

bsessions85:
This is just an idea I had when thinking about [this comment](https://github.com/Automattic/wp-e2e-tests/pull/1521#issuecomment-429706408). This will update the nvm version before installing/updating npm packages if the .nvmrc file has changed. I'm still not sure if it is better to try this or just make sure we're changing both the .nvmrc file and the circleCI image when the node version is updated. Thoughts @alisterscott, @Stojdza ?


alisterscott:
Thanks for this @bsessions85 - this is awesome

Tested it by updating node version and running in Circle and I can confirm it's using the right versions

:shipit:

Stojdza:
>  I'm still not sure if it is better to try this or just make sure we're changing both 

This looks perfect. I would prefer to change it just in one place :)

Issue #1426, opened 2018-08-22
Investigate adding API calls from e2e tests


tammullen:
Try replacing some test steps from an existing test with API calls to see if this could provide an improvement in execution time. The intention would be only to replace test steps that are repeated across tests and are only used to put the test into a certain state.

PR #1535, opened 2018-10-17
Test import in signup flow

creativecoder:
Adds a test for the following import signup flow to `wp-signup-spec`:

1. Visit `/start/import`
1. Enter various invalid urls and expect an error
1. Enter valid url and continue
1. Retry if site importability check fails the first time
1. Enter account details
1. Confirm account email
1. Land on importers page and see preview of site to be imported

## To test locally

1. Set `calypsoBaseURL` to `https://wpcalypso.wordpress.com`
1. Run `wp-signup-spec` `env NODE_ENV=my-config ./node_modules/.bin/mocha specs/wp-signup-spec.js`

To test retry when site importability check fails, sandbox public-api.wordpress.com and remove/modify `is-site-importable` endpoint.


creativecoder:
Questions to consider:

1. Should email activation step be abstracted now that it's tested in 2 places? Where would be a good place to put that?
1. Should account deletion step be abstracted... it's used in several places? Where would be a good place to put that?
1. What Wix site should be used to test import?

Stojdza:
Hey @creativecoder, thanks for adding this flow! 

> Should email activation step be abstracted now that it's tested in 2 places? Where would be a good place to put that?
Should account deletion step be abstracted... it's used in several places? Where would be a good place to put that?

I propose to leave it as it is for now. We have plan to optimize tests, including [adding API](https://github.com/Automattic/wp-e2e-tests/issues/1426) and speed them up. If you have any proposal feel free to share with us. 

> What site should be used to test import?

I'm not sure about this question. As I can see in the test, you only preview imported site. Does that mean import is not finished? And same site can be used for every test execution?

jblz:
Thanks for putting this together @creativecoder!  At a glance it looks like it provides some good, basic coverage for the flow.

Since it's still `In Progress`, is there anything I or anyone on the team can do to help get it ready for review?

specs/wp-signup-spec.js
```diff
@@ -1590,4 +1592,75 @@ describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function() {
...
+				emails.length,
+				1,
+				'The number of newly registered emails is not equal to 1 (activation)'
+			);
+			const activationLink = emails[ 0 ].html.links[ 0 ].href;
+			assert( activationLink !== undefined, 'Could not locate the activation link email link' );
+			await driver.get( activationLink );
+		} );
+
+		step( 'Can then see preview of site to be imported', async function() {
```
jblz: Something to keep in mind, I've seen the `is-site-importable` call fail (super rare I'd imagine, but it happens), so we need to make sure the assertion in this test will be true even if an API call fails (or we'll need to mock it).

creativecoder:
@Stojdza 

> I propose to leave it as it is for now. We have plan to optimize tests, including adding API and speed them up. If you have any proposal feel free to share with us.

Thanks for responding! I didn't have anything specific in mind, just noticed the code duplication.

> As I can see in the test, you only preview imported site.

For this initial test, I think we'll limit to testing until that point in this signup flow (preview of site to be imported). The import itself can take quite a while and has some additional complexities we'll need to think more about if/how to include in e2e tests.

> What site should be used to test import?

This was more a question for those working on importer features (including myself), regarding how we want to handle testing this import flow (will say more about that in a separate comment).

creativecoder:
@jblz 

> Since it's still In Progress, is there anything I or anyone on the team can do to help get it ready for review?

Thanks for asking! I think we have consensus that testing the signup flow until the importer preview step accomplishes the goals of this PR, so the only thing left is deciding how to handle the `is-site-importable` call.

> Something to keep in mind, I've seen the is-site-importable call fail (super rare I'd imagine, but it happens), so we need to make sure the assertion in this test will be true even if an API call fails (or we'll need to mock it).

One the one hand, a primary goal of e2e testing is to test the whole system (including API calls), on the other we want the tests to be reliable. In e2e tests, I usually do not mock external calls, since that would indicate something isn't working and I'd want the test to flag that.

@Stojdza Are there any guidelines in this repo for mocking vs not mocking external API calls?

creativecoder:
Also, currently the signup flow being tested in this PR is only enabled in wpcalypso.wordpress.com. What's the process for introducing this test? Do we need to wait until the flow is available on wordpress.com to merge the test to master?

jblz:
I managed to get this up and running (was my first time running e2e tests locally -- neato!).

I added a test that the site importer input pane is shown in 4454f39.
That's covering that the import section "drills in" to the appropriate engine. 

> Answer the previous question about mocking is-site-importable endpoint

What do you think about forgoing the preview step for now?

bsessions85:
> @bsessions85 do you know something about this? I think that we don't have any guidelines.
I am not aware of any guidelines either, but we will typically try to catch failures like this and retry or post to slack channel that it failed and skip the test (ie. domains tests)

creativecoder:
> let's at least do the catch, post to slack, & skip the test bit here.

Sounds like a plan... will work on adding that.

> Any idea why the CI is failing on this branch?

I was assuming it was because flow isn't enabled in production... but looks like there's more to it than that.

creativecoder:
I've added a retry for site importability check and it skips the test if the retry fails.

Since the test is always checking the same site, I believe that testing for the image in the last step should be stable: it's always checking the same mshots url to load the image.

> Try a rebase.

That helped! Still seeing some errors on mobile related to new tests in the PR 🤔

creativecoder:
Test video shows redirect to `main` signup flow on wordpress.com since this flow isn't enabled in production. So I think this is ready to go--should have passing tests once import signup flow is "released".

@bsessions85 @Stojdza would you mind reviewing the diff with most recent updates and make sure everything is in line with code for this test?

PR #1538, opened 2018-10-18
Domains: Fix /domains flow for logged out

aidvu:
We removed the step to choose how you want your domain
for logged users. We should add the same flow but
for logged in users.


Stojdza:
@aidvu I tried it locally, and it is failing on step before your change - `Can visit the domains start page`, with error `Timed out waiting for element with css selector of '.step-wrapper' to be present and displayed`. Is this related to your changes?

aidvu:
Yep, need to fix something in Calypso.

Issue #1530, opened 2018-10-16
Selecting wrong menu items


alisterscott:
If you run the tests without retrying it's common to see the tests selecting the wrong item in the menu sidebar and then failing as the incorrect screen is shown

Something is funky with how we click these sidebar items - we should investigate why

To reproduce change   `"max_test_attempts": 3,` to   `"max_test_attempts": 1,` in `magellan.json` and commit the change to see the failures in CircleCI for a branch

Stojdza:
This happened recently for 'People' page when I ran tests locally. It was failing randomly. 
@alisterscott What do you think to change the way how we are getting these elements? Instead of `'.sites-navigation [data-tip-target="people"] a'` to add data-e2e attribute to class `menu-link-text` for every item in sidebar?

<img width="867" alt="screen shot 2018-10-18 at 14 39 02" src="https://user-images.githubusercontent.com/7116222/47171823-1299c080-d30a-11e8-8b59-10b8828b978c.png">

alisterscott:
> We can pass .menu-link-text as a selector and item name from the sidebar to selectElementByText()

I think we should try to avoid selecting menu items by text since they're translated and won't work in other languages, eg for i18n tests, and also menu text is more likely to change than a specific attribute. Or do I misunderstand something?

alisterscott:
> We can pass .menu-link-text as a selector and item name from the sidebar to selectElementByText()

I think we should try to avoid selecting menu items by text since they're translated and won't work in other languages, eg for i18n tests, and also menu text is more likely to change than a specific attribute. Or do I misunderstand something?

Stojdza:
@alisterscott I replied to this comment [here](https://github.com/Automattic/wp-e2e-tests/pull/1540#issuecomment-431930428). So yes, data e2e attribute would be more stable solution.

PR #1540, opened 2018-10-21
Find element by text and click on it

Stojdza:
According to discussion from [here](https://github.com/Automattic/wp-calypso/pull/27624#discussion_r223697178) I added [method](https://github.com/Automattic/wp-e2e-tests/compare/try/select-by-text?expand=1#diff-0e3b6a0d13153e063b03d83d728e6521R429) in `driver-helper.js` so that we can find the element by text and click on it. 

It can be useful when we have many elements with the same selector with different content/text, for example, comments in notification sidebar. This way, we don't need to use `xpath` selectors for specific text. [Here](https://github.com/Automattic/wp-e2e-tests/compare/try/select-by-text?expand=1#diff-a8284112885a5346b63bcfc5d1b1789fR30) and [here](https://github.com/Automattic/wp-e2e-tests/compare/try/select-by-text?expand=1#diff-f33fd6bbca9e0118c28fca06ce7e4afaR82) are the examples. And we don't need to add data e2e attributes to wp-calypso. 

To test:
Makse sure that specs `wp-theme-switch-spec.js`, `wp-theme-multisite-spec.js`, `wp-reader-spec.js`, `wp-notification-spec.js` are passing.

### DIFF
--- a/lib/components/notifications-component.js
+++ b/lib/components/notifications-component.js
@@ -28,8 +28,8 @@ export default class NotificationsComponent extends AsyncBaseContainer {
 	}
 
 	async selectCommentByText( commentText ) {
-		const commentSelector = by.xpath( `//div[normalize-space(text())='${ commentText }']` );
-		return await driverHelper.clickWhenClickable( this.driver, commentSelector );
+		let commentSelector = by.css( '.wpnc__excerpt' );
+		return await driverHelper.selectElementByText( this.driver, commentSelector, commentText );
 	}
 
 	async trashComment() {
--- a/lib/driver-helper.js
+++ b/lib/driver-helper.js
@@ -425,3 +425,11 @@ export async function scrollToBottom( driver, selector ) {
 		selectorElement
 	);
 }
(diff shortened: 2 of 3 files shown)


Stojdza:
@alisterscott Maybe we can use this for the issue https://github.com/Automattic/wp-e2e-tests/issues/1530? We can pass `.menu-link-text` as a selector and item name from the sidebar to `selectElementByText()`

alisterscott:
> We can pass .menu-link-text as a selector and item name from the sidebar to selectElementByText()

I think we should try to avoid selecting menu items by text since they're translated and won't work in other languages, eg for i18n tests, and also menu text is more likely to change than a specific attribute. Or do I misunderstand something?

Stojdza:
@alisterscott You're right, I forgot about i18n. Let's use this only to change current xpaths in code :)

Stojdza:
@alisterscott Is this ok to merge? It will be used instead of xpaths.

PR #1502, opened 2018-09-27
Try selenium 4.0.0-alpha.1

Stojdza:
This is related to the issue described in #1366. I was seeing this error in several different places and it happens randomly. According to https://github.com/SeleniumHQ/selenium/issues/5345 it should be fixed with selenium `4.0.0-alpha.1`. 

On CircleCI two tests are failing, and it's related to `promise.all()`, I'm guessing it's not supported in this version of Selenium, but we are using it only in one place. So I would say it works fine on CircleCI. 

Locally, I tried with all tests several times, and I didn't see error `UnhandledPromiseRejectionWarning: Error: EPIPE write EPIPE`, which is good, but still I'm not 100% sure it is solved, because it was happening randomly in the past. 

This PR is just for testing purpose, to see will `4.0.0-alpha.1` solve warnings. 

Fixes #1366

PR #1542, opened 2018-10-22
i18n tests: Use setRect() instead setSize() to set browser size

Stojdza:
I18n tests are [failing](https://circleci.com/gh/Automattic/wp-e2e-tests-i18n) for several days. The problem was introduced with #1502, when we added selenium 4.0.0-alpha.1. According to [changes](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/CHANGES.md) in new Selenium `setSize()` is removed, and `setRect()` is added.


alisterscott:
Woohoo - screenshots are better now too!

PR #1546, opened 2018-10-23
Re-enable free .blog subdomain test

aidvu:


alisterscott:
@aidvu how was this fixed? Was it an API change?

aidvu:
Yep. `dot` wasn't taking into account the vertical. I updated the backend to just return the vertical subdomains if present.

PR #1547, opened 2018-10-23
Add pauseBetweenKeysMS to setWhenSettable()

Stojdza:
Adding 1ms pause between keys, to see what will happen. 

Occasionally, tests are failing because [wrong text](https://23521-57936731-gh.circle-artifacts.com/0/home/circleci/wp-e2e-tests/screenshots/FAILED-EN-MOBILE-can-submit-test-payment-details-1540145224091.png) is _typed_, some letters are missing. 

I ran tests several times on CircleCi against this PR, and didn't see this error. We could try, and see what will happen in the next few days.


alisterscott:
Merged but had to revert as IE11 broke

PR #1548, opened 2018-10-23
Use data-e2e attributes for Sidebar items

Stojdza:
Occasionally, we have failed e2e tests because wrong item is clicked in the sidebar. For example, _Sharing_ is clicked instead of _People_. Until now, we were getting an element, for example _People_, like this `.sites-navigation [data-tip-target="people"] a`. With this change we could get it like this `.menu-link-text[data-e2e-sidebar="People"]`, and hopefully, this way should be stable.

To test:
Make sure that every item in the sidebar has data e2e attribute `data-e2e-sidebar="<item_name>"`.
While waiting [wp-calypso PR](https://github.com/Automattic/wp-calypso/pull/28031) to me merged, it can be tested on wp-calypso branch `try/sidebar-e2e-selectors`.

Blocked by https://github.com/Automattic/wp-calypso/pull/28031

Fixes https://github.com/Automattic/wp-e2e-tests/issues/1530

### DIFF
--- a/lib/components/nav-bar-component.js
+++ b/lib/components/nav-bar-component.js
@@ -28,7 +28,7 @@ export default class NavBarComponent extends AsyncBaseContainer {
 		);
 	}
 	async clickMySites() {
-		const mySitesSelector = by.css( '.masterbar__item-content' );
+		const mySitesSelector = by.css( 'header.masterbar a.masterbar__item' );
 		return await driverHelper.clickWhenClickable(
 			this.driver,
 			mySitesSelector,
--- a/lib/components/nav-bar-component.js
+++ b/lib/components/nav-bar-component.js
@@ -28,7 +28,7 @@ export default class NavBarComponent extends AsyncBaseContainer {
 		);
 	}
 	async clickMySites() {
-		const mySitesSelector = by.css( 'header.masterbar a.masterbar__item' );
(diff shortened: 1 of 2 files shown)


lib/components/nav-bar-component.js
```diff
@@ -28,7 +28,7 @@ export default class NavBarComponent extends AsyncBaseContainer {
 		);
 	}
 	async clickMySites() {
-		const mySitesSelector = by.css( 'header.masterbar a.masterbar__item' );
+		const mySitesSelector = by.css( '.masterbar__item-content' );
```
JavonDavis: @Stojdza This causes tests to fail on mobile for me. For example, try running wp-plan-purchase-spec on mobile.

If this wasn't a part of the instability maybe we could leave it as is before or maybe

```suggestion
		const mySitesSelector = by.css( '.masterbar__item[data-tip-target="my-sites"]' );
```

JavonDavis: @Stojdza This causes tests to fail on mobile for me. For example, try running specs/wp-plan-purchase-spec on mobile.

If this wasn't a part of the instability maybe we could leave it as is before or maybe

```suggestion
		const mySitesSelector = by.css( '.masterbar__item[data-tip-target="my-sites"]' );
```

will work better in this case?

PR #1532, opened 2018-10-16
Fix few Jetpack-related tests

brbrr:
1. Comments tests are failing randomly on Jetpack hosts. 
2. after updating Selenium version - some of the tests started to fail due to updated actions API (https://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/lib/input_exports_Actions.html)

To test: make sure that CI tests are green

PR #1551, opened 2018-10-24
Try reverting selenium to 3.6.0

alisterscott:
Things have been less stable since upgrading to 4.0.0 alpha

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6798,7 +6798,7 @@
         },
         "es6-promise": {
           "version": "3.0.2",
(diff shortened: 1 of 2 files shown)


Stojdza:
Some changes, caused by Selenium 4 alpha, were made in e2e tests after upgrading - https://github.com/Automattic/wp-e2e-tests/pull/1542 and https://github.com/Automattic/wp-e2e-tests/pull/1532
If we go back to Selenium 3.6 we should consider reverting these PRs also. 

Maybe there are more PRs regarding this change, I found these two.

alisterscott:
I think it was the mobile emulation causing the issues not selenium 4 alpha so I'm going to discard this PR

PR #1552, opened 2018-10-24
Disable media specs on calypso.live

alisterscott:
This temporarily disables any spec that uploads media on calypso.live due to this bug: https://github.com/Automattic/dserve/issues/69


alisterscott:
Merging to get calypso.live builds working again

PR #1553, opened 2018-10-24
Override `improvedOnboarding` abtest as `main`

southp:
## Summary
Reflecting the newly-added abtest in https://github.com/Automattic/wp-calypso/pull/27983, this PR override `improvedOnboarding` abtest variation as `main`.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -82,6 +82,7 @@
 		[ "aboutSuggestionMatches_20180704", "control" ],
 		[ "nudgeAPalooza_20180806", "control" ],
 		[ "includeDotBlogSubdomainV2_20180813", "no" ],
+		[ "improvedOnboarding_20181023", "main" ],
 		[ "gSuiteDiscountV2_20180822", "control" ],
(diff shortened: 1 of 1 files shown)


alisterscott:
Thanks for your PR @southp 

I can add it now, but for future reference you also need to add the test name to `"knownABTestKeys": [` in the config

alisterscott:
@southp yes you could run a single test using `.only` and you'll see a warning in the console like `Found an AB Testing key in local storage that isn't known: 'improvedOnboarding'. This may cause inconsistent A/B test behaviour, please check this is okay and add it to 'knownABTestKeys' in default.config`

ramonjd:
Ah, my bad. I missed that bit in the [docs](https://github.com/Automattic/wp-calypso/tree/master/client/lib/abtest#updating-our-end-to-end-tests-to-avoid-inconsistencies-with-ab-tests).

alisterscott:
@ramonjd no worries :)

PR #1560, opened 2018-10-25
Adds a sign up for site scenario for logged in customer

### LINKED ISSUE #1557: Add a sign up test for an existing customer
This could be as simple as hitting /start signed in and make sure it loads

Or we could add a specific account for this - don't use existing account which relies on a single site in Calypso

alisterscott:
This extends the sign up for a free account to then add a new site to the account.

This makes sure we can sign up from a logged in state

Fixes #1557

**Testing Instructions**

Make sure "Sign up for an account only (no site) then add a site" is running and passing desktop and mobile

### DIFF
--- a/lib/components/no-sites-component.js
+++ b/lib/components/no-sites-component.js
@@ -3,9 +3,17 @@
 import { By } from 'selenium-webdriver';
 
 import AsyncBaseContainer from '../async-base-container';
+import * as driverHelper from '../driver-helper.js';
 
 export default class NoSitesComponent extends AsyncBaseContainer {
 	constructor( driver ) {
 		super( driver, By.css( '.empty-content' ) );
+	}
+
+	async createSite() {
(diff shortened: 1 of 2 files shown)


alisterscott:
@ramonjd this _should_ catch any future sign up regressions from a logged in state

nb:
Does this test fail if https://github.com/Automattic/wp-calypso/pull/27983 is not reverted?

alisterscott:
> Does this test fail if Automattic/wp-calypso#27983 is not reverted?

@nb good question - I ran this against a PR I spun up with the changes that were reverted and yes it fails - see https://circleci.com/gh/Automattic/wp-e2e-tests-for-branches/13249

nb:
Great, thanks, @alisterscott, now we’re sure the exact same bug won’t strike again :)

PR #1561, opened 2018-10-25
Set 1ms timeout between keys

Stojdza:
Trying one more time with the timeout between keys. 

I set 1ms timeout for all browsers.

### DIFF
--- a/lib/driver-helper.js
+++ b/lib/driver-helper.js
@@ -207,7 +207,7 @@ export async function setWhenSettable(
 	driver,
 	selector,
 	value,
-	{ secureValue = false, pauseBetweenKeysMS = 0 } = {}
+	{ secureValue = false, pauseBetweenKeysMS = 1 } = {}
 ) {
 	const self = this;
 	const logValue = secureValue === true ? '*********' : value;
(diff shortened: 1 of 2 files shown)


lib/driver-helper.js
```diff
@@ -207,7 +207,7 @@ export async function setWhenSettable(
 	driver,
 	selector,
 	value,
-	{ secureValue = false, pauseBetweenKeysMS = 0 } = {}
+	{ secureValue = false, pauseBetweenKeysMS = 1 } = {}
```
alisterscott: I think we need to keep this at 0 _only_ for IE

alisterscott:
See https://github.com/Automattic/wp-calypso/pull/28083 with matching branch name still fails for IE

lib/driver-helper.js
Stojdza: @alisterscott Yeah, it looks like that. But [this](https://github.com/Automattic/wp-e2e-tests/blob/master/lib/components/secure-payment-component.js#L35) line confused me, it sets to 1ms if it's IE?

Stojdza:
Thanks @alisterscott! Looks good for now

PR #1573, opened 2018-10-30
Add a Reader signup e2e scenario

alisterscott:
Adds a new e2e scenario for signing up for a reader account


alisterscott:
> Should we remove newly created account?

Yes, good idea - I'll add it

PR #1575, opened 2018-10-30
Fix Coupon tests failing on Jetpack Hosts

JavonDavis:
Coupon tests were failing on Jetpack hosts because of the way the test was selecting the Plans tab. A simple check for WPCOM vs Jetpack host resolved this. 

To Test:
`JETPACKHOST=PRESSABLE mocha specs/wp-plan-purchase-spec.js -g 'Viewing a specific plan with coupon'`

### DIFF
--- a/specs/wp-plan-purchase-spec.js
+++ b/specs/wp-plan-purchase-spec.js
@@ -68,7 +68,7 @@ describe( `[${ host }] Plans: (${ screenSize }) @parallel @jetpack`, function() 
 		}
 	} );
 
-	describe.only( 'Viewing a specific plan with coupon:', function() {
+	describe( 'Viewing a specific plan with coupon:', function() {
 		let originalCartAmount, loginFlow;
 
 		before( async function() {
--- a/specs/wp-plan-purchase-spec.js
+++ b/specs/wp-plan-purchase-spec.js
@@ -88,8 +88,15 @@ describe( `[${ host }] Plans: (${ screenSize }) @parallel @jetpack`, function() 
 		} );
 
 		step( 'Can Select Plans tab', async function() {
-			let route = `plans/${ loginFlow.account.loginURL }`;
-			return await driver.get( dataHelper.getCalypsoURL( route ) );
+			const plansPage = await PlansPage.Expect( driver );
+			if ( host === 'WPCOM' ) {
+				await plansPage.openPlansTab();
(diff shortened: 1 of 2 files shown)


JavonDavis:
@Stojdza thanks for finding that for me, it's related to the changes I made here in the original PR https://github.com/Automattic/wp-e2e-tests/pull/1410, this should be all good now 🤞

specs/wp-plan-purchase-spec.js
```diff
@@ -68,7 +68,7 @@ describe( `[${ host }] Plans: (${ screenSize }) @parallel @jetpack`, function()
 		}
 	} );
 
-	describe( 'Viewing a specific plan with coupon:', function() {
+	describe.only( 'Viewing a specific plan with coupon:', function() {
```
Stojdza: Was this `.only` intentional for testing purposes? Just in case I'll leave a comment so we don't forget it :)

Stojdza:
@JavonDavis I'm seeing same error, but this time randomly 🤔I couldn't catch the reason. It happens on both - mobile and desktop, but not every time. Maybe something is slower/faster for me locally? Test is running smoothly on CircleCI.

specs/wp-plan-purchase-spec.js
JavonDavis: 🤦‍♂️ yeah that was my testing purposes, pushed a commit removing it.

JavonDavis:
> @JavonDavis I'm seeing same error, but this time randomly 🤔I couldn't catch the reason. It happens on both - mobile and desktop, but not every time. Maybe something is slower/faster for me locally? Test is running smoothly on CircleCI.

If it's the same error I'll assume it's because it doesn't wait long enough for that little loading animation when plans are appearing to finish before clicking the button. The problem is even during that period the button is still clickable but at that point successfully clicking it has no effect. This is why I have [this line](https://github.com/Automattic/wp-e2e-tests/pull/1575/files#diff-82ee16b545727d7e3b21a05a5c16a960R54). @Stojdza did you try increasing the sleep time a little and see if it would still randomly happen? If the 1 second has it running good in the CI, I don't think I'd increase it

Stojdza:
Thanks for the explanation @JavonDavis It works fine locally with 3s sleep 👌

brbrr:
I'm going to merge this, since Markdown failure is not related to this change

PR #1581, opened 2018-11-01
Add jetpack Markdown block test for Calypso env

brbrr:
Re-uses existing Markdown block test to run in Calypso environment. 
For now, it works only with Jetpack sites (via JN), but it can be easily ported to WPCOM sites. 

To test:
run `JETPACKHOST=PRESSABLE mocha specs-blocks/`

### DIFF
--- a/specs-blocks/gutenberg-markdown-block-spec.js
+++ b/specs-blocks/gutenberg-markdown-block-spec.js
@@ -117,6 +117,7 @@ function markdownTestSteps() {
 
 	step( 'Can publish the post and see its content', async function() {
 		const gEditorComponent = await GutenbergEditorComponent.Expect( driver );
+		await gEditorComponent.ensureSaved();
 		await gEditorComponent.publish( { visit: true } );
 		const postFrontend = await PostAreaComponent.Expect( driver );
 		const html = await postFrontend.getPostHTML();
--- a/specs-blocks/gutenberg-markdown-block-spec.js
+++ b/specs-blocks/gutenberg-markdown-block-spec.js
@@ -39,7 +39,7 @@ if ( screenSize !== 'mobile' ) {
 		driver = await driverManager.startBrowser();
 	} );
 
-	describe( `[${ host }] Gutenberg Markdown block: (${ screenSize }) @jetpack`, function() {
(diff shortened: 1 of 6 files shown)


specs-blocks/gutenberg-markdown-block-spec.js
```diff
@@ -31,37 +42,33 @@ if ( screenSize !== 'mobile' ) {
...
-</ul>
-`;
-			// Easy way to run/develop tests against local WP instance
-			// step( 'Can login to WPORG site', async function() {
-			// 	const loginPage = await WPAdminLogonPage.Visit( driver, 'http://wpdev.localhost/' );
-			// 	await loginPage.login( 'wordpress', 'wordpress' );
-			// } );
-
-			step( 'Can create wporg site and connect Jetpack', async function() {
+		before( 'Prepare Site for testing', async function() {
```
brbrr: At some point, this `before` block can be transferred into some sort of Gutenberg-prep flow, which will prepare sites for all possible cases: WPCOM-Calypso, WPCOM-wp-admin, Jetpack-Calypso, Jetpack-wp-admin.
This new flow also can include `Can start new post` step.

specs-blocks/gutenberg-markdown-block-spec.js
```diff
@@ -70,33 +77,49 @@ if ( screenSize !== 'mobile' ) {
...
-				const postFrontend = await PostAreaComponent.Expect( driver );
-				const html = await postFrontend.getPostHTML();
-				assert( html.includes( expectedHTML ) );
-			} );
+			markdownTest();
 		} );
 	} );
 }
+
+function markdownTest() {
```
brbrr: Not sure if that clear enough, and is this is the best way of code re-use. But tl; dr: it inserts all these steps everywhere the function was called.

simison: Would renaming it make it more clear? E.g. `markdownTestSteps` or `runMarkdownTest`.

alisterscott:
@brbrr does this one still need a review?

JavonDavis:
Doesn't seem like Gutenberg in calypso will be available for Jetpack sites. Closing until that changes.

PR #1585, opened 2018-11-01
Add 'Delete account' flow

### LINKED ISSUE #1576: Delete account flow
We have [account deletion](https://github.com/Automattic/wp-e2e-tests/blob/master/specs/wp-signup-spec.js#L183) in several places. It would be good to put this in one place and call it from there. Maybe something like `delete-account-flow.js`?

Stojdza:
Adding new flow for account deletion and use it in _Sign up_ tests. 

Resolves #1576

### DIFF
--- a/lib/flows/delete-account-flow.js
+++ b/lib/flows/delete-account-flow.js
@@ -1,0 +1,33 @@
+/** @format */
+
+import CloseAccountPage from '../pages/account/close-account-page';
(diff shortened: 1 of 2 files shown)


Stojdza:
@creativecoder We talked about this [here](https://github.com/Automattic/wp-e2e-tests/pull/1535#issuecomment-430812553). Does this look like a useful solution from your point of view?

creativecoder:
@Stojdza This is exactly what I was thinking of in that comment--thanks for implementing!

PR #1587, opened 2018-11-02
Update cart total selector

alisterscott:
PR: https://github.com/Automattic/wp-calypso/pull/28234

This is designed to be backwards compatible to work with existing PRs


alisterscott:
Failure unrelated to change

PR #1590, opened 2018-11-02
Update the signup flow test cases for Business users

taggon:
Since we're going to enable the onboarding checklist for the Business(https://github.com/Automattic/wp-calypso/pull/28117), the corresponding test cases should be updated as well.


alisterscott:
This is passing against the wp-calypso PR here: https://github.com/Automattic/wp-calypso/pull/28117

PR #1593, opened 2018-11-06
Fix window width on mobile test runs

bsessions85:
This PR changes the way we size the window for tests. The first thing it does is it uses the `--window-size` argument on Chrome browser. This opens the window at the size we want and doesn't rely on the `resize` functionality which can give unpredictable widths. The second part removes all the toolbars. The toolbars prevent the window from resizing to where we want it in CircleCI (490px instead of 400px). The `--kiosk` argument along with the `--app` argument  handle removing the toolbars while still allowing new windows to be opened, which is needed in some tests. I have it set to only remove the toolbars in CircleCI so that the buttons and urls can still be accessed in local troubleshooting.

To Test:
Run the tests and make sure the width of the browser is the expected width and is consistent across runs.

### DIFF
--- a/lib/driver-manager.js
+++ b/lib/driver-manager.js
@@ -131,6 +131,7 @@ export function startBrowser( { useCustomUA = true, resizeBrowserWindow = true }
 				} );
 				options.setProxy( this.getProxyType() );
 				options.addArguments( '--no-first-run' );
+				options.addArguments( '--kiosk' );
 				options.addArguments( getChromeWindowSize( screenSize ) );
 
 				if ( useCustomUA ) {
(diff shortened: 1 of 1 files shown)


lib/driver-manager.js
```diff
@@ -244,6 +250,39 @@ export async function resizeBrowser( driver, screenSize ) {
...
+				windowSize = '--window-size=400,1000';
+				break;
+			case 'tablet':
+				windowSize = '--window-size=1024,1000';
+				break;
+			case 'desktop':
+				windowSize = '--window-size=1440,1000';
+				break;
+			case 'laptop':
+				windowSize = '--window-size=1400,790';
```
JavonDavis: This is no big deal but what devices are we capturing with this one? Dimensions seems a bit uncommon just when eyeballing. Was this aiming at smaller notebooks? 13", 12"? Maybe 1280 by 800?  

screen shot 2018-11-06 at 1 58 42 pm

bsessions85: I'm not sure. I just used the sizes that were already being used. Perhaps @alisterscott can answer that

alisterscott: I think it was for visual diffs but we shouldn't worry any more since we don't have them

PR #1601, opened 2018-11-07
Domains: Skip /domains tests for variation

aidvu:
It's a back-end test, so no way atm to set the variation.

### DIFF
--- a/specs/wp-signup-spec.js
+++ b/specs/wp-signup-spec.js
@@ -851,7 +851,7 @@ describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function() {
 				} )
 			);
 
-			// Skip tests for a test we're currently running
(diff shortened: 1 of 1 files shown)


specs/wp-signup-spec.js
```diff
@@ -850,6 +850,11 @@ describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function() {
 					query: `new=${ expectedDomainName }`,
 				} )
 			);
+
+			// Skip for this A/B test variation (back-end test)
+			if ( ! driverHelper.isElementPresent( driver, '#domain-search' ) ) {
```
bsessions85: This didn't work in my testing. Try something like 
```
			if ( ! await driverHelper.isElementPresent( driver, By.css( '#domain-search' ) ) ) {
				return await this.skip();
			}
```

PR #1612, opened 2018-11-09
Add the `calypsoifyGutenberg` key to known tests.

kwight:
See: https://github.com/Automattic/wp-calypso/pull/28374


bsessions85:
Before this is merged, an override should be added so that we always get the default version.

Copons:
Sorry about this @bsessions85, I've added the override as requested.

bsessions85:
Thanks for adding @Copons! Looks good

Issue #1620, opened 2018-11-14
Planning: converting all the simple site editor tests to use Gutenberg Blocks


alisterscott:
Customers with simple sites can currently opt-in their site to use Gutenberg which is a re-skinned wp-admin editor directly linked from Calypso

This is an issue to plan and track progress of this work as it's quite large to ensure all our editor tests work on Gutenberg for simple sites.

I will break this down into steps which we can split across @Automattic/flowpatrol 

- [ ] Create a new simple test site that we can use for testing gutenberg editor separately to the existing test site which will continue to use Calypso
- [ ] Create a copy of the post editor spec that only runs on simple sites and uses new site specified above
- [ ] Create a copy of the page editor spec that only runs on simple sites and uses new site specified above
- [ ] Update "Public Posts: Preview and Publish a Public Post"
- [ ] Update "Basic Public Post"
- [ ] Update "Check Activity Log for Public Post"
- [ ] Update "Schedule Basic Public Post"
- [ ] Update "Private Posts"
- [ ] Update "Password Protected Posts"
- [ ] Update "Trash Post"
- [ ] Update "Edit a Post"
- [ ] Update "Insert a contact form"
- [ ] Update "Insert a payment button"
- [ ] Update "Revert a post to draft"
- [ ] Update "Public Pages"
- [ ] Update "Private Pages"
- [ ] Update "Password Protected Pages"
- [ ] Update "Insert a payment button into a page"

cc @gwwar

alisterscott:
Post editor in Gutenberg has been marked as a canary: #1623

bsessions85:
Payment button block doesn't seem to be out there at this point

alisterscott:
> Payment button block doesn't seem to be out there at this point

I think Payment Button and Contact Forms are still WIP with Jetpack

alisterscott:
Private post mostly done in #1651

alisterscott:
See https://github.com/Automattic/wp-e2e-tests/pull/1656 which converts most of the posts specs to run against Gutenberg in Calypso

alisterscott:
@JavonDavis / @Stojdza I've listed three more items for Gutenberg in Calypso including the Simple Payments and Contact forms - feel free to work on any of these

simison:
Noting that although Tiled gallery and Related posts block might be visible in Calypso-Gutenberg (behind the feature flag), they're not considered quite ready yet.

These should be "ready" (via p1HpG7-5V3-p2):

[...]

PR #1621, opened 2018-11-14
Adds Gutenberg Editor Specs

alisterscott:
This is a part of #1620 

This adds a new test site `https://e2eflowtestinggutenbergsimple.wordpress.com` which is opted in for Gutenberg, and this copies the editor specs to use this new site. 

**We need to keep the existing editor tests for now until 100% of users are using Gutenberg and the Calypso editor is removed completely.**

**Testing Instructions**

1. You'll need to decrypt the config for the new account
2. Make sure the new editor spec `Basic Public Post` works for the new Gutenberg simple site
3. Make sure none of the non-converted specs run

@Automattic/flowpatrol feel free to merge this if if makes sense to continue work on #1620 and since config is binary we should avoid merge conflicts


JavonDavis:
Merging since failure in CI is unrelated.

PR #1623, opened 2018-11-15
Add Gutenberg post canary

alisterscott:
Part of #1623


alisterscott:
Tested against wp-calypso PR and works well

PR #1625, opened 2018-11-15
Gutenbergify: Check Activity Log for Public Post

Stojdza:
Publish post and check activity log. 

Part of https://github.com/Automattic/wp-e2e-tests/issues/1620


Stojdza:
Thanks @JavonDavis! 

> The last step though it seemed the activity page reloaded about 3 or 4 times before passing

 Yeah, that's the usual flow, because sometimes it won't show in activity feed immediately.

PR #1627, opened 2018-11-15
[WIP] Gutenbergify: Public Posts: Preview and Publish a Public Post

JavonDavis:
There's a lot of steps in this test and I wanted to get some eyes on this as I'm about midway. There are a few things I did so far and a few that may cause conflicts.  @Stojdza @alisterscott @bsessions85  

* Renamed `GutenbergEditorHeaderComponent` to `GutenbergEditorComponent - did this because we were doing things with the editor in general and not just the header and thought the name was a little misleading
* Created components for Sidebar and Preview 
* Removed the Sharing step in the original step as that's no longer in the sidebar for the Gutenberg editor(actually not sure how a user would replicate this functionality from the original editor)
* Category and tags now have there own section in the sidebar so separated the method for expanding and collapsing. 
* Categories and tags no longer appear on the sidebar as they did after their saved so couldn't verify it in the same way we did originally
* No longer seeing tags when previewing(not sure if I'm just missing it)
* Chrome level notification about leaving page stops test from progressing to next test even though it's saved before launching the preview. 

Part of #1620

### DIFF
--- a/lib/driver-helper.js
+++ b/lib/driver-helper.js
@@ -454,3 +454,15 @@ export async function clearTextArea( driver, selector ) {
 		i--;
 	}
 }
+
+export async function dismissAlertIfPresent( driver ) {
+	try {
+		driver
+			.switchTo()
+			.alert()
+			.dismiss();
+		return true;
+	} catch ( error ) {
+		return false;
+	}
+}
--- a/lib/gutenberg/gutenberg-editor-component.js
+++ b/lib/gutenberg/gutenberg-editor-component.js
@@ -81,18 +81,10 @@ export default class GutenbergEditorComponent extends AsyncBaseContainer {
 		altTextInput.sendKeys( name );
 	}
 
-	async selectImageBlock() {
-		const appenderSelector = By.css( '.editor-inserter__toggle' );
-		const imageBlockSelector = By.css( '.editor-block-list-item-image' );
-		await driverHelper.clickWhenClickable( this.driver, appenderSelector );
-		return await driverHelper.clickWhenClickable( this.driver, imageBlockSelector );
-	}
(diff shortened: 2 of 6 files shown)


JavonDavis:
> Chrome level notification about leaving page stops test from progressing to next test even though it's saved before launching the preview. 

🤦‍♂️ thinking this is because it's still a WIP, once I get to the publish step I don't think this should be a problem.

alisterscott:
I've merged my latest updates to sidebar component and I won't leave any work pending so you can finish this up and get it merged Firday

alisterscott:
> Chrome level notification about leaving page stops test from progressing to next test even though it's saved before launching the preview.

We should have hooks that dismiss these windows

Otherwise we should do something like override the window onleave function in the browser to make it not show

alisterscott:
> Removed the Sharing step in the original step as that's no longer in the sidebar for the Gutenberg editor(actually not sure how a user would replicate this functionality from the original editor)

https://github.com/Automattic/wp-calypso/issues/28528

alisterscott:
> No longer seeing tags when previewing(not sure if I'm just missing it)

Worth following up and raising a bug in appropriate repository if not

JavonDavis:
> Worth following up and raising a bug in appropriate repository if not

@alisterscott now noticing this is because in the test after saving the draft the tag abruptly disappears and isn't saved. Even if I pause the test and manually click the save it disappears. Interesting enough if I manually walk through the entire flow it stays. I'll keep investigating but if not I'll open an issue in the calypso repo.

JavonDavis:
Resolved in https://github.com/Automattic/wp-e2e-tests/pull/1634

PR #1628, opened 2018-11-15
Add crowdsignalNameBasedSignup AB tests to knownABTestKeys

ice9js:
This PR adds `crowdsignalNameBasedSignup` to `knownABTestKeys`.  
The test will be introduced in [Automattic/wp-calypso#28546](https://github.com/Automattic/wp-calypso/pull/28546) and will only affect `crowdsignal` signup flow.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -67,7 +67,8 @@
 		"signupAtomicStoreVsPressable",
 		"krackenRebootM33",
 		"improvedOnboarding",
-		"privateByDefault"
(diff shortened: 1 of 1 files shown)


alisterscott:
Thanks for this change

aidvu:
You also need to add the test to the `overrideABTests` array. Updated that as part of a cleanup here: https://github.com/Automattic/wp-e2e-tests/pull/1658

PR #1629, opened 2018-11-15
Gutenberg Integration Post Editor Spec [WIP]

Copons:
Currently there are 2 implementations of Gutenberg in Calypso:
- Calypsoify Gutenberg (see p58i-7lU-p2) that redirects to wp-admin.
- The actual Gutenberg integration (accessible via direct URL `/gutenberg` in development, wpcalypso, and horizon).

The current e2e test involving Gutenberg are set up in a way that they actually end up testing Calypsoify Gutenberg, which basically means testing WPCOM Gutenberg, which is directly derived from Core Gutenberg, which has its own e2e suite (using Jest and Puppeteer).

Don't get me wrong: it's important that we keep doing e2e on Calypsoify Gutenberg for as long as it exist.
Though, we should also start testing the actual Gutenberg integration, which is currently being developed on top of WPCOM Gutenberg, which in turn is being built on top of Core Gutenberg. As you can imagine, the chances of breaking at every update are quite big.

Right now it's not available on `wordpress.com`, so my testing instructions involve 2 quirks:
1) Running the tests on local Calypso.
2) Using a config flag to force opening a `/gutenberg/post` URL instead of navigating to `/post` (that would redirect to wp-admin), and removing some instructions only relevant for Core Gutenberg.

The first quirk can be solved by enabling Gutenberg to production behind a 0% A/B test (props to @alisterscott for the suggestion).
I'm not sure how to solve the second quirk, but I'm confident folks that are more familiar with e2e can help me here.

So, this PR involves running the existing (in development) `wp-gutenberg-post-editor-spec.js`.
This also required a couple of simple selector changes in `GutenbergEditorHeaderComponent` to make sure the tests can target both the actual Gutenberg integration and Calypsoify Gutenberg.

### Testing instructions:

Run Calypso locally.

Run `env NODE_ENV=copons ./node_modules/.bin/mocha specs/wp-gutenberg-post-editor-spec.js`

With this config: `local-copons.json`

```json
{
	"calypsoBaseURL": "http://calypso.localhost:3000",
	"slackHook": "",
	"forceGutenbergCalypsoIntegration": true,
	"testAccounts": {
		"gutenbergSimpleSiteUser": [
			"username",
			"password"
		]
	}
}
```

`slackHook`: see https://github.com/Automattic/wp-e2e-tests/blob/master/docs/config.md
`gutenbergSimpleSiteUser`: credentials for a single-site user, already opted-in to Gutenberg

Of course rename `copons` in the command and the config file name as you see fit.


alisterscott:
I've tried a similar approach here: https://github.com/Automattic/wp-e2e-tests/pull/1652

Copons:
Lovely @alisterscott, thanks! ❤️

PR #1630, opened 2018-11-16
Disable import sign up test

alisterscott:
Failing due to a recent change in D20847-code


alisterscott:
@creativecoder a heads-up this is disabled for now

PR #1631, opened 2018-11-16
Import Signup Flow: Fix & re-enable e2e test

jblz:
As [reported](https://github.com/Automattic/wp-calypso/pull/28556#issuecomment-439253201) these tests stopped working around the time we enabled "Private By Default" for our signup flow.

### Changes

* Remove the `before` call that warned and skipped (#1630)
* New function `importPage.siteImporterCanStartImport`
* Refactor the site preview step to use `driverHelper.isEventuallyPresentAndDisplayed`
* Assert that the Email Verification Gate is not shown
* Leverage assertions instead of yielding values for more granular error messages

### To Test

* Configure `YOUR_TEST_ENV`s config file to point against prod
* Check out this branch
* (Optional) limit to just this spec like so:
```
--- a/specs/wp-signup-spec.js
+++ b/specs/wp-signup-spec.js
-       describe( 'Import a site while signing up @parallel', function() {
+       describe.only( 'Import a site while signing up @parallel', function() {
```
* Open the repo's directory in your terminal & run
  * `HEADLESS=1 env NODE_ENV=$YOUR_TEST_ENV ./node_modules/.bin/mocha specs/wp-signup-spec.js`
* Tests should pass, of course

### DIFF
--- a/lib/pages/import-page.js
+++ b/lib/pages/import-page.js
@@ -1,5 +1,6 @@
 /** @format */
 
+import assert from 'assert';
 import { By } from 'selenium-webdriver';
 import * as driverHelper from '../driver-helper.js';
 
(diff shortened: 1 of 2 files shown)


jblz:
Thanks much for taking  alook, @Stojdza. Yes, this behavior got changed by https://github.com/Automattic/wp-calypso/pull/28620 -- I'll pull out the email gate check for now & PR it separately so we can get this re-enabled.

PR #1632, opened 2018-11-16
Gutenberg: Trash post and Edit post tests

bsessions85:
Updated these tests to work on Gutenberg editor
part of #1620 
To Test:
Run both tests in desktop and mobile.

### DIFF
--- a/lib/gutenberg/gutenberg-editor-sidebar-component.js
+++ b/lib/gutenberg/gutenberg-editor-sidebar-component.js
@@ -62,4 +62,11 @@ export default class GutenbergEditorSidebarComponent extends AsyncBaseContainer 
 			}
 		);
 	}
+
+	async trashPost() {
+		const trashSelector = By.css( 'button.editor-post-trash' );
+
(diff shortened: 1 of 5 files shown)


lib/gutenberg/gutenberg-editor-sidebar-component.js
```diff
@@ -62,4 +62,11 @@ export default class GutenbergEditorSidebarComponent extends AsyncBaseContainer
 			}
 		);
 	}
+
+	async trashPost() {
+		const trashSelector = By.css( 'button.editor-post-trash' );
+
+		await driverHelper.waitTillPresentAndDisplayed( this.driver, trashSelector );
```
Stojdza: Have you tried without this line? I think it will work fine without it, that `clickWhenClickable` is enough.

bsessions85: Yeah, it is really there just to give a little more time for the autosave to kick in so that the trash button shows up. It was there in the original so I kept it.

PR #1634, opened 2018-11-17
[Gutenberg] Public Posts: Preview and Publish a Public Post

JavonDavis:
There's a lot of steps in this test and I wanted to get some eyes on this as I'm about midway. There are a few things I did so far and a few that may cause conflicts.  @Stojdza @alisterscott @bsessions85  

* Renamed `GutenbergEditorHeaderComponent` to `GutenbergEditorComponent - did this because we were doing things with the editor in general and not just the header and thought the name was a little misleading
* Created components for Sidebar and Preview 
* Removed the Sharing step in the original step as that's no longer in the sidebar for the Gutenberg editor(actually not sure how a user would replicate this functionality from the original editor)
* Category and tags now have there own section in the sidebar so separated the method for expanding and collapsing. 
* Categories and tags no longer appear on the sidebar as they did after their saved so couldn't verify it in the same way we did originally
* No longer seeing tags when previewing(not sure if I'm just missing it)
* Chrome level notification about leaving page stops test from progressing to next test even though it's saved before launching the preview. 

Part of #1620 

Original PR and conversation: https://github.com/Automattic/wp-e2e-tests/pull/1627

### DIFF
--- a/lib/gutenberg/gutenberg-editor-component.js
+++ b/lib/gutenberg/gutenberg-editor-component.js
@@ -1,7 +1,10 @@
 /** @format */
 import { By } from 'selenium-webdriver';
 import * as driverHelper from '../driver-helper';
+import * as driverManager from '../driver-manager.js';
 import AsyncBaseContainer from '../async-base-container';
+import GutenbergPreviewComponent from './gutenberg-preview-component';
+import webdriver from 'selenium-webdriver';
 
 export default class GutenbergEditorComponent extends AsyncBaseContainer {
 	constructor( driver ) {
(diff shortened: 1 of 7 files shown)


lib/gutenberg/gutenberg-editor-component.js
```diff
@@ -117,4 +120,95 @@ export default class GutenbergEditorHeaderComponent extends AsyncBaseContainer {
...
+	async enterImage( fileDetails ) {
+		const newImageName = fileDetails.imageName;
+		const newFile = fileDetails.file;
+
+		await this.sendFile( newImageName, newFile );
+	}
+
+	async toggleSidebar( open = true ) {
+		const sidebarSelector = '.edit-post-sidebar-header';
+		const sidebarOpen = await driverHelper.isEventuallyPresentAndDisplayed(
```
alisterscott: We should use `const sidebarOpen = await driverHelper.isElementPresent(` instead as this implementation waits 30 seconds to determine if it's open

lib/gutenberg/gutenberg-editor-component.js
```diff
@@ -117,4 +120,95 @@ export default class GutenbergEditorHeaderComponent extends AsyncBaseContainer {
...
+	async ensureSaved() {
+		await driverHelper.clickWhenClickable( this.driver, By.css( '.editor-post-save-draft' ) );
+		const savedSelector = By.css( 'span.is-saved' );
+
+		return await driverHelper.waitTillPresentAndDisplayed( this.driver, savedSelector );
+	}
+
+	static async switchToEditor( driver ) {
+		const tabs = await driver.getAllWindowHandles();
+		await driver.switchTo().window( tabs[ 0 ] );
```
alisterscott: This will leave the other window open, we should close it

lib/gutenberg/gutenberg-preview-component.js
```diff
@@ -0,0 +1,39 @@
...
+import { By } from 'selenium-webdriver';
+import AsyncBaseContainer from '../async-base-container';
+import ViewPostPage from '../pages/view-post-page';
+
+export default class GutenbergPreviewComponent extends AsyncBaseContainer {
+	constructor( driver ) {
+		super( driver, By.css( '#main.site-main' ) );
+	}
+
+	static async switchToPreview( driver ) {
```
alisterscott: we should do this without using the component and use our helpers like:

```
await driverHelper.waitForNumberOfWindows( driver, 2 );
await driverHelper.switchToWindowByIndex( driver, 1 );
```

alisterscott:
> No longer seeing tags when previewing(not sure if I'm just missing it)

Currently working for me so I've enabled this section

PR #1635, opened 2018-11-18
Update config

alisterscott:
New encrypt key


bsessions85:
Tried a new key so this could be done. It is passing now so merging it in

alisterscott:
Thanks

PR #1638, opened 2018-11-19
Gutenberg: Public and Private pages

bsessions85:
Adding tests for public pages and private pages

Notes:
- Sharing settings are not in Gutenberg yet, so related are skipped
- Post visibility is still not working on mobile, so mobile Private pages test is skipped


bsessions85:
Just jetpack tests failing. Disregard that failure

lib/gutenberg/gutenberg-editor-component.js
```diff
@@ -229,4 +233,17 @@ export default class GutenbergEditorComponent extends AsyncBaseContainer {
 		);
 		return hasPublishButton && ! hasRevertButton;
 	}
+
+	async viewPublishedPostOrPage( { reloadPageTwice = false } = {} ) {
```
alisterscott: we don't need the reload twice thing (for now)

lib/gutenberg/gutenberg-editor-component.js
```diff
@@ -229,4 +233,17 @@ export default class GutenbergEditorComponent extends AsyncBaseContainer {
...
+		const viewPostSelector = By.css( '.components-notice__content a' );
+		const driver = this.driver;
+
+		await driverHelper.waitTillPresentAndDisplayed( this.driver, viewPostSelector );
+
+		let url = await driver.findElement( viewPostSelector ).getAttribute( 'href' );
```
alisterscott: we only did this previously since it opened it a new window and it saved us messing around
since this doesn't open a new window can we just click the link? Simpler and also more representative of real use

lib/gutenberg/gutenberg-editor-component.js
```diff
@@ -229,4 +233,17 @@ export default class GutenbergEditorComponent extends AsyncBaseContainer {
...
+		if ( reloadPageTwice === true ) {
```
alisterscott: as above don't need this

PR #1639, opened 2018-11-20
Gutenberg: Insert a contact form: @parallel

JavonDavis:
There's no contact form block but I used a shortcode to simulate similar functionality. The shortcode doesn't render in the preview so there's no way to check in that step. Think this could be a suitable substitute until we have an official contact form block.

### DIFF
--- a/lib/gutenberg/gutenberg-editor-component.js
+++ b/lib/gutenberg/gutenberg-editor-component.js
@@ -73,6 +73,14 @@ export default class GutenbergEditorComponent extends AsyncBaseContainer {
 		return await this.driver.findElement( textSelector ).sendKeys( text );
 	}
 
+	async insertContactForm() {
+		await this.addBlock( 'Shortcode' );
+		const shortcodeSelector = By.css( 'textarea.editor-plain-text' );
+		await driverHelper.waitTillPresentAndDisplayed( this.driver, shortcodeSelector );
+		const shortcodeTextarea = this.driver.findElement( shortcodeSelector );
+		shortcodeTextarea.sendKeys( '[contact-form][/contact-form]' );
+	}
+
 	async errorDisplayed() {
 		await this.driver.sleep( 1000 );
 		return await driverHelper.isElementPresent( this.driver, By.css( '.editor-error-boundary' ) );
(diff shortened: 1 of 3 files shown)


alisterscott:
Thanks for this @JavonDavis 

This works well

How are we going to remember to update this test - or add a new one - when the native contact form block is live (should be soon when Jetpack is updated)?

JavonDavis:
>How are we going to remember to update this test - or add a new one - when the native contact form block is live (should be soon when Jetpack is updated)?

@alisterscott  for now, I think the best thing is just to open an issue which I've done here, https://github.com/Automattic/wp-e2e-tests/issues/1640 but there might be a creative way to use slack to help out in situations like this with a ping of some sort, I'll look into it.

Stojdza:
@JavonDavis Will test fail when native contact form block go live? If yes, that would be good notification :) If no, I suppose it's not urgent in that case, but we should be aware of the change.

JavonDavis:
> @JavonDavis Will test fail when native contact form block go live? If yes, that would be good notification :) If no, I suppose it's not urgent in that case, but we should be aware of the change.

@Stojdza Haha, that would actually be quite an effective one but I'm thinking it won't since I don't see any plans for disabling that shortcode, something like that would affect a lot of existing sites. Once the contact form block is out we could just repurpose this test as an additional test for the shortcode block 🤷‍♂️

lib/gutenberg/gutenberg-editor-component.js
```diff
@@ -73,6 +73,14 @@ export default class GutenbergEditorComponent extends AsyncBaseContainer {
 		return await this.driver.findElement( textSelector ).sendKeys( text );
 	}
 
+	async insertContactForm() {
+		await this.addBlock( 'Shortcode' );
+		const shortcodeSelector = By.css( 'textarea.editor-plain-text' );
+		await driverHelper.waitTillPresentAndDisplayed( this.driver, shortcodeSelector );
+		const shortcodeTextarea = this.driver.findElement( shortcodeSelector );
+		shortcodeTextarea.sendKeys( '[contact-form][/contact-form]' );
```
Stojdza: @JavonDavis Sorry for the late review! 

We should await `findElement` and `sendKeys` since they are returning premises. And return `sendKeys` also.

JavonDavis: 🤦‍♂️ You're right I'll send through a patch

simison:
We now have official Contact form block!  🎉  

It's behind a feature flag on wpcom just like Gutenberg itself (`gutenberg/block/jetpack-preset`), but you can see it here:

https://wpcalypso.wordpress.com/gutenberg/post/

image

Would be amazing to get it covered by e2e test as a follow up for this PR.

PR #1646, opened 2018-11-20
Gutenberg: Updating sections for permalink

bsessions85:
The permalink section is new and it takes a couple seconds for it to show up.

Updated the tab index and added a sleep to wait for it.

Can't wait until we can add some better identifiers on these!


bsessions85:
Merging in to get tests passing

PR #1648, opened 2018-11-21
Update the way we dismiss the Gutenberg NUX popup

alisterscott:
This hides it in the DOM after the page loads automatically


alisterscott:
@Automattic/flowpatrol @brbrr  heads-up I'll merge this to avoid conflicts - in case there's any issues with this approach

PR #1649, opened 2018-11-21
Update sidebar in Gutenberg to select by text instead of index

alisterscott:
The editor sidebar in Gutenberg currently uses an index based approach, the issue is the indexes change between state, eg. a saved draft has a different sidebar to an unsaved draft.

Also, wpcalypso.wordpress.com/gutenberg has a different sidebar order :(

screen shot 2018-11-21 at 5 08 22 pm

This PR updates the sidebar to select by text - not ideal but no other choice at present AFAIK

### DIFF
--- a/lib/gutenberg/gutenberg-editor-sidebar-component.js
+++ b/lib/gutenberg/gutenberg-editor-sidebar-component.js
@@ -79,7 +79,11 @@ export default class GutenbergEditorSidebarComponent extends AsyncBaseContainer 
 	}
 
 	async _expandOrCollapseSectionByText( text, expand = true ) {
-		const sectionSelector = By.xpath( `//button[text()="${ text }"]` );
+		const sectionSelector = await driverHelper.getElementByText(
+			this.driver,
+			By.css( '.components-panel__body-toggle' ),
+			text
+		);
 		await driverHelper.waitTillPresentAndDisplayed( this.driver, sectionSelector );
 		const sectionButton = await this.driver.findElement( sectionSelector );
 		let c = await sectionButton.getAttribute( 'aria-expanded' );
--- a/lib/gutenberg/gutenberg-editor-sidebar-component.js
+++ b/lib/gutenberg/gutenberg-editor-sidebar-component.js
@@ -7,23 +7,15 @@ import * as driverManager from '../driver-manager';
 
 export default class GutenbergEditorSidebarComponent extends AsyncBaseContainer {
 	constructor( driver ) {
-		super( driver, By.css( '.block-editor.gutenberg' ) );
+		super( driver, By.css( '.edit-post-header' ) );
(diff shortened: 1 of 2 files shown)


alisterscott:
@Stojdza I tried using driverHelper select by text but I couldn't work out how to just return the element instead of clicking it - do you reckon that would work?

The xpath approach seems to work fine though

lib/gutenberg/gutenberg-editor-sidebar-component.js
```diff
@@ -7,91 +7,81 @@ import * as driverManager from '../driver-manager';
 
 export default class GutenbergEditorSidebarComponent extends AsyncBaseContainer {
 	constructor( driver ) {
-		super( driver, By.css( '.block-editor.gutenberg' ) );
+		super( driver, By.css( '.edit-post-header' ) );
```
alisterscott: This is so this works on both gutenberg in calypso and wp-admin

lib/gutenberg/gutenberg-editor-sidebar-component.js
```diff
@@ -7,91 +7,81 @@ import * as driverManager from '../driver-manager';
...
-			By.css( '.components-panel__body-toggle' )
-		);
-		return await this.driver.findElements( By.css( '.components-panel__body-toggle' ) );
-	}
-
 	async selectTab( name ) {
 		return await driverHelper.clickWhenClickable(
 			this.driver,
-			By.css( `.edit-post-sidebar__panel-tab[data-label='${ name }'` )
+			By.css( `.edit-post-sidebar__panel-tab[aria-label^=${ name }]` )
```
alisterscott: this is to support calypso gutenberg also

Stojdza:
@alisterscott Yes, the idea for `selectBytext()` was to find element by text between several elements with the same locator and click on it. 

Maybe you can try something similar and add a new method in `driverHelper`, something like:
```
export async function getElementByText( driver, selector, text ) {
	return async () => {
		let allElements = await driver.findElements( selector );
		return await webdriver.promise.filter( allElements, async e => ( await e.getText() ) === text );
	};
}
```
which will return element, and then you can change [this](https://github.com/Automattic/wp-e2e-tests/pull/1649/files#diff-f93cdde636773327ae99d505a36b80daR82) line with:
```
const sectionSelector = await driverHelper.getElementByText( this.driver, By.css( '.components-panel__body-toggle' ), text );
```

Let me know was this helpful.

alisterscott:
> Let me know was this helpful.

Thanks @Stojdza 

It works well - I just can't work out how that returns a By selector and not a webelement - it looks like it returns a webelement 🤔

Stojdza:
> It works well - I just can't work out how that returns a By selector and not a webelement - it looks like it returns a webelement

Yeah, it looks like that. But, should be enough for our needs.

PR #1650, opened 2018-11-21
Skip test if domains tests are disabled

Stojdza:
Test `Sign up for free subdomain site` wasn't affected when SKIP_DOMAIN_TESTS is true. As a result, only this test was constantly [failing](https://circleci.com/gh/Automattic/wp-e2e-tests/24445#tests/containers/0) when domain registration was unavailable.


Stojdza:
Merge this to prevent failing executions on master.

PR #1651, opened 2018-11-22
Add Private Post Spec

alisterscott:
Adds a private post spec for Gutenberg with a couple of bits that don't work

### DIFF
--- a/lib/gutenberg/gutenberg-editor-sidebar-component.js
+++ b/lib/gutenberg/gutenberg-editor-sidebar-component.js
@@ -97,6 +97,21 @@ export default class GutenbergEditorSidebarComponent extends AsyncBaseContainer 
 		}
 	}
 
+	async setCommentsPreference( { allow = true } = {} ) {
+		const labelSelector = await driverHelper.getElementByText(
+			this.driver,
+			By.css( '.components-checkbox-control__label' ),
+			'Allow Comments'
+		);
+		const checkBoxSelectorID = await this.driver.findElement( labelSelector ).getAttribute( 'for' );
+		const checkBoxSelector = By.id( checkBoxSelectorID );
+		if ( allow === true ) {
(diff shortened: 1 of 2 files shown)


lib/gutenberg/gutenberg-editor-sidebar-component.js
```diff
@@ -97,6 +97,21 @@ export default class GutenbergEditorSidebarComponent extends AsyncBaseContainer
 		}
 	}
 
+	async setCommentsPreference( { allow = true } = {} ) {
```
Stojdza: Nicely done! I couldn't succeed to click on the checkbox with given id and class selector

specs/wp-gutenberg-post-editor-spec.js
```diff
@@ -368,129 +367,130 @@ describe( `[${ host }] Gutenberg Editor: Posts (${ screenSize })`, function() {
...
 			step( 'Can allow comments', async function() {
-				let postEditorSidebarComponent = await PostEditorSidebarComponent.Expect( driver );
-				await postEditorSidebarComponent.expandMoreOptions();
-				await postEditorSidebarComponent.setCommentsForPost( true );
+				const gEditorComponent = await GutenbergEditorComponent.Expect( driver );
+				await gEditorComponent.openSidebar();
+				const gEditorSidebarComponent = await GutenbergEditorSidebarComponent.Expect( driver );
+				await gEditorSidebarComponent.selectDocumentTab();
+				await driver.sleep( 3000 );
+				await gEditorSidebarComponent.collapseStatusAndVisibility(); // Status and visibility starts opened
```
Stojdza: Do we need to collapse Status&Visibility? I think it'll work fine if it stays open.

alisterscott: I think this is because we don't have a scroll into view and on mobile it may cause issues if it's open - might be safer to just close it

PR #1652, opened 2018-11-22
Add Gutenberg in calypso canary test

alisterscott:
This is the beginning of running specs against Gutenberg in Calypso

This works by using our existing simple site that is opted into Gutenberg but then using the route to access the Calypso Gutenberg editor directly,

In the case of running against WordPress.com it will use https://wpcalypso.wordpress.com/gutenberg/post

If this is running against calypso.localhost or calypso.live (or another environment) it will use the direct route `/gutenberg/post` but not change the environment

### DIFF
--- a/lib/data-helper.js
+++ b/lib/data-helper.js
@@ -252,9 +252,13 @@ export function getTestDomainRegistarDetails( emailAddress ) {
 	};
 }
 
-export function getCalypsoURL( route = '', queryStrings = [] ) {
+export function getCalypsoURL( route = '', queryStrings = [], { forceWpCalypso = false } = {} ) {
 	let queryString = '';
-	let url = this.configGet( 'calypsoBaseURL' ) + '/' + route;
+	let baseURL = this.configGet( 'calypsoBaseURL' );
+	if ( forceWpCalypso && baseURL === 'https://wordpress.com' ) {
+		baseURL = 'https://wpcalypso.wordpress.com';
+	}
+	let url = baseURL + '/' + route;
 
 	for ( let qs of queryStrings ) {
 		queryString = this._appendQueryString( queryString, qs );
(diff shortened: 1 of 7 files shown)


alisterscott:
@Copons would you mind taking a look at this to see if you think it'll work?

Stojdza:
I can't figure out why, but `Basic Public Post` from _wp-calypso-gutenberg-post-editor-spec.js_ constantly failing for me locally on [this](https://github.com/Automattic/wp-e2e-tests/pull/1652/files#diff-e14d8c59c7736fc1cbc095814d380609R36) line with the error `Unable to locate element: {"method":"css selector","selector":".post-publish-panel__postpublish-header a"}`. Actually, it doesn't click in [this](https://github.com/Automattic/wp-e2e-tests/pull/1652/files#diff-e14d8c59c7736fc1cbc095814d380609R27) step and that's why it's failing. When I put `await this.driver.sleep( 2000 );` before that click it works fine. 
I've tried with `waitTillPresentAndDisplayed` and similar methods, but it didn't help. Do you have similar problems locally? It passes on CircleCI.

JavonDavis:
> I can't figure out why, but `Basic Public Post` from _wp-calypso-gutenberg-post-editor-spec.js_ constantly failing for me locally on [this](https://github.com/Automattic/wp-e2e-tests/pull/1652/files#diff-e14d8c59c7736fc1cbc095814d380609R36) line with the error `Unable to locate element: {"method":"css selector","selector":".post-publish-panel__postpublish-header a"}`. Actually, it doesn't click in [this](https://github.com/Automattic/wp-e2e-tests/pull/1652/files#diff-e14d8c59c7736fc1cbc095814d380609R27) step and that's why it's failing. When I put `await this.driver.sleep( 2000 );` before that click it works fine.
> I've tried with `waitTillPresentAndDisplayed` and similar methods, but it didn't help. Do you have similar problems locally? It passes on CircleCI.

I'm having the same problem locally, I'm thinking it's due to the locator used here https://github.com/Automattic/wp-e2e-tests/blob/5e136a05e92c03a042d6e65a1c2b5d6aaf295f89/lib/gutenberg/gutenberg-editor-component.js#L29 I've opened in a PR for this https://github.com/Automattic/wp-e2e-tests/pull/1654

PR #1653, opened 2018-11-22
Fix uploading images and adding alt text

alisterscott:
This no longer requires a sleep as it waits for the image to be uploaded first

See https://github.com/WordPress/gutenberg/issues/12202


lib/gutenberg/gutenberg-editor-sidebar-component.js
```diff
@@ -229,4 +229,14 @@ export default class GutenbergEditorSidebarComponent extends AsyncBaseContainer
 		await driverHelper.waitTillPresentAndDisplayed( this.driver, trashSelector );
 		return await driverHelper.clickWhenClickable( this.driver, trashSelector );
 	}
+
+	async enterImageAltText( fileDetails ) {
+		const altTextInputSelector = By.css( '.components-textarea-control__input' );
```
alisterscott: This is still pretty generic and could cause non-determinism later on

PR #1662, opened 2018-11-27
Add related posts block test

alisterscott:
This doesn't seem to work yet as it doesn't display on the site's front end but this PR will be ready for when this block is ready


alisterscott:
I'm going to close this and re-add when it's not in dev

PR #1663, opened 2018-11-27
Remove trailing new lines from quotes

alisterscott:
These cause issues in Gutenberg


alisterscott:
Merging this as it is failing on master

Issue #1640, opened 2018-11-20
Update Gutenberg contact from test to use contact form block


JavonDavis:
Currently, the Gutenberg contact form test from https://github.com/Automattic/wp-e2e-tests/pull/1639 uses a shortcode to insert the contact form block. Once the native contact form block is live we should use that or at the very least add to the existing test.

JavonDavis:
Nice! Thanks @simison! 

@gwwar any idea when this will be available from Gutenberg in WPCOM? 

I did test this out on a site with Gutenberg and Jetpack enabled and the form wasn't showing in the preview and publish.. I've attached screenshots and a recording link below. Ignoring the last part of the recording where it navigates back to the wp-admin vs wpcalypso environment was there more information I needed to enter to see it in the Preview and Publish? 


<img width="1678" alt="screen shot 2018-11-22 at 11 08 44 am" src="https://user-images.githubusercontent.com/6520461/48913838-9e41c800-ee47-11e8-951b-0856df40daba.png">
<img width="1680" alt="screen shot 2018-11-22 at 11 09 24 am" src="https://user-images.githubusercontent.com/6520461/48913844-a13cb880-ee47-11e8-9d5c-6e2a69a40780.png">
<img width="1679" alt="screen shot 2018-11-22 at 11 09 42 am" src="https://user-images.githubusercontent.com/6520461/48913847-a39f1280-ee47-11e8-96fa-8ac888f792bd.png">
<img width="1679" alt="screen shot 2018-11-22 at 11 10 11 am" src="https://user-images.githubusercontent.com/6520461/48913849-a6016c80-ee47-11e8-8d56-4b69eb3f7975.png">

Recording link: https://www.dropbox.com/s/6rtk1ipfmy5mr0x/contact-form.mov?dl=0

gwwar:
@lezama or @johnHackworth would be able to better answer that for block timelines.

lezama:
We are aiming for early next week  👍

JavonDavis:
Closed in https://github.com/Automattic/wp-e2e-tests/pull/1695

PR #1664, opened 2018-11-27
Contact Form on wpcalypso 

JavonDavis:
Included contact form test on wpcalypso 

Related: #1640 

Part of #1620 

This test is currently failing since the after the contact form is added via the editor it doesn't show in Preview/Publish. Once this is ready in wpcalypso the test should pass and we can merge.

### DIFF
--- a/lib/gutenberg/gutenberg-editor-component.js
+++ b/lib/gutenberg/gutenberg-editor-component.js
@@ -78,12 +78,21 @@ export default class GutenbergEditorComponent extends AsyncBaseContainer {
 		return await shortcodeTextarea.sendKeys( shortcode );
 	}
 
-	async insertContactForm() {
-		await this.addBlock( 'Shortcode' );
-		const shortcodeSelector = By.css( 'textarea.editor-plain-text' );
-		await driverHelper.waitTillPresentAndDisplayed( this.driver, shortcodeSelector );
-		const shortcodeTextarea = await this.driver.findElement( shortcodeSelector );
-		return await shortcodeTextarea.sendKeys( '[contact-form][/contact-form]' );
+	async insertContactForm( email, subject ) {
+		await this.addBlock( 'Form' );
+		const emailSelector = By.css( '#inspector-text-control-0' );
+		const subjectSelector = By.css( '#inspector-text-control-1' );
+		await driverHelper.waitTillPresentAndDisplayed( this.driver, emailSelector );
+		await driverHelper.waitTillPresentAndDisplayed( this.driver, subjectSelector );
+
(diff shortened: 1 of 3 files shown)


JavonDavis:
Also replaced the use original test using `insertContactForm` to now use `insertShortcode` and created a function that can be used to insert shortcodes in general. If this is needed across other tests I can disable the wpcalypso test and merge.

lib/gutenberg/gutenberg-editor-component.js
```diff
@@ -70,12 +70,29 @@ export default class GutenbergEditorComponent extends AsyncBaseContainer {
...
+	async insertShortcode( shortcode ) {
 		await this.addBlock( 'Shortcode' );
 		const shortcodeSelector = By.css( 'textarea.editor-plain-text' );
 		await driverHelper.waitTillPresentAndDisplayed( this.driver, shortcodeSelector );
 		const shortcodeTextarea = await this.driver.findElement( shortcodeSelector );
-		return await shortcodeTextarea.sendKeys( '[contact-form][/contact-form]' );
+		return await shortcodeTextarea.sendKeys( shortcode );
+	}
+
+	async insertContactForm( email, subject ) {
```
Stojdza: @JavonDavis I propose to extract this as a separate block. We already have some blocks [here](https://github.com/Automattic/wp-e2e-tests/tree/master/lib/gutenberg/blocks). It would be easier for re-use in the future and this file is getting bigger and bigger :)

lib/gutenberg/gutenberg-editor-component.js
```diff
@@ -70,12 +70,29 @@ export default class GutenbergEditorComponent extends AsyncBaseContainer {
 		return await this.driver.findElement( textSelector ).sendKeys( text );
 	}
 
-	async insertContactForm() {
+	async insertShortcode( shortcode ) {
```
Stojdza: I like this solution 👍I think there is no need to make separate PR only for this, but it would be useful in the future definitely.

lib/gutenberg/gutenberg-editor-component.js
JavonDavis: Thanks for the suggestion and I agree it'd be make it easier to reuse this way, I'll make an update for this

JavonDavis: I've moved contact form, shortcodes and images into separate blocks. Maybe we should get this merged in a little earlier before the contact form is ready?

Stojdza: Sounds good! Would you mind to make separate PR for these blocks?

PR #1677, opened 2018-11-29
Try scroll to tag

Stojdza:
https://github.com/Automattic/wp-e2e-tests/pull/1675 didn't solve the tag issue. This fix scroll post to the bottom and then check tag.


alisterscott:
As discussed on Slack the tags aren't being published so this is necessary

PR #1678, opened 2018-11-30
Update secure payment form CSS

alisterscott:
This is an update for https://github.com/Automattic/wp-calypso/pull/28884

Since this is backwards compatible it can be merged any time


alisterscott:
All tests passing against this https://github.com/Automattic/wp-calypso/pull/28884 so merging

PR #1679, opened 2018-11-30
Try to get tagged post to publish on wpcalypso

alisterscott:
Work in Progress

Issues with the editor and saving tags raised here: https://github.com/Automattic/wp-calypso/issues/28974

### DIFF
--- a/lib/gutenberg/gutenberg-editor-sidebar-component.js
+++ b/lib/gutenberg/gutenberg-editor-sidebar-component.js
@@ -148,6 +148,12 @@ export default class GutenbergEditorSidebarComponent extends AsyncBaseContainer 
 		return await tagEntryElement.sendKeys( Key.ENTER );
 	}
 
+	async tagEventuallyDisplayed( tag ) {
+		//const selector = await driverHelper.getElementByText( this.driver, 'span', tag );
+		const selector = By.xpath( `//span[text() = "${ tag }"]` );
+		return await driverHelper.isEventuallyPresentAndDisplayed( this.driver, selector );
+	}
+
 	async _postInit() {
 		return await this.displayComponentIfNecessary();
 	}
--- a/lib/gutenberg/gutenberg-editor-sidebar-component.js
+++ b/lib/gutenberg/gutenberg-editor-sidebar-component.js
@@ -149,8 +149,7 @@ export default class GutenbergEditorSidebarComponent extends AsyncBaseContainer 
 	}
 
 	async tagEventuallyDisplayed( tag ) {
-		//const selector = await driverHelper.getElementByText( this.driver, 'span', tag );
-		const selector = By.xpath( `//span[text() = "${ tag }"]` );
+		const selector = await driverHelper.getElementByText( this.driver, By.css( 'span' ), tag );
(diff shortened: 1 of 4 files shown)


lib/gutenberg/gutenberg-editor-sidebar-component.js
```diff
@@ -148,6 +148,12 @@ export default class GutenbergEditorSidebarComponent extends AsyncBaseContainer
 		return await tagEntryElement.sendKeys( Key.ENTER );
 	}
 
+	async tagEventuallyDisplayed( tag ) {
+		//const selector = await driverHelper.getElementByText( this.driver, 'span', tag );
```
alisterscott: @Stojdza any ideas why this doesn't work? I've tried it locally and it doesn't work

specs/wp-calypso-gutenberg-post-editor-spec.js
```diff
@@ -152,6 +158,16 @@ describe( `[${ host }] Calypso Gutenberg Editor: Posts (${ screenSize })`, funct
 			);
 		} );
 
+		step( 'Can see the post tag in preview', async function() {
```
alisterscott: This should make this fail faster if it isn't saved in the post

specs/wp-gutenberg-post-editor-spec.js
```diff
@@ -152,6 +158,16 @@ describe( `[${ host }] Gutenberg Editor: Posts (${ screenSize })`, function() {
 			);
 		} );
 
+		step( 'Can see the post tag in preview', async function() {
```
alisterscott: This should make this fail faster if it isn't saved in the post

alisterscott:
This still isn't consistently passing on CI

lib/gutenberg/gutenberg-editor-sidebar-component.js
Stojdza: Yeah, selector should be `By.css( 'span' )` not just `'span'`. 

`getElementByText()` probably should be more clear about selectors - should we send, for example, `<className>` or `By.css( '.<className>' )`

Stojdza:
@alisterscott I disabled this same test for Gutenberg Editor also, it was failing on master https://github.com/Automattic/wp-e2e-tests/pull/1680

Can we disable this step where we are checking tag, and enable tests again (until https://github.com/Automattic/wp-calypso/issues/28974 is solved)?

lib/gutenberg/gutenberg-editor-sidebar-component.js
alisterscott: Yeah 🤦‍♂️

alisterscott:
> Can we disable this step where we are checking tag, and enable tests again (until Automattic/wp-calypso#28974 is solved)?

Should I disable checking it is added to the sidebar as well? Or just the front end?

Stojdza:
> Should I disable checking it is added to the sidebar as well? Or just the front end?

I would say just front end, since sidebar check is passing. I'll disable it and check results.

Stojdza:
I'll merge this now and see how's it going on master

Issue #1549, opened 2018-10-23
Support running a local PR against a wp-calypso PR


alisterscott:
When someone is working on a change in wp-calypso they can create the same branch name in this repository which means this PR will be used for tests against the PR

But we should also support running against a PR from CircleCI on the change itself

Something like

1. When running in CircleCI for this project on non-master branch - check the branch exists on wp-calypso
2. If exact same branch exists - then change URL to calypso.live and set branch name and use that

Question:

Should we require that a label is added to the wp-e2e-tests PR to indicate it is matching and for this behaviour to activate - or should it be automatic?

PR #1681, opened 2018-11-30
Run tests against calypso.live if match exists in wp-calypso

bsessions85:
This PR does a check to see if the branch running has a branch with the same name in wp-calypso. If so, it takes the sha and runs it against that hash on calypso.live. This check is not done if running on master.

part of #1549


JavonDavis:
@bsessions85 it seems to be timing out across the other e2e-tests repo, is that expected?

PR #1686, opened 2018-12-04
Wait for radio button to show

Stojdza:
Tests were [failing](https://circleci.com/gh/Automattic/wp-e2e-tests/24805) today at this [line](https://github.com/Automattic/wp-e2e-tests/blob/master/specs/wp-signup-spec.js#L849) due to slower loading of Checkout page.
This fix ensures that test is waiting for the radio button and in case of failing we will see better error message.


bsessions85:
Merging this in

PR #1687, opened 2018-12-04
Update document tab selector

bsessions85:
The label changed in one of the gutenberg versions. Updated selector and updated the function name to remove a type while I was at it.


JavonDavis:
@bsessions85 it seems the Trash Post test and a few others are still failing here... is this still related to the new update?

bsessions85:
@JavonDavis  I am not sure. That one was passing right before this one so I am not sure. There also seems to be another issue with the cart but I haven't gotten around to investigating.

PR #1690, opened 2018-12-05
Remove categories from classic editor spec

alisterscott:
This has been really problematic so removing this for now as we move towards the block editor anyway


Stojdza:
Will merge this now.

PR #1691, opened 2018-12-05
Remove visiting gutenberg page in wp-calypso

alisterscott:
Since this is enabled in wpcalypso now this switches the URL and then uses regular nav


alisterscott:
This didn't work as expected

PR #1693, opened 2018-12-05
A/B Test: add new A/B test for simplified onboarding checklist

danhauk:
Add a new A/B test to simplify the checklist by hiding some details.

Test in https://github.com/Automattic/wp-calypso/pull/29041

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -68,7 +68,8 @@
 		"improvedOnboarding",
 		"privateByDefault",
 		"crowdsignalNameBasedSignup",
-		"krackenM5DomainSuggestions"
+		"krackenM5DomainSuggestions",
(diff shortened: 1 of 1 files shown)


config/default.json
```diff
@@ -68,7 +68,8 @@
 		"improvedOnboarding",
 		"privateByDefault",
 		"crowdsignalNameBasedSignup",
-		"krackenM5DomainSuggestions"
+		"krackenM5DomainSuggestions",
+		"simplifiedChecklistView",
```
bsessions85: The comma after this should be removed. It causes things to break, which is why the canaries are failing

danhauk: Whoops, thanks for catching that!

PR #1694, opened 2018-12-06
Adds a save step to the payment button test

alisterscott:
This ensures the publish button works

Also enables payment button tests for wp-admin for the time being


alisterscott:
Merging this as this test often fails on master without the deliberate save

Issue #995, opened 2018-02-28
Investigate instability in 2fa login tests


hoverduck:
We've been getting a lot of failures on the 2fa login tests recently, so this issue is to drive an investigation and figure out what we can do to make them better.

Here are a couple of the error messages I've seen crop up:
`Your session has expired, please go back to the login screen`
![2fa-session-expired](https://user-images.githubusercontent.com/7233112/36816721-a992e116-1cac-11e8-90c7-108c4428f4fd.png)

`You can only request a code via SMS once per minute. Please wait and try again.`
![2fa-sms-once-per-minute](https://user-images.githubusercontent.com/7233112/36816723-ab083f78-1cac-11e8-9368-219cfd4d952a.png)

The second one can particularly be an issue with the way we're running the tests and picking accounts at random.  The hope is that the randomness coupled with the auto-retry will be enough to mask/bypass the problem, but maybe that's not enough.

alisterscott:
I think this would be likely to be caused by having multiple parallel runs and 'clashes' between accounts running.

Could we use some new account only signups to then add 2fa and test a log on with it?

cc @Tug

alisterscott:
One other thing I thought of is these specs don't have `@parallel` tags on each one so if one fails the whole lot gets re-run which could cause potential clashes.

We should investigate if we can add `@parallel` tags to them to stop multiple re-runs happening

hoverduck:
> We should investigate if we can add @parallel tags to them to stop multiple re-runs happening

I gave this a try briefly earlier today and it seemed to work fine, but I was side-tracked by other things before I could push a PR.  I had a question for @rachelmcr whether moving the `@visdiff` tag lower in the hierarchy would cause problems, though.

rachelmcr:
>I had a question for @rachelmcr whether moving the @visdiff tag lower in the hierarchy would cause problems, though.

Shouldn't be a problem. The only visdiff screenshots in this spec are in the first `Logging In and Out:` test, so that should be the only one needing a `@visdiff` tag. Moving the before and after hooks for `eyesHelper` into that test should be all that's needed.

JavonDavis:
Updated these tests in https://github.com/Automattic/wp-e2e-tests/pull/1407 to only run on a specific label.

PR #1698, opened 2018-12-09
Remove sleep statements from `login-page.js`.

juliaamosova:
In this PR, I removed 5 sleep statements and refactored the code to substitute for them. 

Working only some of the time, **sleep statements are not deterministic**. As a result, we end up with fragile tests that break under certain conditions.

**Sleep 1:**

Refactored `waitTillPresentAndDisplayed()` and removed `sleep()` stament that followed it. 

Instead, I used `driver.wait()` method to wait for up to 20 seconds for the element to be found in the DOM. If the element is added to the DOM in one second, the `driver.wait()` method will complete in one second. Unlike `driver.sleep()`, it will not wait the full amount of time specified. 

Another advantage of using `driver.wait()` method is that in case of an error, the error message will be shown specifying what went wrong. 

**Sleep 2:**

Removed sleep statements after `await driverHelper.setWhenSettable( driver, userNameSelector, username );`. It is redundant because the statement below completes user's input. 

**Sleep 3:**

Removed and replaced with `driver.wait()` method. The explanation is the same as for Sleep 1. 

**Sleep 4:**

Same explanation as for Sleep 2.

**Sleep 5:**

Removed sleep statement as it is redundant because the statement below `waitTillNotPresent()` performs the waiting needed. 

-----------

I also doubled the wait time in `await driverHelper.waitTillNotPresent( driver, userNameSelector, this.explicitWaitMS * 2 );` as the test was failing locally due to timeout. Sometimes, it takes up to 30 seconds for the test to be completed on my machine:

Image Link: https://cloudup.com/cCAuLGMqJ8d

Short video showing the test passing locally:

http://g.recordit.co/YBqADKXgJE.gif

I tested on tablet and mobile locally as well - passed.


lib/pages/login-page.js
```diff
@@ -21,29 +22,32 @@ export default class LoginPage extends AsyncBaseContainer {
 		const userNameSelector = By.css( '#usernameOrEmail' );
 		const passwordSelector = By.css( '#password' );
 
-		await driverHelper.waitTillPresentAndDisplayed( driver, userNameSelector );
-		await this.driver.sleep( 1000 );
+		await this.driver.wait(
```
bsessions85: If this really does work better than `waitTillPresentAndDisplayed`, then it would probably be better to update that rather than just fixing in this one spot

lib/pages/login-page.js
```diff
@@ -21,29 +22,32 @@ export default class LoginPage extends AsyncBaseContainer {
...
+			this.explicitWaitMS,
+			'Could not locate the username or email field.'
+		);
+
 		await driverHelper.setWhenSettable( driver, userNameSelector, username );
-		await this.driver.sleep( 1000 );
 		await driver.findElement( userNameSelector ).sendKeys( Key.ENTER );
 
-		await this.driver.sleep( 1000 );
+		await this.driver.wait(
```
bsessions85: Would `waitTillPresentAndDisplayed` work here instead? Reuse is always preferred

JavonDavis:
Additionally I see you're working on the Login specs, there was some instability with these noted here -> https://github.com/Automattic/wp-e2e-tests/issues/995. As a result, these tests aren't ran in the full suite in the CI anymore to avoid errors due to the session still being active. The `Needs Secure Auth e2e Testing` label triggers these tests to run in the Ci. That being said when you're testing locally you'll need to space out your executions to avoid this or disable the 2fa tests (marked with @secure-auth) if you're testing more frequently.

juliaamosova:
>Additionally I see you're working on the Login specs, there was some instability with these noted here -> #995. As a result, these tests aren't ran in the full suite in the CI anymore to avoid errors due to the session still being active. The Needs Secure Auth e2e Testing label triggers these tests to run in the Ci. That being said when you're testing locally you'll need to space out your executions to avoid this or disable the 2fa tests (marked with @secure-auth) if you're testing more frequently.

Thanks for these details and insight @JavonDavis, very helpful and good to know. Thanks for adding the `Needs Secure Auth e2e Testing` label as well to trigger the test!

lib/pages/login-page.js
juliaamosova: Thanks for double checking on this @bsessions85. I've done additional research and confirm that it is better to use `waitTillPresentAndDisplayed` as it provides more coverage.

- `driver.wait(until.elementLocated)` - checks that an element is _present_ within the HTML DOM of a page. This does not necessarily mean that the element is _visible_. So there's no guarantee that it's _interactable_.

- `waitTillPresentAndDisplayed` function - includes `driver.findElement()` and `element.isDisplayed()` - in other words, it checks that an element is _present_ within the HTML DOM of a page and is _visible_. This does not necessarily mean that the element is _interactable_ or _clickable_ but at least it is already _visible_. 

>Part of the problem with removing sleeps is that they are often times put in on a day when things are running slower or the tests start failing suddenly. A slight delay is added and things start passing again. That being said, it is important to think through what might be happening that caused the sleep to be added in the first place.

`waitTillPresentAndDisplayed` is followed by `setWhenSettable` which finds the element, clears the text from the textarea and enters the value. 

I checked why the sleep statement was added here https://github.com/Automattic/wp-e2e-tests/pull/1528 - tests run too fast and a delay is added. While there is no explanation of why the test is failing exactly, the reason could be that there is no check if the cursor had been placed in the input field or if the input field has a focus. If the test runs too fast, the input field may be disaplyed but not in focus yet while the driver is trying to clear it already. 

I am working on testing my theory above. Will submit another PR as soon as I am done.

lib/pages/login-page.js
juliaamosova: Yes, as per my research in https://github.com/Automattic/wp-e2e-tests/pull/1698#discussion_r240554182, it is better to use `waitTillPresentAndDisplayed`. I will address it in the next PR. Thanks @bsessions85!

lib/driver-helper.js
```diff
@@ -45,6 +45,55 @@ export function clickWhenClickable( driver, selector, waitOverride ) {
 	);
 }
 
+export function clickWhenFocused( driver, selector, pollingOverride, waitOverride ) {
```
juliaamosova: As per my comment here https://github.com/Automattic/wp-e2e-tests/pull/1698#discussion_r240554182 the reason why the tests that run too fast are failing could be that there is no check if the cursor had been placed in the input field or if the input field has a focus. The input field may be disaplyed but not in focus yet while the driver is trying to clear it already.

To address this issue I wrote a new `driverHelper` method `clickWhenFocused`. It is based on the following steps:

- Find element;
- Make it active (in focus) by switching to it;
- Make sure that this element equals the element on which the next action must be performed;
- Click on that element.

The focus is implemented by using `switchTo().activeElement()`. 

`activeElement()` switches to the element that currently has focus within the document currently "switched to", or the body element if this cannot be detected. See reference [here](https://seleniumhq.github.io/selenium/docs/api/java/org/openqa/selenium/WebDriver.TargetLocator.html#activeElement--). 

Once the element is in focus we need to make sure that it matches the element that needs to be clicked on. To do that, we can utilize `getId()` which is a server assigned `Id` to the element of the DOM of the particualr session (not the HTML `Id`) as per reference [here](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/lib/webdriver.js#L1829).  

@bsessions85 let me know your thoughts. Also cc @Stojdza as you wanted to work on re-writing `clickWhenClickable` with a similar approach.

lib/pages/login-page.js
```diff
@@ -22,28 +22,23 @@ export default class LoginPage extends AsyncBaseContainer {
 		const passwordSelector = By.css( '#password' );
 
 		await driverHelper.waitTillPresentAndDisplayed( driver, userNameSelector );
-		await this.driver.sleep( 1000 );
+		await driverHelper.clickWhenFocused( driver, userNameSelector );
```
juliaamosova: Implementing a new `driverHelper` method `clickWhenFocused` to address the sleep statement that was previously here.

lib/pages/login-page.js
```diff
@@ -22,28 +22,23 @@ export default class LoginPage extends AsyncBaseContainer {
...
 		await driverHelper.setWhenSettable( driver, userNameSelector, username );
-		await this.driver.sleep( 1000 );
 		await driver.findElement( userNameSelector ).sendKeys( Key.ENTER );
 
-		await this.driver.sleep( 1000 );
-
 		if ( emailSSO === false ) {
+			await driverHelper.waitTillPresentAndDisplayed( driver, passwordSelector );
+			await driverHelper.clickWhenFocused( driver, passwordSelector );
```
juliaamosova: Implementing `waitTillPresentAndDisplayed` and a new `driverHelper` method `clickWhenFocused` to address the sleep statement that was previously here.

lib/driver-helper.js
Stojdza: @juliaamosova This method is also applicable to buttons and other elements? If yes, did you consider to use it as `clickWhenClickable` method (rewrite existing one)? Those two look similar, and it's easier just to change the existing one and use it through all tests.

juliaamosova: @Stojdza I did consider re-writing `clickWhenClickable` but didn't want to use it before the review of the new method I wrote. I agree that keeping `clickWhenClickable` will be better. The method I wrote should be applicable to all types of elements. Let me test it locally and I will get back with a new PR.

juliaamosova: @Stojdza I refactored `clickWhenClickable` with the changes I introduced in my `clickWhenFocused`method. While it is working locally, I see that some tests are failing here mainly due to timeouts. I am looking to see what could be going on.

lib/driver-helper.js
```diff
@@ -45,6 +45,38 @@ export function clickWhenClickable( driver, selector, waitOverride ) {
 	);
 }
 
+export function waitTillFocused( driver, selector, pollingOverride, waitOverride ) {
```
juliaamosova: @bsessions85 and @Stojdza so far, I addressed 2 out of 5 sleep statements in this PR. I used `waitTillFocused` method that I wrote. I remove the `click` from it (originally had it) since it is not needed. 
I placed the remaining 3 sleep statements that I originally removed back for now.
I think it is best if I address the remaining 3 in this file in a separate PR since the existing one already has a long history and it is getting hard to follow.

lib/driver-helper.js
juliaamosova: @Stojdza after additional testing, I can't confirm that I can re-write `clickWhenClickable` with changes I introduced in `waitTillFocused` method. It seems to work in many cases but not always. I removed the `click` event from the  `waitTillFocused` so that it could be used to make sure the element is in focus only. It works great to address the sleep statements in `login-page.js`.

lib/pages/login-page.js
```diff
@@ -22,28 +22,27 @@ export default class LoginPage extends AsyncBaseContainer {
...
+		await driverHelper.waitTillFocused( driver, userNameSelector );
 		await driverHelper.setWhenSettable( driver, userNameSelector, username );
 		await this.driver.sleep( 1000 );
 		await driver.findElement( userNameSelector ).sendKeys( Key.ENTER );
 
-		await this.driver.sleep( 1000 );
-
 		if ( emailSSO === false ) {
+			await driverHelper.waitTillPresentAndDisplayed( driver, passwordSelector );
+			await driverHelper.waitTillFocused( driver, passwordSelector );
```
Stojdza: Is it necessary to have both methods, `waitTillPresentAndDisplayed` and `waitTillFocused`? It seems that we are waiting for the same element twice.

juliaamosova: @Stojdza I tested without `waitTillPresentAndDisplayed` here and on line 24 as well (https://github.com/Automattic/wp-e2e-tests/pull/1698/files#diff-d2cabd92e3f34ad1459c4dd1862e098fR24). It works without it. But I decided to leave it because `waitTillFocused` doesn't check if the element is displayed and `waitTillPresentAndDisplayed` has that check. I think to be extra sure that the element is displayed, it will be good to leave it. What do you think?

>I wanted to try what would happen if I put waitTillFocused inside clickWhenClickable and setWhenSettable but that did not turn out well :D

I did the same but just with `clickWhenClickable` and it was an epic fail.

juliaamosova: @Stojdza thanks! Does it mean it is good to merge?

Stojdza: Yes :)

juliaamosova: Cool, thanks!

PR #1707, opened 2018-12-11
Wait for publishSelector, then click on it

Stojdza:
In several occasions, we had [this](https://circleci.com/gh/Automattic/wp-e2e-tests/25033#tests/containers/1) failure. I succeeded to reproduce it locally, and it looks like [this](https://github.com/Automattic/wp-e2e-tests/blob/master/lib/gutenberg/gutenberg-editor-component.js#L31) line is passing, but the button is not actually clicked.
Hopefully, wait for this button first will make test more reliable. 

To test:
Make sure that `Calypso Gutenberg Editor: Posts: Insert a payment button` is passing. 
I ran tests several times on CircleCI and didn't see the error again.

### DIFF
--- a/specs/wp-calypso-gutenberg-post-editor-spec.js
+++ b/specs/wp-calypso-gutenberg-post-editor-spec.js
@@ -951,7 +951,7 @@ describe( `[${ host }] Calypso Gutenberg Editor: Posts (${ screenSize })`, funct
 		} );
 	} );
 
-	describe.only( 'Insert a payment button: @parallel', function() {
+	describe( 'Insert a payment button: @parallel', function() {
 		const paymentButtonDetails = {
 			title: 'Button',
 			description: 'Description',
--- a/specs/wp-calypso-gutenberg-post-editor-spec.js
+++ b/specs/wp-calypso-gutenberg-post-editor-spec.js
@@ -951,7 +951,7 @@ describe( `[${ host }] Calypso Gutenberg Editor: Posts (${ screenSize })`, funct
 		} );
 	} );
 
-	describe( 'Insert a payment button: @parallel', function() {
+	describe.only( 'Insert a payment button: @parallel', function() {
 		const paymentButtonDetails = {
 			title: 'Button',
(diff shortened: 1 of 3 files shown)


alisterscott:
Thanks for the PR @Stojdza 

I'm not sure this will make any difference since clickWhenClickable should wait for it to be there anyway

I think there's still some failures on this branch which are retried

Are you sure this isn't a real race-condition issue in the Gutenberg editor? 

What if you checked in a temp fix to this branch that disables the retry and see if it improves?

Stojdza:
@alisterscott Thanks for the review! 

> What if you checked in a temp fix to this branch that disables the retry and see if it improves?

I tried and unfortunately,[ many tests](https://circleci.com/gh/Automattic/wp-e2e-tests/tree/try%2Fwait-publish-selector) are failing when there are no retries 😬But, publish a post is not the problem in any of them. 

I've run tests numerous times with additional `waitTillPresentAndDisplayed`, and couldn't reproduce `.components-notice.is-success` error, both locally and CircleCI. When I disabled it, test failed locally in second attempt.

specs/wp-calypso-gutenberg-post-editor-spec.js
```diff
@@ -987,7 +987,8 @@ describe( `[${ host }] Calypso Gutenberg Editor: Posts (${ screenSize })`, funct
 
 		step( 'Can publish and view content', async function() {
 			const gEditorComponent = await GutenbergEditorComponent.Expect( driver );
-			return await gEditorComponent.publish( { visit: true } );
+			await gEditorComponent.publish( { visit: true } );
+			this.skip();
```
bsessions85: I am assuming this skip was just for testing.

Stojdza: 🤦‍♂️ Good guess :) Thanks for catching this!

Stojdza: Done.

PR #1708, opened 2018-12-12
Updates some dependencies

alisterscott:
Trying to get dependencies up to date to get rid of any messages when running

`npm audit`

This reduces how many issues we have but doesn't remove them all

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -827,9 +827,9 @@
       }
     },
     "JSONStream": {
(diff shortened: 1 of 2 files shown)


alisterscott:
> Too bad we still can't do mailosaur!

Thanks @bsessions85  - do you think we should give updating mailosaur another try? It's been a while - surely they've sorted their perf issues out

PR #1714, opened 2018-12-13
Add tests for the onboarding signup flow

taggon:
Even though new `onboarding` signup flow is still in A/B testing, it is a critical piece of WordPress.com and many current/future works are based on the new flow. Therefore, it would make sense to add some e2e tests to have additional layers of safety.

@Automattic/flowpatrol I'd appreciate if you take a look at this since I'm not 100% sure about the approach I used here to turn on the `onboarding` flow.

### DIFF
--- a/lib/override-abtest.js
+++ b/lib/override-abtest.js
@@ -1,0 +1,46 @@
+/** @format */
+
+import config from 'config';
+import { findIndex } from 'lodash';
+
+function getABTestEraser( name ) {
+	return () => {
+		const overrideABTests = config.get( 'overrideABTests' );
+		const index = findIndex( overrideABTests, item => item[ 0 ] === name );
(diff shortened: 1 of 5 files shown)


bsessions85:
Thanks for this @taggon! I really like what you've done here and I think it will be a valuable addition to our setup. We typically only run tests against a single, default version, but for something like this it makes sense to test both sides.  I don't have any better ideas on how to do this, so I think it is something we can go with and iterate on as we use it more. If there are cases that it doesn't quite work for, then we'll adjust.

taggon:
@bsessions85 Thanks for your review! I got a bit more confident. 😄

taggon:
@bsessions85 Could you approve this if you think it's eligible for merging?

PR #1721, opened 2018-12-16
Remove and address sleep statement in `close-account-page.js`

juliaamosova:
Removed the `sleep` statement as it was not deterministic. Click doesn't always fire even if the button is already displayed. 

In the loop that I implemented, we can safely attempt to click the button until the confirmation dialog pop-up window is present which indicates the end of the loop. The test can move on to the next step.

### DIFF
--- a/lib/pages/account/close-account-page.js
+++ b/lib/pages/account/close-account-page.js
@@ -5,6 +5,9 @@ import { By as by } from 'selenium-webdriver';
 import * as driverHelper from '../../driver-helper.js';
 
 import AsyncBaseContainer from '../../async-base-container';
+import config from 'config';
+
+const explicitWaitMS = config.get( 'explicitWaitMS' );
 
 export default class CloseAccountPage extends AsyncBaseContainer {
 	constructor( driver ) {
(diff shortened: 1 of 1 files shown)


juliaamosova:
I also wanted to add that I looked into rewriting `clickWhenClickable` implementing the loop but didn't think it is possible to apply it to all tests. Each element has its own conditions for when it is ready to be clicked and so it is hard to find a common denominator for them all. 

Finally, I've considered writing a new driverHelper method out of the loop as well but also didn't think it was possible. I looked at other tests with `sleep` statements where the same loop could be applied but the conditions for the loop there had to be written differently.

juliaamosova:
Thanks @Stojdza!

lib/pages/account/close-account-page.js
```diff
@@ -5,18 +5,32 @@ import { By as by } from 'selenium-webdriver';
...
-			by.css( '.account-close button.is-scary' )
-		);
+		const buttonSelector = by.css( '.account-close button.is-scary' );
+		const confirmDialogSelector = by.css( '.account-close__confirm-dialog-confirm-input' );
+		const pauseBetweenClickAttemptsMS = 100;
+
+		await driverHelper.waitTillPresentAndDisplayed( this.driver, buttonSelector );
+
+		// Click doesn't always fire even if the button is already displayed.
+		// We can safely attempt to click the button until the confirmation dialog pop-up window is present.
```
JavonDavis: Hmm did you look into the cause of why the click doesn't always fire? Seems weird to just keep trying to click a couple times hoping it does what it's supposed to when there might be an underlying issue, if you already did then it's fine to have this as a workaround until we can figure out what's up.

juliaamosova: >Hmm did you look into the cause of why the click doesn't always fire?

@JavonDavis I think the issue is Selenium and browser communication related. I've researched this topic extensively and found many sources where engineers face the the same problem. The most comprehensive explanation I found could be seen here: https://github.com/SeleniumHQ/selenium/issues/4075. 

**TL;DR:** The main reason is that element on which we try to click is not in “interactable” state. 

I've tried several things here:

1) Use JavaScript Executor to click on the button;
2) Use JavaScript Executor to focus on the button.

The 2 above made no difference. They work the same way as the Selenium driver (completed only when `sleep` was used). Moreover, they don't return anything (other than `undefined`) so we can't use it as a check if the click or focus actually happened. 

3) Next, I've tried to check if the DOM Content had been loaded before clicking on the button by using JavaScript executor as well. This didn't work as well. 

4) I also tried to check if the Close Account button ID matches with the currently active element ID but it also didn't work without `sleep`:

5) Finally, I also spent some time analyzing the button itself. I was trying to see if we could use some attributes of the button as a check if the button is ready to be clicked but there doesn't seem to be one. 

>Seems weird to just keep trying to click a couple times hoping it does what it's supposed to when there might be an underlying issue

I agree that this is not ideal but it is better than the `sleep` statement that was previously used. In my testing, it took about 200 ms ( ~ 2 click attempts) for the click to go through. It is a good deterministic approach since the loop will wait for the condition to be met prior to proceed. 

>if you already did then it's fine to have this as a workaround until we can figure out what's up.

I would love to get to the bottom of it at some point. I will keep this in mind as I keep working through other sleep statements. As I mentioned above though I think it is a Selenium / browser issue.

PR #1722, opened 2018-12-17
Remove 'domains step' tests as it will be removed from the onboarding flow.

taggon:
Ref. https://github.com/Automattic/wp-calypso/pull/29494
See also #1714


taggon:
1. Rebased.
2. Added an a/b test setting to `default.json` since the related Calypso PR has an a/b test now.

taggon:
This PR is going to be dropped since we need to hide the a/b test for all envs for now.

PR #1725, opened 2018-12-18
Try reducing the number of logins

bsessions85:
This PR saves the cookies needed for login and re-uses them if they exist. I also put a way to force it to log in, just in case we need it

### DIFF
--- a/lib/flows/login-flow.js
+++ b/lib/flows/login-flow.js
@@ -169,7 +169,7 @@ export default class LoginFlow {
 		return await sideBarComponent.selectAddPerson();
 	}
 
-	async loginAndSelectMySite( site = null, { useFreshLogin = false } ) {
+	async loginAndSelectMySite( site = null, { useFreshLogin = false } = {} ) {
 		await this.login( { useFreshLogin: useFreshLogin } );
 
 		const readerPage = await ReaderPage.Expect( this.driver );
--- a/lib/flows/login-flow.js
+++ b/lib/flows/login-flow.js
@@ -128,7 +128,7 @@ export default class LoginFlow {
 		usingGutenberg = false,
 		{ forceCalypsoGutenberg = false, useFreshLogin = false } = {}
 	) {
-		await this.loginAndSelectMySite( site );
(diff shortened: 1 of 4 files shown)


bsessions85:
Thanks @Stojdza!

> Cookie is reusable only for the current run, or for every next run, until expiration?

Since it is on temporary CI container, the cookie would only be reused for the tests within that build on that container. Every new build would do at least one login and then the rest of the tests would re-use the cookie

> What is happening when tests are running in parallel and there is no cookie?

Looking in to confirming this, but in theory, the parallel runs would also use the cookie as long as the cookies have been saved by a log in done somewhere

> When I ran wp-log-in-out-spec.js for the second time, it didn't work, it was stuck on login screen. ? Anyway, do you think that we should set useFreshLogin = true for this spec? Since it is testing log in and we should not skip it with cookie?

I will have to check this one, but I agree, the tests in here should always be fresh and not reusing the cookie

> Locally, tests are much faster when they are using cookie. But, in CircleCI execution time is similar to execution time on master branch. Do you know why there is no much difference?

Researching this one now

Stojdza:
> Looking in to confirming this, but in theory, the parallel runs would also use the cookie as long as the cookies have been saved by a log in done somewhere

I was thinking about the case when there is no cookie, and 8 tests try login and store cookie. Will every test store cooke, or just first one? If every test try to store cookie, is it the same cookie value or there will be 8 different ones?

bsessions85:
> I was thinking about the case when there is no cookie, and 8 tests try login and store cookie. Will every test store cookie, or just the first one? If every test tries to store cookie, is it the same cookie value or there will be 8 different ones?

Every test would have written to the same file so the last one would be the one to stick. I just updated it to only write if the file doesn't exist, so that should be better. 

Here is a debug view of the latest version. You can see where it is logging in versus where it is reusing the cookie- https://circleci.com/gh/Automattic/wp-e2e-tests/25351

Stojdza:
> Here is a debug view of the latest version

Looks great!

Stojdza:
Works well! @JavonDavis another pair of eyes would be helpful here and we can go on master after :)

Issue #1645, opened 2018-11-20
Gutenberg: Add comments test


bsessions85:
When development for Gutenberg settles down some, we will need to add some attributes and some tests to cover comments functionality. The selectors in there now are very generic.

PR #1729, opened 2018-12-19
Gutenberg Comments test

JavonDavis:
This PR adds a test for comments on a post made via Gutenberg Editor

To test:

`specs/wp-comments-spec.js`

### DIFF
--- a/specs/wp-comments-spec.js
+++ b/specs/wp-comments-spec.js
@@ -9,6 +9,7 @@ import LoginFlow from '../lib/flows/login-flow';
 import EditorPage from '../lib/pages/editor-page';
 import PostEditorToolbarComponent from '../lib/components/post-editor-toolbar-component';
 import CommentsAreaComponent from '../lib/pages/frontend/comments-area-component';
+import GutenbergEditorComponent from '../lib/gutenberg/gutenberg-editor-component';
 
 const host = dataHelper.getJetpackHost();
 const screenSize = driverManager.currentScreenSize();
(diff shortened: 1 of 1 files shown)


JavonDavis:
Since Gutenberg is released into prod now I thought it best to just add the comments tests within the wp-comments-spec vs wp-calypso-gutenberg, just to make it a little easier once we phase out the classic editor spec

bsessions85:
So, I should have been clearer on the issue, but what I meant was "Allow Comments" checkbox in the editor. We want to ensure that if that box isn't checked, then comments cannot be added.

JavonDavis:
> So, I should have been clearer on the issue, but what I meant was "Allow Comments" checkbox in the editor. We want to ensure that if that box isn't checked, then comments cannot be added.

Ahhh I see! Misunderstood I've removed the command from PR that would close #1645, @bsessions85 should we still merge this though?

specs/wp-comments-spec.js
```diff
@@ -71,4 +72,41 @@ describe( `[${ host }] Comments: (${ screenSize })`, function() {
...
+			const commentArea = await CommentsAreaComponent.Expect( driver );
+			return await commentArea._postComment( {
+				comment: dataHelper.randomPhrase(),
+				name: 'e2eTestName',
+				email: 'e2eTestName@test.com',
+			} );
+		} );
+
+		step( 'Can post a reply', async function() {
+			await driver.sleep( 10000 ); // Wait to not to post too quickly
```
Stojdza: Looks like you are creating work for @juliaamosova 😃

Why this sleep is needed? Is it doable to wait for the comment to be posted?

bsessions85: This is there, I believe, to keep spam triggers from being set off. If you post comments too quickly you'll get errors saying you need to slow down. This one has to be there.

bsessions85:
> should we still merge this though?

My initial though was no, but I guess it is one more we won't have to change when we switch all the tests to using Gutenberg. I am not sure about whether we should keep the old one at this point though.

specs/wp-comments-spec.js
juliaamosova: If the `sleep` statement is here to address the spam issue indeed, we'd need to keep it. I will, however, research if there is any way around that when get to work on this test.

bsessions85: I just ran this a bunch of times without the sleep and it actually doesn't appear to be a spam thing, it is just waiting for the comment to show up. This should be updated to wait for the comment to show up

JavonDavis: OK I'll merge with the sleep since @juliaamosova is already going through all the sleeps

JavonDavis:
> I am not sure about whether we should keep the old one at this point though

I was thinking the same 🤔🤔 I'm just going to remove the old one I'm not sure it's of much use at the moment and will be of less use in the coming weeks

specs/wp-comments-spec.js
juliaamosova: Sounds good to me! I'll implement the wait and check for the comment as suggested by @bsessions85 which should address the sleep statement here.

PR #1730, opened 2018-12-19
Update selectors for new onboarding flow

bsessions85:
This was broken by https://github.com/Automattic/wp-calypso/pull/29421

Just updating some selectors so that `[WPCOM] Sign Up  (desktop, en) Sign up for an account only (no site) then add a site via new onboarding flow @parallel` will pass again


shaunandrews:
Thanks for the catch and the fix. I'll pay more attention to the e2e tests now.

PR #1732, opened 2018-12-21
Update manage domains tests to run more reliably

bsessions85:
This one took some time to get figured out, but it came down to 3 different issues:

1. The first and second tests would occasionally conflict with each other when running in parallel. The clearing of the cart in one test would clear the cart in the other test and mess up the flow. I used a different test user in one of the tests to clear that up.

2. Sometimes clicking on the button to decline Google Apps would not do anything. I added a catch to retry it if things didn't progress.

3. The parallel tag was on the whole suite in addition to two of the three tests. This caused some funkiness in results and conflicting tests since the suite would run all three tests and the other two tests would also run.

### DIFF
--- a/lib/components/find-a-domain-component.js
+++ b/lib/components/find-a-domain-component.js
@@ -94,7 +94,7 @@ export default class FindADomainComponent extends AsyncBaseContainer {
 	}
 
 	async declineGoogleApps() {
-		this.driver.sleep(2000);
+		this.driver.sleep( 2000 );
 		return await driverHelper.clickWhenClickable(
 			this.driver,
 			this.declineGoogleAppsLinkSelector,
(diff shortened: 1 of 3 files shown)


lib/components/find-a-domain-component.js
```diff
@@ -94,11 +94,23 @@ export default class FindADomainComponent extends AsyncBaseContainer {
...
 			this.explicitWaitMS
 		);
+		try {
+			await driverHelper.waitTillNotPresent(
+				this.driver,
+				this.declineGoogleAppsLinkSelector
+			);
+		} catch ( err ) {
+			//Sometimes the first click doesn't work. Clicking again
+			await driverHelper.clickWhenClickable(
```
Stojdza: Are we sure that second click will always pass?

PR #584, opened 2017-06-21
Add spec to purchase Jetpack plan from wp-admin upgrade nudge

hoverduck:
Fixes #560 

@beaulebens - Can you let me know if this meets what you're expecting for this flow?

1) Log into both WordPress.com and wp-admin (via SSO) to a site on the Free plan
2) Open Jetpack dashboard
3) Click one of the Upgrade buttons (does it matter which one?  From your initial write-up in #560 I was expecting the links to go different places, but every Upgrade button points to `https://jetpack.com/redirect/?source=upgrade&site=SITEURL`, which redirects to `https://jetpack.com/pricing/settings-security-premium`)
4) Click the button at the top of that page to add a Premium plan to your cart and puts you on the payment page.

PR #1739, opened 2018-12-26
Remove and address `sleep` statement in `wp-admin-jetpack-page.js`

juliaamosova:
Removed and addressed `sleep` statement with a `for` loop that checks if the `Connect Jetpack` button ID matches with the currently active element ID (element in focus). 

I used the JavaScriptExecutor to get the focus on the button - as far as I know, there is no other way to do it using webdriver. JavaScriptExecutor is not ideal because it doesn't return anything (or rather returns `undefined`) but it does the job when the action needs to be made.

I also looked at doing the following:

1) Rewrite existing `waitTillFocused` function but I didn't want to use JavaScriptExecutor in it (for the reason described above) so I passed on that option;

2) Make a separate function out of the loop I wrote but the JavaScriptExecutor stopped me again. It gets the element using `getElementsByClassName` which can not always be applied to all elements. I passed on this option as well. 

After considering the 2 options above, I decided to leave the loop as one-time use. 

**To test:**

The change affects the following specs:

- `wp-jetpack-onboarding-spec.js`
- `wp-jetpack-connect-spec.js`
- `connect-disconnect-spec.js`
- `jetpack-connect-flow.js` which is a part of `gutenberg-markdown-block-spec.js`

### DIFF
--- a/lib/pages/wp-admin/wp-admin-jetpack-page.js
+++ b/lib/pages/wp-admin/wp-admin-jetpack-page.js
@@ -16,11 +16,11 @@ export default class WPAdminJetpackPage extends AsyncBaseContainer {
 
 	async connectWordPressCom() {
 		const selector = By.css( 'a.jp-jetpack-connect__button' );
-		const pauseBetweenFocusAttempts = 200;
+		const pauseBetweenFocusAttemptsMS = 200;
 		const connectJetpackButtonFocus =
 			"document.getElementsByClassName( 'jp-jetpack-connect__button')[0].focus();";
 		await driverHelper.waitTillPresentAndDisplayed( this.driver, selector );
-		for ( let i = 0; i < explicitWaitMS / pauseBetweenFocusAttempts; i++ ) {
+		for ( let i = 0; i < explicitWaitMS / pauseBetweenFocusAttemptsMS; i++ ) {
 			await this.driver.executeScript( connectJetpackButtonFocus );
 			let currentActiveElementId = await this.driver
 				.switchTo()
@@ -31,7 +31,7 @@ export default class WPAdminJetpackPage extends AsyncBaseContainer {
 			if ( connectJetpackButtonId === currentActiveElementId ) {
 				break;
 			}
-			await this.driver.sleep( pauseBetweenFocusAttempts );
+			await this.driver.sleep( pauseBetweenFocusAttemptsMS );
 		}
 		return await driverHelper.clickWhenClickable( this.driver, selector );
 	}
--- a/lib/pages/wp-admin/wp-admin-jetpack-page.js
+++ b/lib/pages/wp-admin/wp-admin-jetpack-page.js
@@ -4,6 +4,9 @@ import { By } from 'selenium-webdriver';
 
 import * as driverHelper from '../../driver-helper';
 import AsyncBaseContainer from '../../async-base-container';
+import config from 'config';
+
+const explicitWaitMS = config.get( 'explicitWaitMS' );
 
 export default class WPAdminJetpackPage extends AsyncBaseContainer {
 	constructor( driver ) {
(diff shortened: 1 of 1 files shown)


lib/pages/wp-admin/wp-admin-jetpack-page.js
```diff
@@ -13,8 +16,23 @@ export default class WPAdminJetpackPage extends AsyncBaseContainer {
 
 	async connectWordPressCom() {
 		const selector = By.css( 'a.jp-jetpack-connect__button' );
+		const pauseBetweenFocusAttemptsMS = 200;
+		const connectJetpackButtonFocus =
+			"document.getElementsByClassName( 'jp-jetpack-connect__button')[0].focus();";
```
brbrr: hm, I'm not quite sure what the problem you trying to fix here, but the connection button will be changed: https://github.com/Automattic/jetpack/issues/8618. 

Is the issue you trying to fix is somehow related to the fact that button not doing what it should right after page load? It might be a good idea to fix this on Jetpack side instead.

lib/pages/wp-admin/wp-admin-jetpack-page.js
```diff
@@ -13,8 +16,23 @@ export default class WPAdminJetpackPage extends AsyncBaseContainer {
...
 		await driverHelper.waitTillPresentAndDisplayed( this.driver, selector );
-		await this.driver.sleep( 1000 );
+		for ( let i = 0; i < explicitWaitMS / pauseBetweenFocusAttemptsMS; i++ ) {
+			await this.driver.executeScript( connectJetpackButtonFocus );
+			let currentActiveElementId = await this.driver
+				.switchTo()
+				.activeElement()
+				.getId();
+			let connectJetpackButtonId = await ( await this.driver.findElement( selector ) ).getId();
```
brbrr: are these params around `findElement` is needed? I think chained call with single `await` should do the job

lib/pages/wp-admin/wp-admin-jetpack-page.js
```diff
@@ -13,8 +16,23 @@ export default class WPAdminJetpackPage extends AsyncBaseContainer {
 
 	async connectWordPressCom() {
 		const selector = By.css( 'a.jp-jetpack-connect__button' );
+		const pauseBetweenFocusAttemptsMS = 200;
+		const connectJetpackButtonFocus =
+			"document.getElementsByClassName( 'jp-jetpack-connect__button')[0].focus();";
 		await driverHelper.waitTillPresentAndDisplayed( this.driver, selector );
-		await this.driver.sleep( 1000 );
+		for ( let i = 0; i < explicitWaitMS / pauseBetweenFocusAttemptsMS; i++ ) {
```
brbrr: Is this `for` loop might be replaced with `driver.wait` function? They seem like quite similar

lib/pages/wp-admin/wp-admin-jetpack-page.js
```diff
@@ -4,6 +4,9 @@ import { By } from 'selenium-webdriver';
 
 import * as driverHelper from '../../driver-helper';
 import AsyncBaseContainer from '../../async-base-container';
+import config from 'config';
+
+const explicitWaitMS = config.get( 'explicitWaitMS' );
```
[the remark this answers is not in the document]
juliaamosova: Noted, thanks @brbrr 👍

lib/pages/wp-admin/wp-admin-jetpack-page.js
juliaamosova: >Is the issue you trying to fix is somehow related to the fact that button not doing what it should right after page load? It might be a good idea to fix this on Jetpack side instead.

I am trying to remove and address the `sleep` statement. It was originally added here with no explanation why:

https://github.com/Automattic/wp-e2e-tests/pull/584/files#diff-e88aeb376f9b63bf70f5877bdece3966R14 

My guess is that the test runs too fast and clicks on the button when it is not ready to be clicked on yet. I was not able to reproduce it locally though. 

Since you mentioned the button will be changed, it makes sense to see if the issue will be present after the change takes affect. What do you think?

lib/pages/wp-admin/wp-admin-jetpack-page.js
juliaamosova: You are right, single `await` would do the job here. Thanks for catching it!

juliaamosova:
Closing this PR since `a.jp-jetpack-connect__button` will be removed soon as per here (after Jetpack 6.9 release): 

https://github.com/Automattic/wp-e2e-tests/blob/master/lib/pages/wp-admin/wp-admin-jetpack-page.js#L15 

The situatuation around `sleep` statement will be changed as well. Correcting it as-is now seems reduntant.

PR #1741, opened 2018-12-27
Jetpack tests: Multiple fixes

brbrr:
This PR fixes a bunch of small issues which causes jetpack e2e tests to fail.  

To test: Check CI job, maybe re-run it.

### DIFF
--- a/lib/components/secure-payment-component.js
+++ b/lib/components/secure-payment-component.js
@@ -1,6 +1,7 @@
 /** @format */
 
 import { By, promise } from 'selenium-webdriver';
+import config from 'config';
 
 import AsyncBaseContainer from '../async-base-container';
 import * as driverHelper from '../driver-helper.js';
@@ -8,7 +9,12 @@ import { currentScreenSize } from '../driver-manager';
 
 export default class SecurePaymentComponent extends AsyncBaseContainer {
 	constructor( driver ) {
-		super( driver, By.css( '.checkout__secure-payment-form,.secure-payment-form' ) );
+		super(
+			driver,
+			By.css( '.checkout__secure-payment-form,.secure-payment-form' ),
+			null,
+			2 * config.get( 'explicitWaitMS' )
+		);
 		this.paymentButtonSelector = By.css(
 			'.credit-card-payment-box button.is-primary:not([disabled])'
 		);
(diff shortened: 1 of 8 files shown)


lib/pages/wp-admin/wp-admin-jetpack-page.js
```diff
@@ -12,7 +12,10 @@ export default class WPAdminJetpackPage extends AsyncBaseContainer {
 	}
 
 	async connectWordPressCom() {
-		const selector = By.css( 'a.jp-jetpack-connect__button' );
+		// TODO: remove `a.jp-jetpack-connect__button,` after Jetpack 6.9 release
+		const selector = By.css(
+			"a.jp-jetpack-connect__button,.jp-connect-full__button-container a[href*='register']"
+		);
```
juliaamosova: @brbrr have you tried to fix the problem with the button not always being clickable here? Or was it something else? I wonder if I overcomplicated it here where I've tried to solve the button click issue and remove the `sleep` statement: https://github.com/Automattic/wp-e2e-tests/pull/1739. Can you please take a look when you get a chance? Thanks!

brbrr: No. Jetpack 6.9 will introduce new connection splash screen which introduces new connection button.

lib/components/secure-payment-component.js
```diff
@@ -1,14 +1,20 @@
...
 
 export default class SecurePaymentComponent extends AsyncBaseContainer {
 	constructor( driver ) {
-		super( driver, By.css( '.checkout__secure-payment-form,.secure-payment-form' ) );
+		super(
+			driver,
+			By.css( '.checkout__secure-payment-form,.secure-payment-form' ),
+			null,
+			2 * config.get( 'explicitWaitMS' )
+		);
```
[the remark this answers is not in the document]
brbrr: Yeah, I think I saw the same error. I wasn't quite sure what the root of the problem, but during visual monitoring, I thought that the problem is that some data weren't loaded in 20sec due to the fact that Jetpack(WPORG) site might not perform as fast as WPCOM sites.

lib/pages/wp-admin/wp-admin-jetpack-page.js
juliaamosova: Thanks for the clarification @brbrr 👍

lib/components/secure-payment-component.js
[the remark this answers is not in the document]
juliaamosova: Thanks for confirming @brbrr! I thought of the same reason originally but after testing with `waitTillNotPresent`, I could see that the test doesn't leave the plans page. I am still working on figuring out what could be going on there.

PR #1743, opened 2018-12-27
Remove and address `sleep` statement in `editor-page.js`

juliaamosova:
Removed and addressed `sleep` statement. See the explanation for the `sleep` statement removal in the comment below. 

In this PR, I also introduced a new driverHelper method `waitTillEnabled` - it finds the element and waits for it to be enabled utilizing `isEnabled()`. 

`isEnabled()` - tests whether this element is enabled, as dictated by the `disabled` attribute of the element.

`waitTillEnabled` can be applied to regular buttons. 

**To test:**

The changes affect `wp-media-upload-spec.js`.

### DIFF
--- a/lib/pages/editor-page.js
+++ b/lib/pages/editor-page.js
@@ -240,7 +240,9 @@ export default class EditorPage extends AsyncBaseContainer {
 	}
 
 	async deleteMedia() {
-		await driverHelper.clickWhenClickable( this.driver, by.css( '.editor-media-modal__delete' ) );
+		const deleteSelector = by.css( '.editor-media-modal__delete' );
+		await driverHelper.waitTillEnabled( this.driver, deleteSelector );
+		await driverHelper.clickWhenClickable( this.driver, deleteSelector );
 		return await driverHelper.clickWhenClickable(
 			this.driver,
 			by.css( 'button[data-e2e-button="accept"]' )
--- a/lib/pages/editor-page.js
+++ b/lib/pages/editor-page.js
@@ -68,7 +68,10 @@ export default class EditorPage extends AsyncBaseContainer {
 
 		await this.chooseInsertMediaOption();
 		await this.sendFile( newFile );
-		return await this.driver.sleep( 1000 );
+		return await driverHelper.waitTillEnabled(
+			this.driver,
+			by.css( 'button[data-e2e-button="confirm"]' )
+		);
 	}
 
 	async sendFile( file ) {
@@ -284,8 +287,10 @@ export default class EditorPage extends AsyncBaseContainer {
 	}
 
 	async errorDisplayed() {
-		await this.driver.sleep( 1000 );
-		return await driverHelper.isElementPresent( this.driver, by.css( '.notice.is-error' ) );
+		return await driverHelper.waitTillPresentAndDisplayed(
+			this.driver,
+			by.css( '.notice.is-error' )
+		);
 	}
 
 	async ensureContactFormDisplayedInPost() {
--- a/lib/pages/editor-page.js
+++ b/lib/pages/editor-page.js
@@ -241,7 +241,9 @@ export default class EditorPage extends AsyncBaseContainer {
 
 	async deleteMedia() {
 		const deleteSelector = by.css( '.editor-media-modal__delete' );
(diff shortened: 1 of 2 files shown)


lib/pages/editor-page.js
```diff
@@ -68,7 +68,10 @@ export default class EditorPage extends AsyncBaseContainer {
 
 		await this.chooseInsertMediaOption();
 		await this.sendFile( newFile );
-		return await this.driver.sleep( 1000 );
+		return await driverHelper.waitTillEnabled(
+			this.driver,
+			by.css( 'button[data-e2e-button="confirm"]' )
+		);
 	}
```
juliaamosova: `sendFile()` function ends when the media file is selected but it doesn't mean that it is ready to be used. When I was testing, I noticed that even if the media file is selected, the confirmation button gets enabled a bit later. New driverHelper method `waitTillEnabled` waits for the button to be enabled which provides the necessary wait time. As per `wp-media-upload-spec.js`, we don't need to click on the confirmation button but will be nice to wait until it is enabled prior to be clicked if and when needed.

juliaamosova:
>@juliaamosova I think you're on the right track with waitTillEnabled because waiting in this spec sometimes can be really long and driver.sleep() isn't reliable solution. Anyway, it looks like that waitTillEnabled doesn't return correct value. Here is what I caught:

Thanks for the thorough testing and review @Stojdza! 

Regarding the screenshot you sent: from the terminal, I can see that you are still on the `Can upload an image` step which calls `uploadMedia()` - this function ends when the `Insert` button is enabled. From the screenshot we can see that it is:


Image Link: https://cloudup.com/cAKsGC3RG9W

It looks like `isEnabled()` returns the correct value on `Can upload an image` step. 

The next step - `Can delete image` - calls for `deleteMedia()` which starts with waiting for the `trash can icon` being enabled via `waitTillEnabled`. 

In any case, let me work on it further and make it more stable. I will get back with a new commit.

Stojdza:
> I can see that you are still on the Can upload an image step which calls uploadMedia()

Step `Can upload an image` was successfully finished, this `isEnabled: true` is from `Can delete image` in [this](https://github.com/Automattic/wp-e2e-tests/blob/master/specs/wp-media-upload-spec.js#L51) step (waiting for _deleteSelector_ to be enabled):

```
async deleteMedia() {
      const deleteSelector = by.css( '.editor-media-modal__delete' );
      let status = await driverHelper.waitTillEnabled( this.driver, deleteSelector );
      console.log( '>>>isEnabled: ' + status );
```

juliaamosova:
>Step Can upload an image was successfully finished, this isEnabled: true is from Can delete image in this step (waiting for deleteSelector to be enabled):

@Stojdza thanks for the clarification! Sorry, I thought the status of `isEnabled()` in your screenshot was coming from a previous step. I see what you mean now. Let me see what can be done here.

juliaamosova:
@Stojdza I noticed that the `delete` button gets enabled after all but the trash bin icon on it gets displayed a tiny bit later so I added `waitTillPresentAndDisplayed` method for trash bin icon. It seems to improve the stability of the test but as I keep testing it locally I see that it still fails sometimes. 

I think the failure has something to do with the icon anyways. It is added via SVG Path. I also think that the `Close account` button has a similar issue: https://github.com/Automattic/wp-e2e-tests/pull/1721 -  it also has a trash bin icon on it and the click doesn't get registered there. 

I am still investigating.

Stojdza:
@juliaamosova What's the status of this one?

juliaamosova:
@Stojdza thanks for the ping! I am sorry it had been hanging here for that long. I've been wanting to get back working on it in January after my 2 weeks AFK time but did not get a chance. My current (support) team has a meetup this week. I won't be able to get to it before next week. On Monday, Feb 11th I am starting in a new role (EW) but will likely be consumed with the transition. I think we can close it for now. I will, however, get back to it via a new PR with reference to this one once settled on a new team. What do you think?

Stojdza:
@juliaamosova Sounds good 👍

juliaamosova:
Thanks @Stojdza! I am closing this PR for now then.

juliaamosova:
Closing this PR as per this comment: https://github.com/Automattic/wp-e2e-tests/pull/1743#issuecomment-460643820

PR #1744, opened 2018-12-27
Remove and address 2 sleep statements in `post-editor-sidebar-component.js`

juliaamosova:
In this PR, I removed and addressed 2 `sleep` statements:

- the first `sleep` is addressed by using `waitTillPresentAndDispplayed` which provides the necessary wait time to make sure the `Add new category` button is displayed;

- the second `sleep` is addressed by using `waitTillFocused` which provides the wait time to make sure the `Category name input` field is in focus which is needed prior to being able to enter the category name in the next step. 

**To test:**

The change affects `wp-calypso-gutenberg-post-editor-spec.js`.

### DIFF
--- a/lib/components/post-editor-sidebar-component.js
+++ b/lib/components/post-editor-sidebar-component.js
@@ -97,10 +97,10 @@ export default class PostEditorSidebarComponent extends AsyncBaseContainer {
 		const saveCategoryButtonSelector = By.css( 'div.dialog__action-buttons button.is-primary' );
 		const driver = this.driver;
 
-		driver.sleep( 500 );
+		await driverHelper.waitTillPresentAndDisplayed( driver, addNewCategoryButtonSelector );
 		await driverHelper.clickWhenClickable( driver, addNewCategoryButtonSelector );
 		await driverHelper.waitForFieldClearable( driver, categoryNameInputSelector );
-		driver.sleep( 500 );
+		await driverHelper.waitTillFocused( driver, categoryNameInputSelector );
 		await driverHelper.setWhenSettable( driver, categoryNameInputSelector, category );
 		await driverHelper.clickWhenClickable( driver, saveCategoryButtonSelector );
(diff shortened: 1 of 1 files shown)


juliaamosova:
>Did you know why sleep here was introduced?

@Stojdza yes, I looked it up when was working on it:

- sleep 1 - this one is tricky - `The test is clicking too fast and the modal isn't opening`: https://github.com/Automattic/wp-e2e-tests/commit/decb9e20b3dd7401cc81b32c89e254a24e750ef7 - the `sleep`, however, was only for half a second which `waitTillPresentAndDisplayed` provides.

- sleep 2 - was added the `Initial commit to master`: https://github.com/Automattic/wp-e2e-tests/commit/76cceca758de9b16ee276fef8395ae32bf055700 - there is no explanation for the `sleep` but my guess is that test was trying to enter the category name before the inout field was in focus and so the wait was added.

Stojdza:
🤦‍♂️ I oversaw this - `driver.sleep( 500 );` was missing `await`, so those sleeps didn't have any function. Have you tried to run with removed sleeps and without `waitTillPresentAndDispplayed ` and `waitTillFocused`?

juliaamosova:
>🤦‍♂️ I oversaw this - driver.sleep( 500 ); was missing await, so those sleeps didn't have any function.

@Stojdza I should have caught it as well so it's on me 😬nice catch!

>Have you tried to run with removed sleeps and without waitTillPresentAndDispplayed and waitTillFocused?

I did test it without `waitTillPresentAndDispplayed` and `waitTillFocused` when was working on it and just now again. It works well without them. But my understanding of the reason behind `sleep` statements is that they are not always needed but do help on some days when tests run too fast. This makes it difficult to replicate locally since the tests don't fail. 

I vote for keeping these 2 functions as they provide more determenistic (detailed) testing. But I will lean on your expertise if you think othervise.

Stojdza:
> I vote for keeping these 2 functions as they provide more determenistic (detailed) testing. But I will lean on your expertise if you think othervise.

+1

PR #1749, opened 2018-12-30
Remove and address 2 `sleep` statements in `wp-calypso-gutenberg-post-editor-spec.js`

juliaamosova:
In this PR I removed 2 `sleep` statements and addressed them by using new function `waitTillComponentsPanelIsDisplayed()`. 

Both `sleep` statements were introduced in the initial commit here: 

1) https://github.com/Automattic/wp-e2e-tests/pull/1652/files#diff-a831d02cf373c3e40058cda81fd99b67R85

2) https://github.com/Automattic/wp-e2e-tests/pull/1652/files#diff-a831d02cf373c3e40058cda81fd99b67R416

There is no history that shows failing tests without `sleep` statements. However, my guess is that they were added to make sure that components panel is ready to interact with when `Document` tab is selected on the sidebar. Therefore I wrote a new function `waitTillComponentsPanelIsDisplayed()` which waits for the components panel to be displayed. I replaced both `sleep` statements with the new function.

**To test:** 

The change affects `wp-calypso-gutenberg-post-editor-spec.js`.

### DIFF
--- a/lib/gutenberg/gutenberg-editor-sidebar-component.js
+++ b/lib/gutenberg/gutenberg-editor-sidebar-component.js
@@ -20,10 +20,7 @@ export default class GutenbergEditorSidebarComponent extends AsyncBaseContainer 
 		);
 	}
 	async selectDocumentTab() {
-		return await this.selectTab( 'Document' );
-	}
-
-	async waitTillComponentsPanelIsDisplayed() {
+		await this.selectTab( 'Document' );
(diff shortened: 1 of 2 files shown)


juliaamosova:
>Have you considered option to add same check in `selectDocumentTab()`? That way you could skip sleep and `waitTillComponentsPanelIsDisplayed()`.

Thanks for this @Stojdza! I like how you suggested refactoring the code as it fits the format of the rest of the test better. Let me push another commit.

PR #1750, opened 2019-01-02
Try to fix premium signup spec

alisterscott:
Use a scroll into view for the payment button to make sure the text can be extracted


alisterscott:
Merging this as it is failing on master

PR #1754, opened 2019-01-02
Fixing secure payment failures

bsessions85:
Two of the signup tests are failing fairly regularly with the error
`no such element: Unable to locate element: {"method":"css selector","selector":".credit-card-payment-box button.is-primary:not([disabled])"}`

This small change seems to fix that. I suspect the sleep removed in this file was masking it and we just needed to find the right selector to check for to ensure things are ready.

Tests that were failing:
- Sign Up (mobile, en) Sign up for a site on a personal paid plan coming in via /create as personal flow in GBP currency
- Sign Up (mobile, en) Sign up for a site on a premium paid plan coming in via /create as premium flow in JPY currency


bsessions85:
merging this in

PR #1755, opened 2019-01-02
Fix signup tests

JavonDavis:
Signup tests were failing locally for some varying reasons including 

1. Instability in `ensureNotLoggedIn` in the signup tests
2. Tests starting at step 4 (see screenshot below) instead of step 1

screen shot 2019-01-02 at 11 51 09 am

3. Timeout waiting for signup tests leading to 
```
TimeoutError: Timed out waiting for element with css selector of '.step-wrapper' to be present and displayed
Wait timed out after 20386ms
    at /Users/javon/Projects/a8c/wp-e2e-tests/node_modules/selenium-webdriver/lib/webdriver.js:834:17
```
4. Weird errors like 
screen shot 2019-01-07 at 4 14 16 pm

A few if not all of these I think are because of the first one on the list. 

** What I did ** 
* Included a small wait of 2s(which might be unnecessary but I was playing it safe) and an additional step to clear the cache through Chrome settings in the after step of each signup spec.
* I can't explain why but once the coupon test was in the signup spec the test below it failed, I spent a lot of time trying to figure out what's up but I decided it might be easier to just put both coupon specs in their own file for now. 

** What else I tried ** 
* Included waits and breakpoints in several steps to reduce failures(no luck)
* Updated ensureNotLoggedIn to include the additional step currently in the after step although this worked well in the signup tests it seemed to mess up a lot of stuff in other areas

------ 

I'm thinking we're not seeing these errors in CI thanks to the ability to retry failed tests a couple times before marking it as a failure but I'm hoping this can make running locally a little easier. I plan to spend a little more time looking at this further but thinking this might be good enough to merge now(?) then I'd like to see how it affects https://github.com/Automattic/wp-e2e-tests/pull/1761

Also, currently for me when running signup tests locally on master I get the following

screen shot 2019-01-07 at 4 29 46 pm

When running the same tests on this branch the tests failed aren't 0 but drop to 6.


lib/driver-helper.js
```diff
@@ -528,3 +528,23 @@ export async function dismissAlertIfPresent( driver ) {
 		return false;
 	}
 }
+
+export async function clear_cache( driver ) {
```
southp: I think this should be camelCased?

specs/wp-coupons-spec.js
```diff
@@ -0,0 +1,244 @@
...
+
+before( async function() {
+	this.timeout( startBrowserTimeoutMS );
+	driver = await driverManager.startBrowser();
+} );
+
+describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function() {
+	this.timeout( mochaTimeOut );
+
+	describe( 'Sign up for a non-blog site on a premium paid plan through main flow in USD currency using a coupon @parallel @visdiff', function() {
```
southp: FYI this test will fail when the GeoIP-mapped currency is not USD, because

1. The expected currency mismatched, as we discussed in https://github.com/Automattic/wp-e2e-tests/pull/1758
1. There are currencies that don't use decimal digits, so expecting the coupon-applied value to be exactly 0.99 times the original is not correct.

I have a candidate fix for the above two issues in https://github.com/Automattic/wp-e2e-tests/pull/1758/ .

lib/driver-helper.js
```diff
@@ -528,3 +528,23 @@ export async function dismissAlertIfPresent( driver ) {
...
+	let url = await driver.getCurrentUrl();
+	if ( url.startsWith( 'data:' ) === false && url !== 'about:blank' ) {
```
southp: When I printed the url here, most of the time the url wouldn't be either `data:` nor `about:blank`. Does that indicate some unexpected behavior?

lib/driver-helper.js
```diff
@@ -528,3 +528,23 @@ export async function dismissAlertIfPresent( driver ) {
...
+		return;
+	}
+	// Clear the cookies and cache for the ChromeDriver instance
+	// navigate to the settings page
+	await driver.get( 'chrome://settings/clearBrowserData' );
+	const clearCacheSelector = by.css( '* /deep/ #clearBrowsingDataConfirm' );
```
southp: There are serialized calypso state data in the local storage, so I'm wondering if we should make sure all the checkboxes has been checked before we let the webdriver click the button. e.g.
```
    const optionCheckboxesSelector = by.css( '* /deep/ #checkbox' );

    // wait for the checkboxes to appear
    await isEventuallyPresentAndDisplayed( driver, optionCheckboxesSelector );

    // toggle all the checkboxes
    await clickWhenClickable( driver, optionCheckboxesSelector );
```
before this line.

For example, in my case running the test `'Sign up for a domain only purchase coming in from wordpress.com/domains in EUR currency @parallel'` at line and then `'Sign up for a site on a business paid plan w/ domain name coming in via /create as business flow in CAD currency @parallel'` will make the later to fail. There seems to be some redux state left after running the former, and causes visiting `/start` being redirected to `/start/business/themes`.

Theoretically speaking `window.localStorage.clear()` should have cleaned all of them, but it doesn't seem that way.

southp:
I'm wondering if there is any case that a test is dependent on the other in the signup test suite. If no, maybe we can consider to initiate a new browser instance after each test instead of trying to figure out a way to clean everything up? Just a random thought :)

JavonDavis:
> ./node_modules/.bin/mocha specs/wp-signup-spec.js still failed in a seemingly random way on my local environment. Maybe it's expected as you've said it's still shaky?

Yeah but I'm hoping it was it fewer failures than on master?

Stojdza:
@JavonDavis What do you think to update `ensureNotLoggedIn` and put `clear_cache` inside? This way it looks like we are doing similar job twice - `ensureNotLoggedIn` at the beginning of the test and `clear_cache` at the end?

> I'm thinking we're not seeing these errors in CI thanks to the ability to retry failed tests a couple times before marking it as a failure but I'm hoping this can make running locally a little easier

We can disable retries [here](https://github.com/Automattic/wp-e2e-tests/blob/master/magellan.json#L7), and see what is happening?

lib/driver-helper.js
JavonDavis: Yeah it should be, I'll change this

lib/driver-helper.js
JavonDavis: Not necessarily, for example in the `dismissAllAlerts` function we use that page 
https://github.com/Automattic/wp-e2e-tests/blob/3654d16829e8fff9483756d14430cc69c5a9fd42/lib/driver-manager.js#L317

JavonDavis:
> I'm wondering if there is any case that a test is dependent on the other in the signup test suite. If no, maybe we can consider to initiate a new browser instance after each test instead of trying to figure out a way to clean everything up? Just a random thought :)

@southp I thought about it but I hadn't tried since this means starting and stopping a new session which I think would slow down the suite a lot, might be worth exploring though for stability sake

JavonDavis:
> @JavonDavis What do you think to update ensureNotLoggedIn and put clear_cache inside? This way it looks like we are doing similar job twice - ensureNotLoggedIn at the beginning of the test and clear_cache at the end?

@Stojdza I did try this, it's what I was referring to when I said 

```
Updated ensureNotLoggedIn to include the additional step currently in the after step although this worked well in the signup tests it seemed to mess up a lot of stuff in other areas
```

I'm thinking maybe I should have spent more time trying to resolve the side effects of the change but seeing as the signup tests seem to be only suite affected I was wondering if it was worth the effort? 

> We can disable retries here, and see what is happening?

Thanks for the suggestion! I'll give this a go

lib/driver-helper.js
JavonDavis: Hmm I've been giving this a try today and I seem to see a little more improvement, still working on it, look out for a few more commits as I work through it a bit more

> Theoretically speaking window.localStorage.clear() should have cleaned all of them, but it doesn't seem that way.

Yeah it's weird that something seems to be still left over, I'm wondering if somehow the test isn't waiting for the script to finish executing 🤷🏽‍♂️

southp:
Except for the automatic retrying, my assumption is that multiple chrome instances running by multiple worker processes also contribute here. From our observation so far, one of the main reasons that the signup test suite is failing is because the state leftover between tests. Thus running tests by different instances could probably mitigate the issue.

specs/wp-coupons-spec.js
JavonDavis: Merged #1758

JavonDavis:
screen shot 2019-01-14 at 4 39 21 pm

@Automattic/flowpatrol finally have all these tests working locally with the additional step for clearing the cache, could I get another review on this? Thanks!

lib/driver-helper.js
```diff
@@ -530,6 +530,26 @@ export async function dismissAlertIfPresent( driver ) {
...
+
+	// wait for the button to appear
+	await isEventuallyPresentAndDisplayed( driver, clearCacheSelector );
+
+	// click the button to clear the cache
+	await clickWhenClickable( driver, clearCacheSelector );
+
+	// wait for the button to be gone before returning
+	await driver.sleep( 10000 );
```
JavonDavis: It really takes a while to go away sometimes, not sure of a better way than to just wait here

bsessions85: driverHelper.waitTillNotPresent()?

JavonDavis: Thanks! It worked

JavonDavis: @bsessions85 I think this could've actually been a great use casse for Github's new code suggestion feature, drawback is that it bypasses the linter

bsessions85:
Would it be better to just start a new browser for each of these tests? That would give you a fresh session and we wouldn't have to depend on trying to open a settings page, click stuff and make sure it goes away. It would also keep it from having to add so many sleeps.

Stojdza:
I forgot to say earlier, great analysis @JavonDavis! This issue really bothers me when I run sign up spec locally. 

I've tested these changes, and this way is much more stable than any other. But I also had some failures with it, including this one: `TimeoutError: Timed out waiting for element with css selector of '* /deep/ #clearBrowsingDataConfirm' to be clickable`

Since clearCache in after block extends execution time and there are still some failures, my main concern is should we keep these changes or try _new browser for each of these tests_?

bsessions85:
@JavonDavis @Stojdza  I was curious about using a new browser for each test, so I gave it a shot. I tried it and still got a few failures, but it was much better than before and without relying on the setting page.

To do it I just called a new function that restarts the browser (quitBrowser & startBrowser) in the before of each test. The only problem I had to work around was the chromedriver service still running. I put in a check to see if it is running and it all worked fine. I can give you code or share a branch if you're interested.

JavonDavis:
> To do it I just called a new function that restarts the browser (quitBrowser & startBrowser) in the before of each test. The only problem I had to work around was the chromedriver service still running. I put in a check to see if it is running and it all worked fine. I can give you code or share a branch if you're interested.

I was currently just working through this! Can you share the workaround please and thanks! if there's a branch you can just point me there otherwise the code here works

JavonDavis:
> I've tested these changes, and this way is much more stable than any other. But I also had some failures with it, including this one: TimeoutError: Timed out waiting for element with css selector of '* /deep/ #clearBrowsingDataConfirm' to be clickable

Weird I ran this multiple times and never got this but not since I removed the sleeps... @Stojdza mind giving it a go on this commit 3fb1312?

bsessions85:
@JavonDavis  I put in a PR with the changes, just so they are easier to spot. driver-manager is the file to look at. #1778

Stojdza:
> Weird I ran this multiple times and never got this but not since I removed the sleeps... @Stojdza mind giving it a go on this commit 3fb1312?
Yeah, I got that error when I ran tests with this commit included.

JavonDavis:
This is handled in #1780

PR #1758, opened 2019-01-03
Set the expected currency

southp:
## Summary

This PR adds the expected currency through `setCurrencyForPayments()` for the `Sign up for a non-blog site on a premium paid plan through main flow in USD currency using a coupon` test, so that the test result won't be affected due to geo IP.

## How to reproduce

1. Use a VPN to change your IP address to somewhere non-US, e.g. Japan.
1. Run the test, and it will fail at the currency matching step, since the currency will be in JPY.

## Test plan

Run the test from this branch, and the currency matching step should pass as expected.

### DIFF
--- a/specs/wp-calypso-gutenberg-post-editor-spec.js
+++ b/specs/wp-calypso-gutenberg-post-editor-spec.js
@@ -364,16 +364,15 @@ describe( `[${ host }] Calypso Gutenberg Editor: Posts (${ screenSize })`, funct
 				await gSidebarComponent.trashPost();
 			} );
 
-			// Not working https://github.com/Automattic/wp-calypso/issues/28813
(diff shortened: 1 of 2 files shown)


JavonDavis:
@southp This is going to fail because the coupon won't work in the sanbox environment, that's why this test also doesn't complete the purchase and simply checks if the coupon was correctly applied or not. Now that you've pointed out the currency issue the name becomes misleading, it might be simpler to remove the USD currency from the name of the test since we already have a test for USD currency that sets it in the sandbox and this one is mainly for coupon validation.

southp:
@JavonDavis Ah, thanks for the explanation. I have updated the test name and the content so that it should pass regardless of GeoIP derived currency. I'd be grateful if you could have a revisit when you have some time 🙇

JavonDavis:
Merged in some fixes from master

PR #1759, opened 2019-01-03
Update gdpr cookie to account for live branch tests

bsessions85:
Thought of this idea over the weekend so I decided to give it a shot.
It is to keep the cookie banner from showing when running against livebranches. It should also work against wordpress.com


bsessions85:
Merging in to get things running again

alisterscott:
Thanks for persevering with this one @bsessions85

PR #1761, opened 2019-01-04
Replace the checklist with the site preview as the first thing after signing up: take 2

southp:
## Summary

This PR attempts to update all the tests affected by the change introduced by Automattic/wp-calypso#29031 , which uses the site preview as the first view after signing up instead of the checklist.

### DIFF
--- a/specs/wp-manage-domains-spec.js
+++ b/specs/wp-manage-domains-spec.js
@@ -201,7 +201,7 @@ describe( `[${ host }] Managing Domains: (${ screenSize })`, function() {
 		} );
 	} );
 
-	describe.only( 'Transfer a domain to an existing site (partial) @parallel', function() {
+	describe( 'Transfer a domain to an existing site (partial) @parallel', function() {
 		const domain = 'automattic.com';
 
 		before( async function() {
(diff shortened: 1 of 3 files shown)


specs/wp-manage-domains-spec.js
```diff
@@ -201,7 +201,7 @@ describe( `[${ host }] Managing Domains: (${ screenSize })`, function() {
 		} );
 	} );
 
-	describe( 'Transfer a domain to an existing site (partial) @parallel', function() {
+	describe.only( 'Transfer a domain to an existing site (partial) @parallel', function() {
```
Stojdza: @southp Was this `.only` left intentionally? I've almost missed it, but I saw that test run was only [~1min long](https://circleci.com/gh/Automattic/wp-e2e-tests/25826)  :)

southp: Argh! Thanks for catching it. I've removed it in d3aa3ba.

Stojdza:
@southp Feel free to merge it right after https://github.com/Automattic/wp-calypso/pull/29031

PR #1762, opened 2019-01-04
Update login-flow to account for development environments

gwwar:
I was running some tests locally and noticed some failures for local runs, because we visit devdocs instead of the reader after login. I added a check there to account for the case, but let me know if there's something better to compare against.

### Testing Instructions

Try to run the wp-invite-users-spec.js against a local calypso instance, and the config using `"calypsoBaseURL": "http://calypso.localhost:3000",`


bsessions85:
Failed test is unrelated.

bsessions85:
Merging so it doesn't get stale

PR #1763, opened 2019-01-07
Fix/few more jetpack failures

brbrr:
Fixes two failing Jetpack tests. Also, I noticed some missing `await` calls, so I added them. Also, I updated the Markdown block test to use default JN template instead of gutenpack, since it is not needed anymore.

To test locally:
`BROWSERSIZE=mobile JETPACKHOST=PRESSABLE npx mocha specs/wp-post-editor-spec.js -g 'Edit a Post:'`
`JETPACKHOST=PRESSABLE npx mocha specs-jetpack-calypso/wp-jetpack-connect-spec.js -g 'Pre-connect from Jetpack.com using "Install Jetpack" button:'`
`JETPACKHOST=PRESSABLE npx mocha specs/wp-comments-spec.js -g 'newly created post:'`

### DIFF
--- a/specs/wp-comments-spec.js
+++ b/specs/wp-comments-spec.js
@@ -60,6 +60,8 @@ describe( `[${ host }] Comments: (${ screenSize })`, function() {
 		} );
 
 		step( 'Can post a reply', async function() {
+			// NOTE: we need to wait to prevent "You are posting comments too quickly. Slow down." error
+			await driver.sleep( 10000 );
 			const commentArea = await CommentsAreaComponent.Expect( driver );
 			await commentArea.reply(
 				{
--- a/lib/components/post-editor-sidebar-component.js
+++ b/lib/components/post-editor-sidebar-component.js
@@ -129,7 +129,7 @@ export default class PostEditorSidebarComponent extends AsyncBaseContainer {
 	async setCommentsForPost( allow = true ) {
 		let driver = this.driver;
 		const selector = By.css( 'input[name=comment_status]' );
-		driverHelper.waitTillPresentAndDisplayed( driver, selector );
+		await driverHelper.waitTillPresentAndDisplayed( driver, selector );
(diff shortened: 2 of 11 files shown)


brbrr:
@alisterscott `[Status] Needs Jetpack e2e Testing` is not needed here, since  `ci/circleci` is already ran against PRESSABLE target

specs/wp-comments-spec.js
```diff
@@ -60,6 +60,8 @@ describe( `[${ host }] Comments: (${ screenSize })`, function() {
 		} );
 
 		step( 'Can post a reply', async function() {
+			// NOTE: we need to wait to prevent "You are posting comments too quickly. Slow down." error
+			await driver.sleep( 10000 );
```
Stojdza: In which case this error happens and is 10s enough? Does `waitTillPresentAndDisplayed` can help instead of sleep?

brbrr: There is some sort of spam blocker which is rejecting comment submission it was right after the first one. Maybe there is a way to hack it, but since that delay is static - static timeout should do the job

alisterscott:
> @alisterscott [Status] Needs Jetpack e2e Testing is not needed here since ci/circleci already runs against PRESSABLE target. ci/wp-e2e-tests-full-jetpack failing because of the site used in these tests are down.

Sorry I forgot we still did that. So we don't run Jetpack specific changes against WPCOM in that circumstance?

brbrr:
Yeah, which is far from ideal. Also, I don't really like the current implicit behavior. Can we switch `[Status] Needs Jetpack e2e` target to`PRESSABLE`?

bsessions85:
@brbrr  I can update the bridge to use `PRESSABLE`. Do you want it changed for branches on wp-e2e-tests and wp-calypso?

brbrr:
We chatted about that in slack

PR #1764, opened 2019-01-08
Improve cleanup of business sign up tests

bsessions85:
I was noticing lots of slack messages about cleanup failing. This  extra check clears them up in my test runs. 

To Test: 
Ensure cleanup completes without issues on `'Sign up for a site on a business paid plan w/ domain name coming in via /create as business flow in CAD currency @parallel'`


bsessions85:
Merging simple change to quiet  slack a bit

PR #1767, opened 2019-01-08
Reduce number of screenshot failure slack messages

bsessions85:
We were getting double messages when a screenshot didn't work for some reason. This supresses the second one..


bsessions85:
Merging simple change to quiet slack

PR #1768, opened 2019-01-09
Updating `dismissAllAlerts` to use async/await style

JavonDavis:
Small update to `dismissAllAlerts` to use async/await.

### DIFF
--- a/lib/mocha-hooks.js
+++ b/lib/mocha-hooks.js
@@ -95,7 +95,7 @@ afterEach( async function() {
 
 // Dismiss any alerts for switching away
 afterEach( async function() {
-	this.timeout( afterHookTimeoutMS );
+	await this.timeout( afterHookTimeoutMS );
 	const driver = global.__BROWSER__;
 
 	if (
@@ -103,7 +103,7 @@ afterEach( async function() {
 		this.currentTest.state === 'failed' &&
 		( config.get( 'closeBrowserOnComplete' ) === true || global.isHeadless === true )
 	) {
-		await driverManager.dismissAllAlerts( driver );
+		await driverManager.acceptAllAlerts( driver );
 	}
 } );
 
@@ -132,7 +132,7 @@ afterEach( async function() {
 
 // Push SauceLabs job status update (if applicable)
 after( async function() {
-	this.timeout( afterHookTimeoutMS );
+	await this.timeout( afterHookTimeoutMS );
 	const driver = global.__BROWSER__;
 
 	if ( config.has( 'sauce' ) && config.get( 'sauce' ) ) {
--- a/lib/mocha-hooks.js
+++ b/lib/mocha-hooks.js
@@ -86,15 +86,15 @@ afterEach( async function() {
 			} );
 		},
 		err => {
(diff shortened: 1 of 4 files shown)


lib/mocha-hooks.js
```diff
@@ -86,15 +86,15 @@ afterEach( async function() {
 			} );
 		},
 		err => {
-			slackNotifier.warn( `Could not take screenshot due to error: '${ err }'`,
-				{ suppressDuplicateMessages: true }
-			);
+			slackNotifier.warn( `Could not take screenshot due to error: '${ err }'`, {
+				suppressDuplicateMessages: true,
```
bsessions85: This needs to be in the brackets

lib/mocha-hooks.js
```diff
@@ -86,15 +86,15 @@ afterEach( async function() {
...
+			} );
 		}
 	);
 } );
 
 // Dismiss any alerts for switching away
-afterEach( function() {
+afterEach( async function() {
 	this.timeout( afterHookTimeoutMS );
```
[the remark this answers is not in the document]
JavonDavis: Yep that and another one below I'll update

lib/mocha-hooks.js
JavonDavis: weird, I don't remember updating this but I'll make the change, thanks

JavonDavis:
> I tried running a gutenberg editor spec when adding a fail statement within a single spec in the file, and it didn't work as the alert remained open
> 
> I then tried the same thing on master and it also didn't work 🤔
> 
> I was thinking we could try an approach I've used previously where I've run JavaScript to override the beforeunload event in the browser to not show the alert, however it didn't seem to work in Calypso 🤔
> 
> What do you think about such an approach? In ruby it looked something like:
> 
> browser.execute_script('window.onbeforeunload = null')

@alisterscott this gave me a pretty hard time today I tried different approaches, including the one you suggested with no luck... I then realised in this case we're dismissing the alert when to actually leave the page we'd need to accept the alert, for this to work I had to replace `dismissAllAlerts` with `acceptAllAlerts` and accept the alert if present in the after step of the signup specs. I'm not sure if there'll be any adverse effects to the change from dismiss to alert in `mocha-hooks` but I'll check that out in CI.

lib/mocha-hooks.js
JavonDavis: @bsessions85 it is in brackets, the linter just formatted it differently

alisterscott:
>  I then realised in this case we're dismissing the alert when to actually leave the page we'd need to accept the alert, 

It drove me crazy yesterday too. And oh wow, I didn't even think of that. Good work figuring that out.

JavonDavis:
`wp-calypso-gutenberg-post-editor-spec.js` seems to be passing in CI but is failing locally for me, @Automattic/flowpatrol anyone else experiencing this? In any case I think it's unrelated to this change and can be handled in a separate PR, thoughts on merging this?

Stojdza:
@JavonDavis It fails for me locally as well. I suppose we are good to go with these changes 👍

PR #1772, opened 2019-01-11
Refactor: extract the common code for checking if the site preview is present.

southp:
## Summary

This PR attempts to introduce a pattern of reusing the same testing step code across tests to reduce code duplication. In sum, it mainly includes three ingredients:

1. A singleton method to share the browser instance across code modules.
1. A `{ name, body }` object structure to encapsulate test step code.
1. A `addStep` helper method which generates a testing step using the above data structure.

With those, it then uses `Can then see the site preview` step as the demonstrating scenario.


taggon:
Interesting idea! Writing the same test repeatedly is a kind of pain point. This PR would be a good start to get rid of that. I'm going to add my thought here.

The `mocha` framework already has a guide to shared behaviors:
https://github.com/mochajs/mocha/wiki/Shared-Behaviours

> The "context" (this) is the same object within the "before each", "after each" hooks, and the test-case itself, allowing you to utilize this instead of closures to store data.

So you don't need to create a singleton browser instance. Just another function call would be good. Here's my rough sketch of the implementation.

```javascript
// wp-signup-specs.js

...

before( async function() {
  this.timeout( startBrowserTimeoutMS );
  this.driver = driver = await driverManager.startBrowser();
} );

...

// Shared steps. You can move the following block into a separate file.
const shared = {
  shouldSeeSitePreview() {
    step( 'Can then see the site preview', async function() {
      ...
    } );
  }
};

// Then call it within a test suite.
...
    share.shouldSeeSitePreview();

    step( 'Can log out and request a magic link', async function() {
      ...
    } );
...

// In case you'd prefer to put shared steps into another file.
// shared/wp-signup-specs.js
export function shouldSeeSitePreview() {
 ...
}
```

Stojdza:
Works and looks great! Thanks for this initiative @southp :) 

@bsessions85 Would you mind also to take a look at this one?

bsessions85:
I like it. Thanks for taking it on! I think if we can make it work like @taggon mentioned, it will be great

southp:
Thanks for the great tip, @taggon . I have made use of the test context mechanism to simplify the code as you described. Please have a revisit when you have some time 🙇

PR #1774, opened 2019-01-11
Fix: Update the site info component to match the calypso-side changes

southp:
## Summary
Update the site info component to match the changes from the calypso side, corresponding to https://github.com/Automattic/wp-calypso/pull/30114.


southp:
Also, added two ab-test keys in 38a0bcd and f5ab75d to fix the intermittently failing tests.

PR #1778, opened 2019-01-15
Running in new browser- Signup test

bsessions85:
@JavonDavis  Here are the changes I mentioned.  This PR is just so they are easy to see against your branch


JavonDavis:
Thanks!

Stojdza:
Ran this locally several times, and it passed every time 👌

JavonDavis:
Now handled in #1780

PR #1780, opened 2019-01-16
Fix signup spec

Stojdza:
We had some problems executing `wp-signup-spec.js` locally - if one test fails, it will affect every next test in line and they will fail also. After several different tries and approaches, I concluded that some tests were missing account deletion at the end, newly registered user is not signed out and that makes a problem for flow in the next test. 

This PR ensures that every new account is deleted, and instead of `step`, `deleteAccount()` is in `after` block. That way, if test fails account will be deleted. 

To test:
Run `wp-signup-spec.js` and make sure it passes. If some test fails, next test shouldn't fail.

### DIFF
--- a/lib/flows/delete-account-flow.js
+++ b/lib/flows/delete-account-flow.js
@@ -3,9 +3,7 @@
 import CloseAccountPage from '../pages/account/close-account-page';
 import LoggedOutMasterbarComponent from '../components/logged-out-masterbar-component';
 import * as SlackNotifier from '../slack-notifier';
-import NavBarComponent from '../components/nav-bar-component';
 import AccountSettingsPage from '../pages/account/account-settings-page';
-import ProfilePage from '../pages/profile-page';
 
 export default class DeleteAccountFlow {
 	constructor( driver ) {
@@ -13,11 +11,11 @@ export default class DeleteAccountFlow {
 	}
 	async deleteAccount( name ) {
 		return ( async () => {
-			const navBarComponent = await NavBarComponent.Expect( this.driver );
-			await navBarComponent.clickProfileLink();
-			const profilePage = await ProfilePage.Expect( this.driver );
-			await profilePage.chooseAccountSettings();
-			const accountSettingsPage = await AccountSettingsPage.Expect( this.driver );
+			let url = await this.driver.getCurrentUrl();
+			if ( url.startsWith( 'data:' ) === true ) {
+				await this.driver.sleep( 2000 );
+			}
+			const accountSettingsPage = await AccountSettingsPage.Visit( this.driver );
 			await accountSettingsPage.chooseCloseYourAccount();
 			const closeAccountPage = await CloseAccountPage.Expect( this.driver );
 			await closeAccountPage.chooseCloseAccount();
--- a/lib/flows/delete-account-flow.js
+++ b/lib/flows/delete-account-flow.js
@@ -3,6 +3,7 @@
 import CloseAccountPage from '../pages/account/close-account-page';
 import LoggedOutMasterbarComponent from '../components/logged-out-masterbar-component';
 import AccountSettingsPage from '../pages/account/account-settings-page';
+import * as SlackNotifier from '../slack-notifier';
 
 export default class DeleteAccountFlow {
 	constructor( driver ) {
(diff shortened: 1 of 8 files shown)


specs/wp-signup-spec.js
```diff
@@ -82,6 +82,7 @@ describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function() {
 		const expectedBlogAddresses = dataHelper.getExpectedFreeAddresses( blogName );
 		const emailAddress = dataHelper.getEmailAddress( blogName, signupInboxId );
 		let magicLoginLink;
+		let isLoggedIn = false;
```
Stojdza: `isLoggedIn` is a flag which is true if the user is logged in.

JavonDavis:
@Stojdza were you able to get a clean run? I ran it a couple times and about 7 or 8 specs fail

Stojdza:
@JavonDavis yes, several times. And when it fails, it's only one test. I'll take a look again.

lib/flows/delete-account-flow.js
```diff
@@ -3,21 +3,19 @@
...
-			const navBarComponent = await NavBarComponent.Expect( this.driver );
-			await navBarComponent.clickProfileLink();
-			const profilePage = await ProfilePage.Expect( this.driver );
-			await profilePage.chooseAccountSettings();
-			const accountSettingsPage = await AccountSettingsPage.Expect( this.driver );
+			let url = await this.driver.getCurrentUrl();
+			if ( url.startsWith( 'data:' ) === true ) {
+				await this.driver.sleep( 2000 );
+			}
+			const accountSettingsPage = await AccountSettingsPage.Visit( this.driver );
```
Stojdza: Start deleting an account from AccountSettingsPage() instead of navigating to it.

lib/flows/delete-account-flow.js
```diff
@@ -3,21 +3,19 @@
...
 	async deleteAccount( name ) {
 		return ( async () => {
-			const navBarComponent = await NavBarComponent.Expect( this.driver );
-			await navBarComponent.clickProfileLink();
-			const profilePage = await ProfilePage.Expect( this.driver );
-			await profilePage.chooseAccountSettings();
-			const accountSettingsPage = await AccountSettingsPage.Expect( this.driver );
+			let url = await this.driver.getCurrentUrl();
+			if ( url.startsWith( 'data:' ) === true ) {
+				await this.driver.sleep( 2000 );
```
Stojdza: Wait 2s if test brakes and after that URL is `data:`. I didn't find another way to wait and open AccountSettingsPage().

JavonDavis:
@Stojdza the domain only purchase spec is causing problems locally for me, it consistently fails on the step to click the cancel button and subsequently throws the following error. I've been trying to see what's up with this but not seeing why the cancel button seems to be clicked but actually stops the canceling process. I'm quickly testing this by just running `Sign up for a domain only purchase coming in from wordpress.com/domains in EUR currency @parallel` and `Sign up for a site on a business paid plan w/ domain name coming in via /create as business flow in CAD currency @parallel`

`TimeoutError: Timed out waiting for element with css selector of '.notice.is-success.is-dismissable' to be present and displayed`

Whenever that spec fails it seems to affect the spec after it which I think then causes failures in the next 2 specs. The specs after that pass.

specs/wp-signup-spec.js
```diff
@@ -175,8 +175,15 @@ describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function() {
 			return await ReaderPage.Expect( driver );
 		} );
 
-		step( 'Can delete our newly created account', async function() {
-			return await new DeleteAccountFlow( driver ).deleteAccount( blogName );
+		after( 'Can delete our newly created account', async function() {
```
bsessions85: Can these be simplified by putting the catch in deleteAccount?

bsessions85: I just noticed that taking it out of deleteAccount was part of the changes. Is there a specific reason for that?

bsessions85:
@JavonDavis  I ran both desktop and mobile and only had one failure

specs/wp-signup-spec.js
Stojdza: @bsessions85 Basically, it's because of cases like [this](https://github.com/Automattic/wp-e2e-tests/pull/1780/files#diff-9bb2a6ea7acb62d90467ce1cf8ce18e6R933), when we put several different actions in `after` block. If `deleteAccount` also contains a similar message, we will have two messages if it fails. On the other hand, yes, it would simpler to put it in `deleteAccount`, and since we have only one `after` with more than `deleteAccount` I'll revert this change.

Stojdza:
Thanks @JavonDavis! Maybe I should have done it too before I started 😅

PR #1785, opened 2019-01-28
Media editor spec

### LINKED ISSUE #1769: Request: Test media editor flow
A bug was recently introduced that passed the full e2e suite.

Please add coverage for the media editor flow as described in https://github.com/Automattic/wp-calypso/issues/30053:

> This is a recently created issue, likely from a recent PR and is affecting live users, including paying users who cannot change their site icon.
> 
> 1. In Media, choose an image
> 2. Choose Edit
> 3. Edit Image
> 4. `Sorry, there was a problem loading the image. Please close this editor and try selecting the image again.`
> 
> You can get to the same error by uploading a site icon in My Site > Settings
> 
> #### What I expected
> I would expect the editor to load, or the site icon to upload.
> 
> #### What happened instead
> This error persists, no matter the route or image.
> 
> #### Screenshot / Video
> screen shot 2019-01-09 at 9 51 39 am

JavonDavis:
Add spec for editing image from media in sidebar. 

fixes #1769


lib/pages/media-page.js
```diff
@@ -0,0 +1,50 @@
...
+import * as driverManager from '../driver-manager';
+
+export default class MediaPage extends AsyncBaseContainer {
+	constructor( driver ) {
+		super( driver, By.css( '.media-library__content' ) );
+	}
+
+	async selectRandomMedia() {
+		await this.driver.sleep( 3000 ); // TODO: try and remove this sleep
+		await driverHelper.clickWhenClickable( this.driver, By.css( '.media-library__list-item' ) );
```
alisterscott: Is this selecting a random media? Or always the first media item?

specs/wp-media-editor-spec.js
```diff
@@ -0,0 +1,68 @@
...
+	driver = await driverManager.startBrowser();
+} );
+
+describe( `[${ host }] Media: Edit Media (${ screenSize }) @parallel @jetpack`, function() {
+	this.timeout( mochaTimeOut );
+
+	describe( 'Edit Existing Media:', function() {
+		before( async function() {
+			const loginFlow = new LoginFlow( driver );
+			await loginFlow.login( { useFreshLogin: true } );
```
Stojdza: Is there any specific reason to use `useFreshLogin`? I tried without it and spec works well.

lib/pages/media-page.js
Stojdza: Yeah, it will always select the first image. 

@JavonDavis We already have [a function](https://github.com/Automattic/wp-e2e-tests/blob/master/lib/pages/editor-page.js#L139) which does the same, without sleep. Can you re-use it?

specs/wp-media-editor-spec.js
JavonDavis: No specific reason, overlooked this in a copy paste, I'll take it out

JavonDavis:
> @JavonDavis We already have a function which does the same, without sleep. Can you re-use it?

@Stojdza I was looking to see if we had something like this already, I missed this! It's no problem just bringing in the page object and using the function but in terms of reading the code, one might wonder why the editor page is here. What do you think is the best approach? 

* Just use the Editor page? ( Which I'm perfectly fine with I'm just thinking it might be a bit misleading when reading this spec)
* Move the code out of the Editor page and place into a common component used by both pages? (Seems like a bit too much) 
* Use the media Page in the Editor Page(🤔this might make more sense, I think)
* Some other option?

Stojdza:
> one might wonder why the editor page is here 

Totally agree with this. Since the current solution was the fastest solution and it works well I'm also fine with it. 

> Use the media Page in the Editor Page(🤔this might make more sense, I think)

Yeah, this sounds like a good solution, at the end it's a media page, right? If it takes too much time, let's go with the current solution.

sirreal:
Thanks!

PR #1790, opened 2019-01-31
Bump date on improved onboarding test

adamziel:
We're changing the date in the following PR: https://github.com/Automattic/wp-calypso/pull/30512


adamziel:
Since the only failure is the same as on master branch, I'm merging this one

PR #1794, opened 2019-02-01
Update slack hooks

Stojdza:
New slack hooks are added. We should see tests execution status in related slack channels.


bsessions85:
@Stojdza Did you by chance use the old CONFIG_KEY to encrypt this?

Stojdza:
@bsessions85 That shouldn't be the case, but I'll do whole process again

PR #1806, opened 2019-02-21
Domains: Add .home.blog as one of the expected addresses

aidvu:
We also return .blog subdomains, so look for that too


aidvu:
Fixes https://github.com/Automattic/wp-calypso/pull/30914 e2e failure.

PR #1808, opened 2019-02-21
Reduce timeout for signup processing to 150s

Stojdza:


specs/wp-signup-spec.js
```diff
@@ -1719,7 +1719,7 @@ describe( `[${ host }] Sign Up  (${ screenSize }, ${ locale })`, function() {
 		} );
 	} );
 
-	describe( 'Sign up for an account only (no site) then add a site via new onboarding flow @parallel', () => {
+	describe.only( 'Sign up for an account only (no site) then add a site via new onboarding flow @parallel', () => {
```
bsessions85: missed an only

Stojdza:
🤦‍♂️ Thanks @bsessions85!

PR #1815, opened 2019-02-27
Temporarily disable signup from /domains test

Stojdza:
Temporarily disable test while we don't find out what is causing [failures](https://circleci.com/gh/Automattic/wp-e2e-tests/26674). 

This is what I caught while running locally:
screen shot 2019-02-27 at 12 29 37


Stojdza:
Thanks @bsessions85!

PR #1817, opened 2019-03-01
Fix: Update expiry date for credit card

niranjan-uma-shankar:
Currently tests fail whenever a credit card info is expected, since the expiry date for the card is set to 02/19. 

screen shot 2019-03-01 at 16 57 47

This PR fixes the issue by updating the expiry date.


Stojdza:
Thanks for this @niranjan-uma-shankar!

PR #1818, opened 2019-03-01
Gutenframe - updated tests

Stojdza:
Updated tests for changes related to Block editor in iframe (Gutenframe). 

We should keep this PR up to date and merge it when Gutenframe lands on production. 

cc @gwwar @mmtr


bsessions85:
This looks good. We'll probably want to turn on the feature in production, run the tests, and then merge after they pass so we can see the tests pass

Stojdza:
Thanks @mmtr!

PR #1819, opened 2019-03-04
Change i18n Slack channel for screenshots

akirk:
I've updated the encrypted config file, can someone help double check that I only modified the config variable needed? Thanks!


akirk:
Thanks!

PR #1822, opened 2019-03-06
Add new AB test in default.json

Stojdza:
Found this one while testing locally.


Stojdza:
@bsessions85 Yeah, I should definitely get used to adding changes to wp-calypso :) 

Here is the PR https://github.com/Automattic/wp-calypso/pull/31257

PR #1827, opened 2019-03-13
Change jetpack monthly a/b test name

johnHackworth:
An A/B test changed its name and date, and these changes should fix e2e tests


jsnajdr:
Shouldn't this be in the Calypso repo, as the e2e tests were moved there a few days ago?

Stojdza:
Correct! @johnHackworth Now you can update A/B tests [here](https://github.com/Automattic/wp-calypso/blob/master/test/e2e/config/default.json#L62)

config/default.json
```diff
@@ -63,7 +63,7 @@
 		"skipThemesSelectionModal",
 		"springSale30PercentOff",
 		"jetpackSignupGoogleTop",
-		"onlyJetpackMonthly",
+		"jetpackMontlyPlansOnly",
```
tyxla: Noting here as well that we have a typo here.

johnHackworth:
d'oh! :man_facepalming: 

I missed that... thank you folks, I'll fix it :+1:

PR #1829, opened 2019-04-11
Add entry for new AB test: conciergeQuickstartSession

victorespigares:
We'll be running a new Calypso test (`conciergeQuickstartSession`) with different copy for the Concierge Upsell session.

Calypso PR: https://github.com/Automattic/wp-calypso/pull/32176

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -75,6 +75,7 @@
 		"builderReferralThemesBanner",
 		"domainSearchButtonStyles",
 		"twoYearPlanByDefault",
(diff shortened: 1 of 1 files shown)


bsessions85:
Thanks @victorespigares. We have moved all of the e2e tests in to the wp-calypso repo and have started reading directly from active-tests.js so we no longer need to do anything for the e2e tests!

victorespigares:
Awesome @bsessions85! Thanks for the update 👍

